PR #67, opened 2018-06-22
Ancestor samples

jeromekelleher:
There is a number of breaking changes here:

1. The AncestorData file format version is bumped to version 1.0. Older versions won't be readable. The API and structure has changed slightly --- use ``python -m tsinfer <ancestors file>`` to see the new structure.
2. Older ancestors tree sequences won't work either, as the way they are stored has been updated. All these must be regenerated.
3. Any version 1.0 samples files with sequence_length of 0 won't be accepted as valid input any more. These can be patched up automatically with the ``update_seq_len.py`` script that's included here.

### DIFF
--- a/dev.py
+++ b/dev.py
@@ -229,13 +229,41 @@ def tutorial_samples():
             progress.update()
     progress.close()
 
+
+def subset_sites(ts, position):
+    """
+    Return a copy of the specified tree sequence with sites reduced to those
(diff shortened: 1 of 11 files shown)


jeromekelleher:
OK, looks like this works. Merging once CI finishes.

cc @hyanwong and @awohns, see notes above about breaking changes.

PR #69, opened 2018-07-10
add nicer plots for ancestors by physical length & number of sites

hyanwong:


evaluation.py
```diff
@@ -1138,28 +1144,112 @@ def run_ancestor_comparison(args):
...
+            plt.yscale("log")
+        ax = plt.gca()
+        ax.set_xlim(xmin=0)
+        if ancestors_are_estimated:
+            plt.title("Ancestor lengths as estimated by tsinfer")
+            ax.step(exact_line_x[:-1], exact_mean_line_y, label="True mean", where='post', color="limegreen")
+            ax.step(exact_line_x[:-1], exact_median_line_y, label="True median", where='post', color="limegreen", linestyle=":")
+            plt.xlabel("Ancestors_freq (youngest to oldest)")
+            ax.set_xlim(xmin=0)
+            _ = ax.tick_params(axis='x', which="major", length=0)
```
jeromekelleher: Why the ``_ = ax.tick_params(`` pattern here?

hyanwong: When I run this in the interpreter, ax.tick_params() returns a huge list of the params used. Is there a better way to tell python not to print out the results of the call?

hyanwong: Happy to remove. For testing, I usually run this in the interpreter, as it's hard to debug otherwise. Would appreciate some tips on how to do this better.

jeromekelleher: OK, lets remove and merge this. Can talk about how to develop off line.

jeromekelleher:
Looks good, thanks @hyanwong

PR #75, opened 2018-07-13
Cleanup evaluation

jeromekelleher:


jeromekelleher:
There's a few notable changes here:

1. I added an ``AncestorData.ancestors_length`` property which returns the length of ancestors in physical coordinates. It is definitely not a good idea to subtract the site indexes like we did before, as this isn't even in terms of sites: it's in terms of **inference** sites. So, we could have large numbers of actual sites between two inference sites. There is no reason not to use simple distance here, so I've gotten rid of all the lengths I found that were in terms of sites.

2. I had to make quite a few changes to the plotting code to make it pep8 compliant. Mainly, lines are too long and complicated. It helps a lot in terms of making the code more readable to assign intermediate values to descriptively named variables.

jeromekelleher:
> Do I need to install flake8?
Yes.

PR #79, opened 2018-07-13
Update ancestors alg

jeromekelleher:
Not to be merged --- just making branch available for use.

### DIFF
--- a/dev.py
+++ b/dev.py
@@ -114,40 +114,14 @@ def tsinfer_dev(
         print("num_sites = ", ts.num_sites)
     assert ts.num_sites > 0
 
-    G = generate_samples(ts, error_rate)
-    with tsinfer.SampleData(
-            sequence_length=ts.sequence_length,
-            path="tmp.samples",
-            chunk_size=10) as sample_data:
-        for j in range(ts.num_samples // 2):
-            sample_data.add_individual(ploidy=2, metadata={"name": "ind_{}".format(j)})
-        for site, genotypes in zip(ts.sites(), G):
-            sample_data.add_site(site.position, genotypes)
-    with tsinfer.SampleData(sequence_length=ts.sequence_length) as subset_samples:
-        for j in range(ts.num_samples // 2):
-            subset_samples.add_individual(ploidy=2, metadata={"name": "ind_{}".format(j)})
-        for site, genotypes in zip(ts.sites(), G):
-            if site.id % 2 == 0:
-                subset_samples.add_site(site.position, genotypes)
-    print(sample_data)
-    print(subset_samples)
-
-    # ts = tsinfer.infer(sample_data, simplify=False)
-    # for tree in ts.trees():
-    #     print(tree.draw(format="unicode"))
-
-    # # sample_data.finalise()
-    # # print(sample_data)
-    # print("LOADED")
-    # with tsinfer.SampleData.load("tmp.samples") as sample_data:
-    #     print(sample_data)
-    # print(sample_data)
+    sample_data = tsinfer.SampleData.from_tree_sequence(ts)
 
     ancestor_data = tsinfer.generate_ancestors(sample_data, engine=engine)
-    ancestors_ts = tsinfer.match_ancestors(sample_data, ancestor_data, engine=engine)
-    # output_ts = tsinfer.match_samples(subset_samples, ancestors_ts, engine=engine)
-    output_ts = tsinfer.match_samples(sample_data, ancestors_ts, engine=engine)
-    # dump_provenance(output_ts)
+    print(ancestor_data)
+#     ancestors_ts = tsinfer.match_ancestors(sample_data, ancestor_data, engine=engine)
+#     # output_ts = tsinfer.match_samples(subset_samples, ancestors_ts, engine=engine)
+#     output_ts = tsinfer.match_samples(sample_data, ancestors_ts, engine=engine)
+#     # dump_provenance(output_ts)
 
 
 def dump_provenance(ts):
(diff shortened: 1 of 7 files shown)


hyanwong:
Is this latest commit (00ebf1d) ready for testing?

jeromekelleher:
Yes please! This version will only remove a sample from S if it disagrees with the consensus for two sites in a row. Hopefully this will make it more robust to error...

FYI, I ended up making a couple of other changes:
1. For short-term convenience, all ancestors now are only for a single focal site. I'm not convinced that having multiple focal sites did much at scale, or that the way we were bundling them together made sense. In any case, path-compression should pick up the patterns that result when we split up an ancestor that is really composed of multiple focal sites. I can change this back to the way it was if we think it'll make a difference (and the current code isn't good enough).

2. In the previous algorithm we always had a trailing set of 0s after the site we computed an ancestral state for, up to the site at which we decided to break. Now, the ancestor stops at the last site that we computed an ancestral state. I hadn't really thought this through before, but I think the current approach is probably the right way to do it. Easy to change though if this turns out not to be the case.

hyanwong:
Testing now. FWIW I would be inclined to leave the trailing zeros you mention in (2): the true break could be anywhere up to the non-matching site. By breaking earlier, we may risk forcing breaks into positions adjacent to the last non-matching site, rather than getting the matching algorithm to locate the most likely (more generally, I would be inclined towards longer ancestors rather than shorter ones, if allowed by performance issues)

hyanwong:
Disappointingly, the subsampling results for 00ebf1d seem to be worse than the previous iteration, at least for small sample sizes (I'm labelling this c8f3abf but in the list above it seems to be 1add28f). For example, for a 1Mb length with no error, the KC metric now hovers around 2.275 (or 2.325 if polytomies broken), whereas before it was around 2.175 (2.220). With error, we are no longer getting an increasing distance with increasing sample size, but that's because we are starting off worse.


[00ebf1d-metrics_subsampling.pdf](https://github.com/tskit-dev/tsinfer/files/2220743/00ebf1d-metrics_subsampling.pdf)
[c8f3abf-metrics_subsampling.pdf](https://github.com/tskit-dev/tsinfer/files/2220744/c8f3abf-metrics_subsampling.pdf)

hyanwong:
Can we possibly switch the previous behaviour under (2) (leave the trailing zeros in) if it's easy to do so, and I can try running again, to see if this is the root of the problem?

jeromekelleher:
OK, let's see what the rest of the results say. I can swap back in the trailing zeros easily tomorrow if you want to try that out.

hyanwong:
Just got the other results in (metrics_all_tools_accuracy). Also now slightly worse than c8f3abf, even with error. I guess the easiest thing is to swap trailing zeros in, and it will be slightly more difficult to go back to multiple focal sites?

jeromekelleher:
> Just got the other results in (metrics_all_tools_accuracy). 00ebf1d is also now slightly worse than c8f3abf, even with error. I guess to test why this might be, the easiest thing is to swap trailing zeros in, and it will be slightly more difficult to go back to multiple focal sites?

Interesting... Yes, I'll swap back in the trailing zeros first.

hyanwong:
Two plots attached, if you want to check - easiest to flip between them to look at the differences.

[00ebf1d-metrics_all_tools_accuracy.pdf](https://github.com/tskit-dev/tsinfer/files/2222417/00ebf1d-metrics_all_tools_accuracy.pdf)
[c8f3abf-metrics_all_tools_accuracy.pdf](https://github.com/tskit-dev/tsinfer/files/2222419/c8f3abf-metrics_all_tools_accuracy.pdf)

jeromekelleher:
I've updated this @hyanwong --- we now pad the ancestors out with trailing zeros up to (and including) the  older site at which we failed. I set the value of the site where we break to 0 basically because it's more convenient. An argument could be made either way. Let's see how it goes...

hyanwong:
Just for the record, I should point out here that reinstating the trailing zeros (g169c3cc) reinstated the older (desirable) accuracy measures with increasing mutation rate, as seen in c8f3abf. However, it doesn't seem to have quite reinstated the previous accuracy-with-subsampling behaviour, which, for the no-error case, is still a little worse in absolute KC values in g169c3cc compared to c8f3abf. The pattern for the with-error case in g169c3cc is a little more desirable, however.

I've uploaded the plots that show this in this commit https://github.com/mcveanlab/treeseq-inference/commit/92df915a3c963dd62d9d2887dd7e1cebcee19c65. The performance on the GDF5 dataset is also not entirely restored (see the origsize10_with_poly / filtered case below). This implies to me that the "ancestors now only for a single focal site" change might not be doing quite what we expect. But this may only apply to small sample sizes.

```
#revision c8f3abf (less draconian stopping rule than MASTER)
                       filtered     full  allsnps
origsize10_with_poly   7.000000 7.615773 9.949874
subsamp6k_with_poly    7.071068 7.483315 9.643651


#revision 00ebf1d (less draconian stopping rule, allow for some error, no trailing zeros)
                       filtered     full   allsnps
origsize10_with_poly   7.549834 7.745967 10.000000
subsamp6k_with_poly    7.280110 6.855655  9.539392

#revision g169c3cc (less draconian stopping rule, allow for some error, trailing zeros)
                       filtered     full   allsnps
origsize10_with_poly   7.549834 7.745967 10.000000
subsamp6k_with_poly    7.280110 6.855655  9.643651
```

jeromekelleher:
Since we're moving in the right direction in terms of large scale real data with this, I'm going to merge this. We're clearly not done in terms of ancestor generation, but it's worthwhile keeping this current algorithm in the history for future reference.

PR #89, opened 2018-07-27
Add lots of new quality plots, plus text output

hyanwong:
also change progress/num_processes to match tsinfer.

You can output the text of bad ancestors by using the `-b` switch


jeromekelleher:
Need to add colorama to  requirements/development.txt

PR #102, opened 2018-08-29
Tools for checking ancestors tree sequences for validity

jeromekelleher:
This will interest you @hyanwong --- there's some code in here for checking the properties of a tree sequence to ensure that it's a valid ancestors_ts. I've also added a method to take any tree sequence and modify it so that it can be used as an ancestors tree sequence in sample matching (i.e., this allows us to match samples against  the original tree sequence). A nice consequence of this is that it gives us another way of doing perfect inference, without needed the input to have SMC properties.

The ``check_ancestors_ts`` should enumerate the required properties. One thing I've not resolved is that there seems to be a partial dependence on the time of nodes being sorted. You can use the trick
```python
tables.simplify(samples=np.argsort(tables.nodes.time)[::-1].astype(np.int32))
```
to make sure that it's true if you're still having issues with your constructed tree sequences.

### DIFF
--- a/dev.py
+++ b/dev.py
@@ -113,14 +113,9 @@ def tsinfer_dev(
         print("num_sites = ", ts.num_sites)
     assert ts.num_sites > 0
 
-    ts = tsinfer.eval_util.insert_perfect_mutations(ts)
+    samples = tsinfer.SampleData.from_tree_sequence(ts)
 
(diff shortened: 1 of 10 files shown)


hyanwong:
This is very useful, thanks!

PR #111, opened 2018-09-07
Add SRB ancestors from samples

jeromekelleher:
WIP.


jeromekelleher:
Merging this. We're probably not going to end up using the SRB ancestors for now, but there's some useful framework code and it doesn't affect the nominal code path.

PR #123, opened 2018-09-25
Initial draft of the mean ancestry calculation.

jeromekelleher:


tsinfer/eval_util.py
```diff
@@ -716,10 +716,138 @@ def count_sample_child_edges(ts):
...
-#     defined as the fraction of samples below a given node from each population,
-#     averaged along the length of sequence that the node is ancestral to at
-#     least one sample.
-#     """
+
+
+def node_span(ts):
+    """
+    Returns the "span" of all nodes in the tree sequence. This is defined as the
+    total length of all trees that the node participates in. A node "participates in"
```
petrelharp: I usually think of "total length" of a tree as being the sum of all the branch lengths (times the distance along the sequence it extends for, maybe); but here you just mean the distance along the sequence.

tsinfer/eval_util.py
```diff
@@ -716,10 +716,141 @@ def count_sample_child_edges(ts):
...
+        for edge in edges_in:
+            for u in [edge.parent, edge.child]:
+                if start[u] == -1:
+                    start[u] = left
+    for u in tree.nodes():
+        S[u] += ts.sequence_length - start[u]
+    return S
+
+
+def mean_sample_ancestry(ts, sample_sets, show_progress=False):
```
hyanwong: Do we want to call this "sample_ancestry" or simply something like "mean_sample_location"? We want to be careful that to avoid terminology that implies anything about the actual location/details of this ancestral node

tsinfer/eval_util.py
```diff
@@ -716,10 +716,141 @@ def count_sample_child_edges(ts):
...
+    """
+    Computes the mean sample ancestry for each node in the tree sequence. This is
+    defined as the fraction of samples below a given node from each population,
+    averaged along the length of sequence that the node is ancestral to at
```
hyanwong: There's a word or two missing from the description here: "the length of sequence that the node is ancestral to at least one sample" doesn't make sense to me.

PR #132, opened 2018-10-22
This is a small change to evaluation.py

hyanwong:
I can't work out why the commits on Oct 15th are also included, as they have already been accepted on master. However, the listed changes correctly include those from 3de60ba, which are the only changes here.


jeromekelleher:
You need to rebase this. I had a quick go and ended up in a conflict mess. There's no point in keeping your master in sync with upstream, and  merging changes is a bad idea. (Also PRs off your master isn't great practise).

The simplest thing to do is to make a fresh feature branch from upstream/master, copy  the changes you have here into it manually (copy the changed file somewhere first), and make a new PR.

PR #137, opened 2018-12-12
Remove TempJSON codec & upgrade numcodecs.

### LINKED ISSUE #136: Broken on recent numcodecs
```
module 'numcodecs.compat' has no attribute 'buffer_tobytes'
```

Should be simple fix of removing the workaround for broken JSON codec.

jeromekelleher:
Closes #136

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -50,6 +50,7 @@ setup(
         "humanize",
         "daiquiri",
         "msprime>=0.6.1",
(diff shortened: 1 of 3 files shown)


jeromekelleher:
As we were using the 'json2' codec and an identical encoding, this should cause no problems with existing files.

PR #143, opened 2019-04-03
Allow user-specified ages

hyanwong:
There are no unit tests for this new functionality. I guess we should discuss how to write them.

I have an adjustment for `evaluation.py` to incorporate user ages for sites into the evaluation suite, though (this is not included in this PR, but could be)

### DIFF
--- a/_tsinfermodule.c
+++ b/_tsinfermodule.c
@@ -108,9 +108,10 @@ static PyObject *
 AncestorBuilder_add_site(AncestorBuilder *self, PyObject *args, PyObject *kwds)
 {
     int err;
-    static char *kwlist[] = {"site_id", "frequency", "genotypes", NULL};
+    static char *kwlist[] = {"site_id", "age", "genotypes", NULL};
+    PyObject *ret = NULL;
     int site_id;
-    unsigned long frequency;
+    double age;
     PyObject *genotypes = NULL;
     PyArrayObject *genotypes_array = NULL;
     npy_intp *shape;
@@ -118,8 +119,8 @@ AncestorBuilder_add_site(AncestorBuilder *self, PyObject *args, PyObject *kwds)
     if (AncestorBuilder_check_state(self) != 0) {
         goto out;
     }
-    if (!PyArg_ParseTupleAndKeywords(args, kwds, "ikO!", kwlist,
-            &site_id, &frequency, &PyArray_Type, &genotypes)) {
+    if (!PyArg_ParseTupleAndKeywords(args, kwds, "idO!", kwlist,
+            &site_id, &age, &PyArray_Type, &genotypes)) {
         goto out;
     }
     genotypes_array = (PyArrayObject *) PyArray_FROM_OTF(genotypes, NPY_UINT8,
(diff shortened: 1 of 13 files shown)


hyanwong:
Looks good, thanks Jerome

jeromekelleher:
OK, that should do it. The only wrinkle is that I've bumped the major versions on both the ancestors and sample files so that they can use the 'age' concept now. This means that this version of tsinfer won't be able to read old samples/ancestors files. Probably the way to go is to provide a ``tsinfer upgrade`` command which will update these. I'd suggest creating an issue for this putting this as a pre-requisite for release though, rather than tacking it into this PR.

This was long overdue, thanks for getting the ball rolling @hyanwong!

jeromekelleher:
Ah, no, there is one thing to sort out before merging. I'm using a terrible algorithm in one part of the C code which needs to be fixed up. I'll sort it out tomorrow morning and merge then.

Would you mind taking a quick look over please if you get a change @hyanwong?

hyanwong:
Looks good on a first pass through. Thanks @jeromekelleher. One thing that occurred to me is that ages are really associated with *mutations* not sites. This is only relevant if there are multiple mutations per site, which we don't support in tsinfer at the moment anyway. But I mention it so as to avoid boxing ourselves into a corner later, in case we eventually want to support this. As it happens, if we have a time for each of the mutations at a site, I suspect there is an easy extension to the ancestor builder that will work when constructing ancestors. But it's unclear to me how you would get times for different mutations at the same site.

jeromekelleher:
It's a good point, thanks @hyanwong. I think it's the sort of thing we just have to deal with later when we encounter it. Changing the file formats isn't such a big deal (as there'll be an upgrade path), so I'm not too worried about it.

jeromekelleher:
OK, going to merge this.

hyanwong:
@jeromekelleher - is there a version number for this change, so that I can specify a min_tsinfer_version for Python code that requires the age functionality?

Also, I have changes to evaluation.py that test adding correct ages in to the inference path - do you want me to push them into master via a PR?

hyanwong:
@jeromekelleher - sorry, another comment:  if we only have ages for *some* of the sites, I worry that it might be rather too easy to mix true ages and frequencies by mistake, simply by omitting to pass in an age (or passing in None) to `add_site(age=...)`. My suggestion for avoiding this is to explicitly pass some null-type marker (e.g. NaN) as an age, and mark the site as "not-for-inference". I guess this might work (I don't think we need ages for non-inference sites).

If this seems like a good idea I'd be happy to document it. If it's an "approved" way to do avoid this sort of mistake, I suggest that if anyone passes `NaN` as an age, the site is automatically flagged as "not for inference".

Thoughts?

jeromekelleher:
I agree it's an easy mistake to make. Nans are tricky though. Let's talk about it in person at some point. The code isn't released so it's fine to leave it as it is for now.

There's no version number other than the hash. But this will be released as 0.2.0 at some point if you want to use that.

Please open PRs if you want to make changes.

hyanwong:
Fine. Chatting in person is the way to go, yes.

hyanwong:
p.s. any negative number would presumably be a reasonable null marker for `age=N/A`. So even `tskit.NULL` might work

hyanwong:
@jeromekelleher - I've been using tskit.NULL and this works well, throwing an error if I try inference using sites with age=tskit.NULL, so I think this is a good solution. 

Perhaps we should save two (temporary) flags when constructing a sample_data file: "has_user_age" if add_site is ever called with an `age` parameter and "has_freq_as_age" if it is ever called without. Then on finalising, we can check if both flags are set, and issue an appropriate warning (but allow the user to proceed anyway).

PR #158, opened 2019-06-18
Remove FIXME, as UNKNOWN_ALLELE now defined in constants.py

hyanwong:
This is an old constant, duplicating the definition in constants.py


jeromekelleher:
Good call, thanks.

PR #168, opened 2019-07-18
Switch to (signed) int8 for haplotype storage

hyanwong:
This also switches the C code to using signed ints for genotypes. Guess that's OK. It will fail tests unless it is used with a tskit version that includes https://github.com/tskit-dev/tskit/pull/267 which exports tskit.MISSING_DATA, which I'm now using as a replacement for tsinfer.constants.UNKNOWN_ALLELE (I think this makes sense)

### DIFF
--- a/_tsinfermodule.c
+++ b/_tsinfermodule.c
@@ -123,7 +123,7 @@ AncestorBuilder_add_site(AncestorBuilder *self, PyObject *args, PyObject *kwds)
             &site_id, &age, &PyArray_Type, &genotypes)) {
         goto out;
     }
-    genotypes_array = (PyArrayObject *) PyArray_FROM_OTF(genotypes, NPY_UINT8,
+    genotypes_array = (PyArrayObject *) PyArray_FROM_OTF(genotypes, NPY_INT8,
             NPY_ARRAY_IN_ARRAY);
     if (genotypes_array == NULL) {
         goto out;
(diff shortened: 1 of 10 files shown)


jeromekelleher:
This looks like a good start. There's a chunk of work to do to complete the change though:

1. Need to bump the file format version numbers. I'm not sure if this is a major or minor version change: will old code be able to read the new files and will new code be able to read the old files?

2. Need to update the low-level C code. The allele_t need to change to int8_t, and propagate the changes. 

3. We need to put in tskit>=0.2.0 in requirements.py and setup.py because of the tskit.MISSING_DATA constant. As such, we should wait until 0.2.0 is shipped before taking this further probably.

hyanwong:
Thanks @jeromekelleher . It looked to me like allele_t was (bizarrely) already int8_t - but I might have been skipping over the code too quickly. ISWYM about file format & tskit requirements. I'm going to keep on developing suff off this branch, as there's more stuff to do (e.g. allow missing data in samples), but as long as it passes the tests on my laptop with the dev version of tskit, I'm assuming that we will be able to update and push to tsinfer master later.

hyanwong:
> It looked to me like allele_t was (bizarrely) already int8_t

Unless I'm mistaken, allele_t is defined here: https://github.com/tskit-dev/tsinfer/blob/cf3c68aa3ce300fa046637103b72800218670205/lib/tsinfer.h#L50

jeromekelleher:
How does this relate to #168? If it's been superseded, can you close this pr please @hyanwong?

hyanwong:
Obsoleted by https://github.com/tskit-dev/tsinfer/pull/169

Issue #146, opened 2019-04-12
Default simplify to use 'keep unary' option in tskit?


jeromekelleher:
Tskit now supports the ``keep_unary`` simplify option (https://github.com/tskit-dev/tskit/pull/143) which will probably be quite useful for tskit. This will hopefully give us a reduction in ts file size, and should probably be the default format to output.

Any thoughts @hyanwong?

jeromekelleher:
If so, will depend on version 0.1.6 of tskit.

jeromekelleher:
For current 1kg trees we get:
```
-rw-r--r--. 1 jk bioinf 997M Apr 12 13:47 1kg_chr1.full-simplify.trees
-rw-r--r--. 1 jk bioinf 859M Apr 12 13:42 1kg_chr1.keep-unary.trees                                                                             
-rw-r--r--. 1 jk bioinf 1.1G Apr 12 13:41 1kg_chr1.nosimplify.trees     
```
So we do get a decent reduction in file size from keep_unary.

hyanwong:
Wow. I'm not understanding something here - I can't see how keeping extra nodes would *reduce* the file size, unless the extra nodes allow edges to be merged in some way. For example, in a slim simulation won't all unary nodes include a huge number of intermediate individuals (all the relevant ancestral individuals in every generation since the MRCA)?

jeromekelleher:
Fully simplifying does add extra edges for tsinfer generated tree sequences because we're splitting up potentially long edges to where they happen to participate in coalescences. In some cases the unsimplified tree (containing everything) can be smaller than the simplified trees, but usually only when we're doing very badly anyway.

jeromekelleher:
Closed in #176.

PR #176, opened 2019-08-12
Set keep_unary on simplify, and document (including path compression docs)

hyanwong:
We should also switch `simplify` in these cases to set `keep_unary=True`, and document accordingly.

### DIFF
--- a/tsinfer/inference.py
+++ b/tsinfer/inference.py
@@ -139,10 +139,10 @@ def verify(samples, tree_sequence, progress_monitor=None):
 
 
 def infer(
-        sample_data, progress_monitor=None, num_threads=0, path_compression=True,
-        simplify=True, engine=constants.C_ENGINE):
+        sample_data, num_threads=0, path_compression=True, simplify=True,
+        engine=constants.C_ENGINE, progress_monitor=None):
     """
-    infer(sample_data, num_threads=0)
+    infer(sample_data, num_threads=0, path_compression=True, simplify=True)
 
     Runs the full :ref:`inference pipeline <sec_inference>` on the specified
     :class:`SampleData` instance and returns the inferred
@@ -153,13 +153,19 @@ def infer(
     :param int num_threads: The number of worker threads to use in parallelised
         sections of the algorithm. If <= 0, do not spawn any threads and
         use simpler sequential algorithms (default).
+    :param bool path_compression: Whether to merge edges that share identical
+        paths (essentially taking advantage of shared recombination breakpoints).
+    :param bool simplify: Whether to remove extra tree nodes and edges that are not
+        on a path between the root and any of the samples. To do so, the final tree
+        sequence is simplified by appling the :meth:`tskit.TreeSequence.simplify` method
+        with ``keep_unary`` set to True (default = ``True``).
     :returns: The :class:`tskit.TreeSequence` object inferred from the
         input sample data.
     :rtype: tskit.TreeSequence
     """
     ancestor_data = generate_ancestors(
-        sample_data, engine=engine, progress_monitor=progress_monitor,
-        num_threads=num_threads)
+        sample_data, num_threads=num_threads,
+        engine=engine, progress_monitor=progress_monitor,)
     ancestors_ts = match_ancestors(
         sample_data, ancestor_data, engine=engine, num_threads=num_threads,
         path_compression=path_compression, progress_monitor=progress_monitor)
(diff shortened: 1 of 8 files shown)


tsinfer/inference.py
```diff
@@ -153,6 +153,11 @@ def infer(
     :param int num_threads: The number of worker threads to use in parallelised
         sections of the algorithm. If <= 0, do not spawn any threads and
         use simpler sequential algorithms (default).
+    :param bool path_compression: Whether to merge edges that share identical
+        paths (essentially taking advantage of shared recombination breakpoints).
+    :param bool simplify: Whether to simply the resulting tree sequence using
+        :meth:`tskit.TreeSequence.simplify`, which ensures that it is in canonical form
```
jeromekelleher: I guess the canonical form bit here is going to be misleading if we have ``keep_unary=True``. Probably worth making that change here if we're documenting this API?

hyanwong: Yes, do you want me to change the simplify calling code, and therefore the docs too?

jeromekelleher: Yes, this is probably a good thing to do. There'll be some breakage in the tests probably, though.

hyanwong: Yep. I'm in the middle of breaking tests anyway. What's a few more to add to the mix...

hyanwong: @jeromekelleher: how do we specify that we want to `keep_unary` in a call to `infer`. Should we have e.g. `simplify = None | (True) | (False)`, where `None` means `keep_unary` is passed to simplify? That seems a little obscure to me. Or should we have an option like `simplify_keep_unary = True | (False)`, and if people really want to do a full simplify, they can simply call it on the resulting TS?

jeromekelleher: I'd say we either call simplify with ``keep_unary`` or we don't call simplify at all. If  the user wants the fully simplifed ts, then they can do that themselves easily enough.

hyanwong: You mean we keep the parameter labelled `simplify` but just state in the docs that this means do so with `keep_unary=True`? Is there a problem with keeping that consistent with other usage?

jeromekelleher: Exactly, yeah. We're still calling simplify, just with non-default parameters.

hyanwong:
@jeromekelleher - do we want to document the `progress_monitor` parameter too? E.g. at the moment the documentation uses something like:

```
def match_ancestors(
        sample_data, ancestor_data, progress_monitor=None, num_threads=0,
        path_compression=True, extended_checks=False, engine=constants.C_ENGINE):
    """
    match_ancestors(sample_data, ancestor_data, num_threads=0, path_compression=True)
    ...
    """
```

And (weirdly to me), the `progress_monitor` param, which is 3rd, is omitted from the docstring. Presumably this means that people using this function without named parameters will be misled into thinking that they can specify `num_threads` by setting it as a 3rd (unnamed) parameter.

jeromekelleher:
> do we want to document the progress_monitor parameter too? 

No, this is too complicated, will probably change over time and is really for internal use. We don't want people using this.

> And (weirdly to me), the progress_monitor param, which is 3rd, is omitted from the docstring. Presumably this means that people using this function without named parameters will be misled into thinking that they can specify num_threads by setting it as a 3rd (unnamed) parameter.

That is weird --- we should change the signature of the function to match the docs.

hyanwong:
> we should change the signature of the function to match the docs.

Fine. I'll do this in the same PR shall I? We don't mind changing the order in this case? I guess we put it either before or after the (similarly an correctly undocumented) `engine` param?

jeromekelleher:
> Fine. I'll do this in the same PR shall I? We don't mind changing the order in this case? I guess we put it either before or after the (similarly an correctly undocumented) engine param?

That would be good, thanks.

hyanwong:
This now works and passes unit tests on my machine, but since `keep_unary` is only in tskit 0.2.0, it fails CI. There's also a small change to get the tests to pass under 0.2.0, now that tskit.MISSING_DATA is output for isolated nodes, which what https://github.com/tskit-dev/tsinfer/pull/176/commits/2dbac3a0cd6046390ea76e9156c4dee286c0626a is all about. I guess we will wait for the 0.2.0 release until merging this PR. Is that likely to be soon, @jeromekelleher ?

jeromekelleher:
tskit 0.2.0 should be shipped in a week or so --- best to wait for it to avoid complications.

hyanwong:
Right - perfect. Thanks.

jeromekelleher:
Closes #146

PR #178, opened 2019-08-14
Allow iterating in reverse over SampleData genotypes

hyanwong:
I coded this up and ended up not using it. But it might be that other people find it more generally useful. Even if not, it would be good to know if this general approach and my style is useful (creating an inner, iterator class to allow reverse chunked iteration). Knowing this may be helpful to address https://github.com/tskit-dev/tskit/issues/56

### DIFF
--- a/tsinfer/formats.py
+++ b/tsinfer/formats.py
@@ -226,6 +226,18 @@ def chunk_iterator(array):
         yield chunk[j % chunk_size]
 
 
+def rev_chunk_iterator(array):
+    """
+    Utility to iterate in reverse over the rows in the specified array efficiently
+    by accessing one chunk at a time.
+    """
+    chunk_size = array.chunks[0]
+    for j in range(array.shape[0]):
+        if j % chunk_size == 0:
+            chunk = array[-(j + chunk_size): (None if j == 0 else -j)][:]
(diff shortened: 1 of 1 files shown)


jeromekelleher:
This is the basic approach we'd need all right --- I wouldn't use a nested class though, this seems like an unnecessary complication. The real question is do we actually need this though: putting stuff in because we think it might come in handy is a sure recipe for bloated and hard to maintain code.

On the other hand, this arguable belongs upstream in zarr. I've already added an efficient chunkwise iterator there. This could be adapted to make this reversible, which would be a good feature.

hyanwong:
Yep, if it can be rolled in to `zarr` that would be much better. I'm only coding these up because I've been trying to use this sort of functionality to test aspects of `tsinfer` code. Without it, I've just taken alternative approaches instead, but they aren't so clean.

jeromekelleher:
> Yep, if it can be rolled in to zarr that would be much better. I'm only coding these up because I've been trying to use this sort of functionality to test aspects of tsinfer code. Without it, I've just taken alternative approaches instead, but they aren't so clean.

For tests it's best to dump the whole lot into a matrix rather than use fancy iterators, which you can iterate in whatever order you like. We don't need the memory efficiency.

PR #185, opened 2019-09-02
Use map_mutations from tskit, rather than bespoke locate_mutations_on…

hyanwong:
…_tree()

Remove old implementation

change docs

Tidy comments and variable names

### DIFF
--- a/tsinfer/inference.py
+++ b/tsinfer/inference.py
@@ -627,54 +627,6 @@ class Matcher(object):
     def encode_metadata(self, value):
         return json.dumps(value).encode()
 
-    def locate_mutations_on_tree(self, tree, variant, mutations):
-        """
-        Find the most parsimonious way to place mutations to define the specified
-        genotypes on the specified tree, and update the mutation table accordingly.
-        """
-        site = variant.site.id
-        samples = np.where(variant.genotypes == 1)[0]
-        num_samples = len(samples)
-        logger.debug("Locating mutations for site {}; n = {}".format(site, num_samples))
-        # Nothing to do if this site is fixed for the ancestral state.
-        if num_samples == 0:
-            return
-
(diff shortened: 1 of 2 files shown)


tsinfer/inference.py
```diff
@@ -696,21 +649,46 @@ def insert_sites(self, tables):
...
                     metadata=self.encode_metadata(site.metadata))
                 if site.inference == 1:
                     tables.mutations.add_row(
                         site=site.id, node=node[inferred_site],
                         derived_state=variant.alleles[derived_state[inferred_site]])
                     inferred_site += 1
                 else:
-                    assert ts.num_edges > 0
-                    self.locate_mutations_on_tree(tree, variant, tables.mutations)
+                    if np.all(variant.genotypes == tskit.MISSING_DATA):
```
jeromekelleher: There's bunch of stuff about missing data that I don't understand here. If you'd like this merged quickly then you'll need to remove the stuff relating to missing data --- since there's no missing data support in tskit right now, then by necessity this is currently untested. Otherwise, if you don't want to do this, you may as well roll it in with the missing data update. But, I'll need to spend a day working through this to figure out all the consequences.

hyanwong: Well, this part is simply checking for the extreme case of completely missing data in a non-inference site, which would otherwise cause map_mutations to fail. I can remove it for this PR, and just assume that we'll never hit that issue if we aren't allowing missing data input in the first place.

jeromekelleher: As I said, up to you to do play it, if you want to remove the missing data references from here to get this merged quickly, then great.

hyanwong:
So this is now the version without any reference to missing data.

jeromekelleher:
Perfect, thanks. I'll merge once CI passes.

hyanwong:
Thanks. Seems done now.

Issue #149, opened 2019-05-01
Change ``age`` to ``time``


jeromekelleher:
For consistency. See #145 for discussion.

Issue #144, opened 2019-04-04
Need `tsinfer upgrade` command


jeromekelleher:
#143 added the ability to specify ages for sites, and in doing so changed the Samples and Ancestor file formats. As such, the current version of tsinfer in master will not be able to read files produced by the released version.

We need to provide an upgrade path for these old files via a ``tsinfer upgrade`` CLI command. Updgrading the files will be straightforward enough, as we can just read them in with Zarr and rename a column (in ancestors) or add and compute one (in samples).

cc'ing @hyanwong  and @awohns, so that you are aware this is an issue and a solution is planned.

hyanwong:
Presumably this also needs to account to the change from `np.uint8` to `np.int8` (allowing for missing data) in SampleData files?

jeromekelleher:
Yes.

hyanwong:
Also breaking changes have been (or will be) made by #186 (change "age" to "time"), #190 (add "time" to individuals, not just variants), and possibly #188 (`sites_inference` stored as `bool` rather than `np.uint8`)

hyanwong:
New changes in tsinfer 0.3, SampleData format 5 are caused by adding `individuals_flags` and (from version 4 to version 5) using tskit.UNKNOWN_TIME rather that `-np.inf` in the sites_time column.

jeromekelleher:
There's a bit too much to do here for 0.2.0 - I've had a look, and writing an upgrade() tool is messy. I'm going to close the issue as a "wontfix", but if someone really does need to have this functionality we can reopen. It's worth noting that tskit 0.3 is going to require a completely new file format (see #373), and we will make a lot of effort to keep things stable from there on.

hyanwong:
Yes, agree (I think you mean tsinfer 0.3 above)

PR #186, opened 2019-09-03
Change "age" references to "time"

hyanwong:
Fixes https://github.com/tskit-dev/tsinfer/issues/149

This also changes the underlying storage name from `sites/age` to `sites/time`, and the name of the equivalent parameter in `SampleData.add_site()`, so it will break code if anyone is specifying the "age" parameter by name not position. But I think @awohns is the only user of this functionality at the moment.

If we want `tsinfer` to be terminologically consistent with `tskit` like this, then this is the time to do it!

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,4 +1,45 @@
 ********************
+[0.1.5] - 2019-09-25
+********************
+
+
+**Breaking changes**:
+
+- Bumped SampleData file format version to 2.0, then to 3.0 as a result of the additions
+  below. Older SampleData files will not be readable and must be regenerated.
+
+- Users can specify variant ages, via ``sample_data.add_sites(... , time=user_time)``.
+  If not ``None``, this overrides the default time position of an ancestor, otherwise
+  ancestors are ordered in time by using the frequency of the derived variant (#143).
+  This addition bumped the file format to 2.0
+
+- Change "age" to "time" to match tskit/msprime notation, and to avoid confusion
+  with the age since birth of an individual (#149). Together with the 2 changes below,
+  this addition bumped the file format to 3.0.
+
+- Add the ability to record user-specified times for individuals, and therefore
+  the samples contained in them (currently ignored during inference). Times are
+  added using ``sample_data.add_individual(... , time=user_time)`` (#190). Together
+  with the changes above and below, this addition bumped the file format to 3.0.
+
+- Change ``tsinfer.UNKNOWN_ALLELE`` to ``tskit.MISSING_DATA`` for marking unknown regions
+  of ancestral haplotypes (#188) . This also involves changing the allele storage to a
+  signed int from ``np.uint8`` which matches the tskit v0.2 format for allele storage 
+  (see https://github.com/tskit-dev/tskit/issues/144). Together with the 2 changes above,
+  this addition bumped the file format to 3.0.
+
+**New features**:
+
+- Map non-inference sites onto the tree by using the built-in tskit
+  ``map_mutatations`` method. With further work, this should allow triallelic sites
+  to be mapped (#185)
+  
+- The default tree sequence returned after inference when ``simplify=True`` retains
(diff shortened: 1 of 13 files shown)


jeromekelleher:
Looks good thanks, we definitely want to do this. Before we merge we need to

a) Bump the file format major version number(s), as this is a breaking change (old code will no longer be able to read these files, and we'll no longer be able to read old files.). We should open an issue tracking whether we want to add an 'upgrade' function to bring old .samples files up to date.

b) Pad out the test coverage if the missing stuff is Python (don't worry if it's C).

hyanwong:
> a) Bump the file format major version number(s), as this is a breaking change (old code will no longer be able to read these files, and we'll no longer be able to read old files.). 

I don't know where this is stored. Would it be OK for you to do this - much quicker I guess.

> We should open an issue tracking whether we want to add an 'upgrade' function to bring old .samples files up to date.

We have https://github.com/tskit-dev/tsinfer/issues/144 - can we use that?

> b) Pad out the test coverage if the missing stuff is Python (don't worry if it's C).

The main coverage seems to be no tests of the `print_state` functions used for debugging. I'm about to stick in a few tests that simply call them and check they don't bomb out. No point in doing anything more detailed, as I'm guessing the output will change as we find other issues to debug.

jeromekelleher:
> I don't know where this is stored. Would it be OK for you to do this - much quicker I guess.

It's in the formats.py file somewhere. Should be easy enough to track down? I'm not going to be working on this for another week or two.

> We have #144 - can we use that?

Yep, that's good.

> The main coverage seems to be no tests of the print_state functions used for debugging. I'm about to stick in a few tests that simply call them and check they don't bomb out. No point in doing anything more detailed, as I'm guessing the output will change as we find other issues to debug.

Sounds good.

hyanwong:
> > I don't know where this is stored. Would it be OK for you to do this - much quicker I guess.
> 
> It's in the formats.py file somewhere. Should be easy enough to track down? I'm not going to be working on this for another week or two.

Do you mean [this](https://github.com/tskit-dev/tsinfer/blob/919174ca5716663308d4f9ab75393d6b83e0d40f/tsinfer/formats.py#L704) - if so should I change it to `(2, 1)` or `(3, 0)`?

jeromekelleher:
> Do you mean this - if so should I change it to (2, 1) or (3, 0)?

Yes, that's it. (We should start a Changelog also to log all these breaking changes in case we forget)

hyanwong:
Format bumped, and changelog written. I wonder, however, if we should wait to include https://github.com/tskit-dev/tsinfer/issues/188 before merging (if we agree that it is a good idea), as I guess changing from `uint8` to `bool` for sites_inference might mean a change in the underlying zarr storage format?

hyanwong:
Actually, I think we should also pick off the commit that changes the allele type from unsigned to signed (https://github.com/hyanwong/tsinfer/commit/6900f51faa19bb3f1c8d1aa87b25dcf104c96d30) and incorporate this too, as that will also create a breaking change. Then we make all the breaking changes at once, and incorporate into format 3.0. Shall I make a separate PR for the uint->int change @jeromekelleher ?

hyanwong:
Just rebased and this is now ready to merge, which should bump us to tsinfer file versions 3.0. It may be worth releasing tsinfer v0.1.5 as a result of these changes. @jeromekelleher ?

jeromekelleher:
LGTM, thanks @hyanwong, will merge once CI is done.

Re releasing, I think we should provide an upgrade path for old files (#144)  before we do this. It'll be hella painful to have to reprocess all the UKB files (e.g.), and the upgrade should be easy to do.

hyanwong:
Good point about the upgrade path. I'm not sure how to do this, and don't really want to take it on, however.

jeromekelleher:
Probably makes more sense to wait on the 0.2.0 release https://github.com/tskit-dev/tsinfer/milestone/1 I think --- we've already enough changes for a major version, so we'd be better to bundle them all together.

Issue #148, opened 2019-05-01
Some way to specify sample ages


hyanwong:
For ancient samples (ping @awohns), we want some way to specify that the node times are not 0. The obvious place to do this is in `SampleData.add_individual()`, but the problem is that node times are not stored on an individual, but instead on the sample nodes themselves.

Nevertheless, I think this is probably the best way forwards. The steps would be:
1) We can call `SampleData.add_individual()`with an optional `age` or `time` parameter (see https://github.com/tskit-dev/tsinfer/issues/145#issuecomment-488240081), storing this in a new numpy array (possibly `samples_times`: this could be of size ploidy*individual, rather than the same length as e.g. `samples_metadata`)
2) When `SampleData.add_site()` is then called, we check that age allocated to the site is compatible with the samples, raising an error if not. That is, we find the oldest age (`a`) of all the samples with a `1` for that site, and check that the allocated age for that site (which could be the freq category) is at least as old as `a`
3) When `match_samples` is called, we restrict matching against only those ancestors strictly older than each sample, and allocate sample node times according to the `samples_times` array.

hyanwong:
NB: this assumes that samples can't actually be ancestors. If we want to allow this possibility, we might also need to change the `match_ancestors` method. OTOH - we could allow people to do this by hand using `augment_ancestors`.

jeromekelleher:
Agreed on the need for the basic functionality. Will require a fair bit of work to do internally though I think.

hyanwong:
I think this is a good issue for @awohns to tackle, as it will be pivotal to his aDNA approaches. I will ask him to have a look.

awohns:
@hyanwong and I were just discussing this @jeromekelleher. 
There is a question about whether this 'time' parameter should be stored as individuals/time or samples/time in the SampleData Zarr arrays. One the one hand, tskit stores times for _sample_ nodes. On the other hand, when you add a 'time' you are adding it in the sample_data.add_individual() function.

hyanwong:
After discussion with @jeromekelleher we decided to store the time of an individual, rather than a sample. This is coded up in #190

jeromekelleher:
Closed in #190.

hyanwong:
[...]

PR #190, opened 2019-09-24
Support for adding times to individuals and samples

hyanwong:
Starts addressing #148 - useful for @awohns I hope.

This doesn't actually add anything to the inference pathway, merely adds the possibility to store historical time for individuals (and therefore samples). I don't know whether to also add a check in the inference pathway to raise an error (for the moment) if we try to infer with historical samples present?


jeromekelleher:
LGTM, merging.

PR #192, opened 2019-09-25
Switch to (signed) int8 & tskit.MISSING_DATA to mark unknown regions …

hyanwong:
…of ancestral haplotypes

This should complete the move to tsinfer file format 3.0

### DIFF
--- a/tsinfer/formats.py
+++ b/tsinfer/formats.py
@@ -17,7 +17,7 @@
 # along with tsinfer.  If not, see <http://www.gnu.org/licenses/>.
 #
 """
-Manage tsinfer's various HDF5 file formats.
+Manage tsinfer's various file formats.
 """
 import collections.abc as abc
 import datetime
@@ -757,7 +757,7 @@ class SampleData(DataContainer):
             dtype=np.float64)
         sites_group.create_dataset(
             "genotypes", shape=(0, 0), chunks=(self._chunk_size, self._chunk_size),
-            compressor=self._compressor, dtype=np.uint8)
+            compressor=self._compressor, dtype=np.int8)
         sites_group.create_dataset(
             "inference", shape=(0,), chunks=chunks, compressor=self._compressor,
             dtype=np.uint8)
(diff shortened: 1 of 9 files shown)


tsinfer/formats.py
```diff
@@ -1116,7 +1116,7 @@ def add_site(
         :return: The ID of the newly added site.
         :rtype: int
         """
-        genotypes = np.array(genotypes, dtype=np.uint8, copy=False)
```
jeromekelleher: I don't think we should do this --- this'll be a significant extra cost here when we're building samples files. I don't mind if we're not quite as general in the inputs, this is a performance critical step.

jeromekelleher: To be clear, let's keep it as ``genotypes = np.array(genotypes, dtype=np.int8, copy=False)``

hyanwong: Ah, ISWYM. But I think the safe_np_int_cast will (if given an array of `np.uint8`) not make a copy, but simply return the array as-is. It calls

```
return input_array.astype(dtype, casting='safe', copy=copy)
```

and I'm assuming that if dtype is already the same as input_array.dtype, it will return the original array, no?

hyanwong: In other words, if the user wants to be fast, they can submit the data as an array of dtype `np.int8` and it should be just as fast as it is at present. I think...

jeromekelleher: I see, the checks only occur when the same call we're making anyway fails. OK, you're right, let's use ``safe_np_int_cast``. We still need ``copy=False`` though.

hyanwong: Agree, we need copy=False, but this is the default for `safe_np_int_cast()`

jeromekelleher: Right, doh!

jeromekelleher:
Merged, thanks @hyanwong!

tsinfer/formats.py
hyanwong: (NB I set the default to copy = False due to your comment at https://github.com/tskit-dev/tskit/pull/245#discussion_r296316747)

jeromekelleher: Yes, sorry, I forgot about this.

PR #195, opened 2019-10-17
Check that sample and ancestor files are finalised

### LINKED ISSUE #125: Check if samples and ancestors files are finalised on input

hyanwong:
Fixes #125 - as raised by @brianzhang01

### DIFF
--- a/tests/test_inference.py
+++ b/tests/test_inference.py
@@ -21,10 +21,12 @@ Tests for the inference code.
 """
 import unittest
 import random
+import os.path
 import string
 import io
 import json
 import unittest.mock as mock
+import tempfile
 
 import numpy as np
 import msprime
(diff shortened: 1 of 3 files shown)


tsinfer/formats.py
```diff
@@ -435,6 +435,13 @@ def _check_write_modes(self):
         if self._mode not in (self.EDIT_MODE, self.BUILD_MODE):
             raise ValueError("Invalid operation: must be in edit or build mode")
 
+    def _check_finalised(self):
+        if not self.finalised:
+            error_msg = "The {} file".format(self.format_name)
```
brianzhang01: I'm pretty sure we should change `"The {} file"` --> `"The {} file "`, with a space after "file"

hyanwong: Good point

tests/test_inference.py
```diff
@@ -54,6 +55,77 @@ def assertTreeSequencesEqual(self, ts1, ts2):
         self.assertEqual(t1.mutations, t2.mutations)
 
 
+class TestUnfinalisedErrors(unittest.TestCase):
+    def test_generate_ancestors(self):
+        with tempfile.NamedTemporaryFile() as tmpfile:
+            for path in [None, tmpfile.name]:
+                sample_data = tsinfer.SampleData(path=path, sequence_length=2)
+                sample_data.add_site(1, genotypes=[0, 1, 1, 0], alleles=["G", "C"])
+                print(dir(sample_data))
```
jeromekelleher: Spamming the output here.

tests/test_inference.py
```diff
@@ -54,6 +55,77 @@ def assertTreeSequencesEqual(self, ts1, ts2):
...
+                # match_ancestors fails when samples unfinalised
+                unfinalised = tsinfer.SampleData(sequence_length=2)
+                unfinalised.add_site(1, genotypes=[0, 1, 1, 0], alleles=["G", "C"])
+                self.assertRaises(
+                    ValueError, tsinfer.match_ancestors, unfinalised, anc)
+
+    def test_match_ancestors_unfinalised_ancestors(self):
+        with tempfile.NamedTemporaryFile() as tmpfile:
+            for path in [None, tmpfile.name]:
+                sample_data = tsinfer.SampleData(sequence_length=2)
```
jeromekelleher:
```python
with  tsinfer.SampleData(sequence_length=2) as sample_data:
     sample_data.add_site(1, genotypes=[0, 1, 1, 0], alleles=["G", "C"])
```
would be preferable for the correct case I think.

tests/test_inference.py
```diff
@@ -54,6 +55,77 @@ def assertTreeSequencesEqual(self, ts1, ts2):
...
+
+    def test_make_ancestor_data(self):
+        with tempfile.NamedTemporaryFile() as tmpfile:
+            for path in [None, tmpfile.name]:
+                sample_data = tsinfer.SampleData(path=path, sequence_length=2)
+                sample_data.add_site(1, genotypes=[0, 1, 1, 0], alleles=["G", "C"])
+                self.assertRaises(ValueError, tsinfer.AncestorData, sample_data)
+
+    def test_match_ancestors_unfinalised_samples(self):
+        with tempfile.NamedTemporaryFile() as tmpfile:
```
jeromekelleher: I don't think we need to repeat the check on "with a file or not" for each test. We can probably just do this once each for samples files and ancestors files.

tests/test_inference.py
```diff
@@ -54,6 +55,77 @@ def assertTreeSequencesEqual(self, ts1, ts2):
...
+                sample_data = tsinfer.SampleData(path=path, sequence_length=2)
+                sample_data.add_site(1, genotypes=[0, 1, 1, 0], alleles=["G", "C"])
+                print(dir(sample_data))
+                self.assertRaises(ValueError, tsinfer.generate_ancestors, sample_data)
+                sample_data.finalise()
+                tsinfer.generate_ancestors(sample_data)
+
+    def test_make_ancestor_data(self):
+        with tempfile.NamedTemporaryFile() as tmpfile:
+            for path in [None, tmpfile.name]:
```
jeromekelleher: Annoyingly, this won't work on Windows. What you have to do is 
```python
with tempfile.TemporaryDirectory() as tmpdir:
    tmpfile = pathlib.Path(tmpdir) / "tmp.samples"
    # and so on
```

tsinfer/formats.py
```diff
@@ -435,6 +435,13 @@ def _check_write_modes(self):
...
+            if self.path:
```
jeromekelleher:
```python
if self.path is not None:
```

tests/test_inference.py
hyanwong: Well spotted. Ta.

tests/test_inference.py
hyanwong: It looks from `test_formats.py` that even using TemporaryDirectory these are skipped in windows anyway, so I'll just skip.

tests/test_inference.py
hyanwong: OK

jeromekelleher:
Looks good thanks, merging.

PR #196, opened 2019-10-17
Make tests pass with edges with id

hyanwong:
Now that edge objects contain an id in tskit (latest dev version), we need to change the tsinfer unit tests a little.


jeromekelleher:
Looks good, ta.

PR #197, opened 2019-10-25
Allow sites_time to be set

### LINKED ISSUE #145: Make SampleData.sites_age editable
It seems likely people will want to edit the age field. Make this editable like the ``inference`` column.

hyanwong:
Fixes #145

Also I corrected a bug in `test_update_sites_inference_non_copy_mode()` (note test name slightly changed for consistency with `test_update_sites_time_non_copy_mode`), where we previously tested setting the inference sites with an empty array, that was going to fail even if not in copy mode. I could spin this into a separate PR, but it seems too trivial for that.

Also some code duplication here in e.g. `test_copy_update_sites_inference` and `test_copy_update_sites_time` but it seemed the easiest way to write the tests.

### DIFF
--- a/tests/test_formats.py
+++ b/tests/test_formats.py
@@ -701,15 +701,17 @@ class TestSampleData(unittest.TestCase, DataContainerMixin):
             input_file.add_site(position=0, alleles=alleles, genotypes=genotypes)
         editable_sample_data = input_file.copy()
         # Try editing: use setter in the normal way
-        editable_sample_data.sites_inference = np.array([False])
+        editable_sample_data.sites_inference = [True]
+        editable_sample_data.sites_time = [0.0]
         # Try editing: use setter via setattr
-        setattr(editable_sample_data, 'sites_inference', np.array([True]))
+        setattr(editable_sample_data, 'sites_inference', [False])
+        setattr(editable_sample_data, 'sites_time', [1.0])
         editable_sample_data.add_provenance(datetime.datetime.now().isoformat(), {})
-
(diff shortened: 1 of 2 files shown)


hyanwong:
Closing and reopening to trigger a travis re-run (failed due to something else)

tests/test_formats.py
```diff
@@ -744,7 +746,28 @@ def test_copy_update_inference_sites(self):
                 self.assertEqual(list(copy.sites_inference), inference)
                 self.assertEqual(list(data.sites_inference), [True, True, True, True])
 
-    def test_update_inference_sites_bad_data(self):
+    @unittest.skipIf(sys.platform == "win32",
```
jeromekelleher: Now you're using a tempdir the windows problem shouldn't be an issue any more.

hyanwong: I'm not sure tempdir solves the problem. I've just tried uncommenting (and changing the file names / dirs to be unique) and it still fails with

```
PermissionError: [WinError 32] The process cannot access the file because it is being used by another process:
```

See https://ci.appveyor.com/project/tskit-dev/tsinfer/builds/28377125

There are a load of other places that use tempdir but are also commented out in the same file, presumably for the same reason.

jeromekelleher: OK, let's not worry about it then. Thanks for checking.

hyanwong: Right, dropped the win testing commit I used. This should be ready to merge.

PR #198, opened 2019-10-25
Allow max_file_size to be set

### LINKED ISSUE #152: Disk quota exceeded error in zarr/storage.py
Running tsinfer from python on a cluster (which has file size quota) gives me an error:
```
Traceback (most recent call last):

  File "parse_seqgen.py", line 44, in <module>

    with tsinfer.SampleData(path=outfile) as sample_data:

  File "/net/gs/vol1/home/lpsmith/.local/lib/python3.6/site-packages/tsinfer/formats.py", line 704, in __init__

    super().__init__(**kwargs)

  File "/net/gs/vol1/home/lpsmith/.local/lib/python3.6/site-packages/tsinfer/formats.py", line 256, in __init__

    store = self._new_lmdb_store()

  File "/net/gs/vol1/home/lpsmith/.local/lib/python3.6/site-packages/tsinfer/formats.py", line 314, in _new_lmdb_store

    return zarr.LMDBStore(self.path, subdir=False)

  File "/net/gs/vol1/home/lpsmith/.local/lib/python3.6/site-packages/zarr/storage.py", line 1614, in __init__

    self.db = lmdb.open(path, **kwargs)

lmdb.Error: /net/gs/vol1/home/lpsmith/tsinfer_run/50tip_native0/treedump: Disk quota exceeded
```
I was able to fix it (sort of) by changing the zarr/storage.py line from:

map_size = (2**40 if sys.maxsize >= 2**32 else 2**28)

to:

map_size = (2**28)

which is large enough for my purposes.  Just checking sys.maxsize does not help: as far as the system is concerned, you can have files much larger than 2^40.  It's the *quota* system that instead needs to be checked (somehow).

hyanwong:
Fixes #152

No tests implemented. @jeromekelleher: can you suggest the best way to test this?

### DIFF
--- a/tests/test_formats.py
+++ b/tests/test_formats.py
@@ -239,7 +239,7 @@ class TestSampleData(unittest.TestCase, DataContainerMixin):
                         0, alleles=['0', '1'],
                         genotypes=np.zeros(base_size, dtype=np.int8))
             # Work around https://github.com/tskit-dev/tsinfer/issues/201
-            small_sample_file.close()
+            small_sample_file.data.store.close()
 
     def test_acceptable_max_file_size(self):
         with tempfile.TemporaryDirectory(prefix="tsinf_format_test") as tempdir:
--- a/tests/test_formats.py
+++ b/tests/test_formats.py
@@ -34,6 +34,7 @@ import msprime
 import numcodecs
 import numcodecs.blosc as blosc
 import zarr
+import lmdb
 import tskit
 
 import tsinfer
@@ -206,6 +207,58 @@ class TestSampleData(unittest.TestCase, DataContainerMixin):
                 self.assertEqual(array.compressor, compressor)
             with tsinfer.load(filename) as other:
                 self.assertEqual(other, input_file)
+
+    def test_bad_max_file_size(self):
+        with tempfile.TemporaryDirectory(prefix="tsinf_format_test") as tempdir:
+            filename = os.path.join(tempdir, "samples.tmp")
+            for bad_size in ['a', '', -1]:
+                self.assertRaises(
+                    ValueError, formats.SampleData, path=filename,
+                    max_file_size=bad_size)
+            for bad_size in [[1, 3], np.array([1, 2])]:
+                self.assertRaises(
+                    TypeError, formats.SampleData, path=filename, max_file_size=bad_size)
+
+    def test_too_small_max_file_size_init(self):
+        with tempfile.TemporaryDirectory(prefix="tsinf_format_test") as tempdir:
+            # Fail immediately if the max_size is so small we can't even create a file
+            filename = os.path.join(tempdir, "samples.tmp")
+            self.assertRaises(
+                lmdb.MapFullError, formats.SampleData,
+                path=filename, sequence_length=1, max_file_size=1)
+
+    def test_too_small_max_file_size_add(self):
+        with tempfile.TemporaryDirectory(prefix="tsinf_format_test") as tempdir:
+            base_size = 2**16  # Big enough to allow the initial file to be created
+            # Fail during adding a large amount of data
+            with self.assertRaises(lmdb.MapFullError):
+                filename = os.path.join(tempdir, "samples.tmp")
+                with formats.SampleData(path=filename, sequence_length=1,
+                                        max_file_size=base_size) as small_sample_file:
+                    small_sample_file.add_site(
+                        0, alleles=['0', '1'],
+                        genotypes=np.zeros(base_size, dtype=np.int8))
+            # Work around https://github.com/tskit-dev/tsinfer/issues/201
+            small_sample_file.close()
+
+    def test_acceptable_max_file_size(self):
+        with tempfile.TemporaryDirectory(prefix="tsinf_format_test") as tempdir:
+            # set a reasonably large number of sites and samples, and check we
+            # don't bomb out
+            n_samples = 2**10
+            n_sites = 2**12
+            np.random.seed(123)
+            filename = os.path.join(tempdir, "samples.tmp")
+            with formats.SampleData(path=filename, sequence_length=n_sites,
+                                    compressor=False, max_file_size=None) as samples:
+                for pos in range(n_sites):
+                    samples.add_site(
+                        pos, alleles=['0', '1'],
+                        genotypes=np.random.randint(2, size=n_samples, dtype=np.int8))
+            self.assertEqual(samples.num_sites, n_sites)
+            self.assertEqual(samples.num_samples, n_samples)
+            self.assertGreater(samples.file_size, n_samples*n_sites)
+            samples.close()
 
     def test_defaults_no_path(self):
         ts = self.get_example_ts(10, 10)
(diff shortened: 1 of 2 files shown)


jeromekelleher:
Test it with something small (say, 1 byte, 1K, etc) and verify that building the file fails?

Will also want to verify that a meaningful error is thrown is we pass in 0 bytes, negative values, a string, etc.

hyanwong:
> Test it with something small (say, 1 byte, 1K, etc) and verify that building the file fails?

Done this. I needed to set a (semi arbitrary) value (max_file_size=2**16) at one point to allow the file to actually be created before testing to overflow it. I only tested this for SampleData files - the code path is the same for AncestorData files, but they are harder to fill arbitrarily.

> Will also want to verify that a meaningful error is thrown is we pass in 0 bytes, negative values, a string, etc.

Also done, both for SampleData and AncestorData.

hyanwong:
Reopneing to force appveyor rebuild

tests/test_formats.py
```diff
@@ -128,6 +129,30 @@ def test_defaults_with_path(self):
...
+            base_size = 2**16  # Big enough to allow the initial file to be created
+            filename = os.path.join(tempdir, "samples.tmp")
+            # Fail immediately if the max_size is so small we can't even create a file
+            self.assertRaises(
+                 lmdb.MapFullError, formats.SampleData,
+                 path=filename, sequence_length=1, max_file_size=1)
+            sample_file = formats.SampleData(
+                    path=filename, sequence_length=1, max_file_size=base_size)
+            self.assertRaises(
+                lmdb.MapFullError, sample_file.add_site, 0,
```
jeromekelleher: Do we always flush the data to file straightaway here? Might be better to do:
```python
with self.assertRaises(lmdb.MapFullError):
    sample_file.add_site, 0, alleles=['0', '1'], genotypes=np.zeros(base_size, dtype=np.int8))
    sample_file.flush() # Or whatever we call it.
```

tests/test_formats.py
```diff
@@ -892,6 +915,14 @@ def test_defaults_with_path(self):
             with tsinfer.load(filename) as other:
                 self.assertEqual(other, ancestor_data)
 
+    def test_bad_max_file_size(self):
+        with tempfile.TemporaryDirectory(prefix="tsinf_format_test") as tempdir:
+            filename = os.path.join(tempdir, "ancestors.tmp")
+            for bad_size in ['a', '', -1, [1, 3], np.array([1, 2])]:
+                self.assertRaises(
+                    (ValueError, TypeError), formats.AncestorData,
```
jeromekelleher: Can we separate these out into two assertions please? This way I don't know which one throws a ValueError and which one throws a TypeError.

tsinfer/formats.py
```diff
@@ -304,21 +313,21 @@ def _open_readonly(self):
...
-            # Arbitrarily, we allocate 1GiB (2**30 bytes). Windows users who want to run
-            # large inferences may need to increase the hardcoded map_size below.
-            return zarr.LMDBStore(self.path, map_size=2**30, subdir=False)
-        else:
+        if map_size is None:
             return zarr.LMDBStore(self.path, subdir=False)
+        else:
+            try:
+                map_size = int(map_size)
+                if map_size <= 0:
```
jeromekelleher: This is  convoluted. Why not:
```python
map_size = int(map_size)   # Will throw appropriate error
￼if map_size <= 0: 
     raise ValueError("max_file_size must be > 0")
```

tsinfer/formats.py
```diff
@@ -699,6 +717,13 @@ class SampleData(DataContainer):
     :param int chunk_size: The chunk size used for
         `zarr arrays <http://zarr.readthedocs.io/>`_. This affects
         compression level and algorithm performance. Default=1024.
+    :param int max_file_size: If a file is being used to store this data, set
+        a maximum size in bytes for the stored file. The default is to allow
+        relatively large files, so if you are running on a system with disk
```
jeromekelleher: This is vague --- what's a relatively large file? We need to define what our default is, not force the user to figure it out.

tests/test_formats.py
hyanwong: Done

tests/test_formats.py
hyanwong: Oh, I've never used the `with self.assertRaises()` format in anger before. Thanks.

tsinfer/formats.py
hyanwong: Doh, yes, sorry

tsinfer/formats.py
hyanwong: OK. The default is 2**30 on windows, or the zarr default otherwise. I'll modify the text accordingly.

hyanwong: How about 

```
        If a file is being used to store this data, set
        a maximum size in bytes for the stored file. If None, the default value
        for `zarr.LMDBStore  <http://zarr.readthedocs.io/>` is used
        (except on windows, see note above). If you are running on a system with
        storage space constraints, you may wish to explictly set a value that will
        not use up your storage quota. Default=None (or 2**30 on windows systems).
```

hyanwong: The GiB equivalent to 2**30 is in the note above. I decided to put it there as it's confusing to use different units in the `Default=xxx` section, although I could, I suppose, repeat this, i.e.  `Default=None (or 2**30, i.e. 1GiB, on windows systems).`

Re the zarr default, personally I'm more used to the upstream libraries like zarr deciding a sensible default for me, and it just seemed like less maintenance for us to simply rely on zarr setting something reasonable.  But I'm not that bothered either way. It [seems like](https://zarr.readthedocs.io/en/stable/_modules/zarr/storage.html) zarr defaults to `map_size = (2**40 if sys.maxsize >= 2**32 else 2**28)` which I guess we might want to copy?

hyanwong:
I have tested this on a windows box and the non-coverage lines identified by codecov are visited on windows, so I suspect that codecov has not been set up in this repo to gather coverage information from the AppVeyor builds.

jeromekelleher:
I've updated this a bit to try and clarify @hyanwong. If you're OK with the changes I'll merge.

hyanwong:
> I've updated this a bit to try and clarify @hyanwong. If you're OK with the changes I'll merge.

Fine by me, although I might leave in the actual integer value for the parameter in somewhere (e.g. in brackets), as an example. Otherwise you have to read rather carefully to realise you need to specify the number in bytes, when we quote gigabytes and terabytes.

Do I need to merge your changes into my PR, or can you do it on your end?

jeromekelleher:
> Fine by me, although I might leave in the actual integer value for the parameter in somewhere (e.g. in brackets), as an example. Otherwise you have to read rather carefully to realise you need to specify the number in bytes, when we quote gigabytes and terabytes.

Good call, updated.

> Do I need to merge your changes into my PR, or can you do it on your end?

No, I can do it here. I've just added an extra commit to your PR and rebased.

hyanwong:
Great, merge it!

PR #199, opened 2019-10-25
Test windows conda update (fix tempdir)

hyanwong:
Not for merging - just to try to get windows tempdir permissions working

### DIFF
--- a/tests/test_formats.py
+++ b/tests/test_formats.py
@@ -725,8 +725,6 @@ class TestSampleData(unittest.TestCase, DataContainerMixin):
         self.assertNotEqual(copy.uuid, data.uuid)
         self.assertTrue(copy.data_equal(data))
 
-    @unittest.skipIf(sys.platform == "win32",
-                     "windows simultaneous file permissions issue")
     def test_copy_update_sites_inference(self):
         with formats.SampleData() as data:
             for j in range(4):
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -13,6 +13,7 @@ install:
   # We have symlinks in the repo, which need to be treated specially on Windows.
   - cmd: git config core.symlinks true
   - cmd: git reset --hard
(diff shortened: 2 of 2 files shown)


tests/test_formats.py
```diff
@@ -725,15 +743,13 @@ def test_copy_new_uuid(self):
...
         with formats.SampleData() as data:
             for j in range(4):
                 data.add_site(position=j, alleles=["0", "1"], genotypes=[0, 1, 1, 0])
         self.assertEqual(list(data.sites_inference), [True, True, True, True])
 
-        with tempfile.TemporaryDirectory(prefix="tsinf_format_test") as tempdir:
+        with temporary_directory(prefix="tsinf_format_test") as tempdir:
             filename = os.path.join(tempdir, "samples.tmp")
             for copy_path in [None, filename]:
                 copy = data.copy(path=copy_path)
```
jeromekelleher: I don't think contextmanager stuff this is necessary --- we're just being sloppy about closing files. Isn't the ``copy`` still open here at the end?

hyanwong: Thanks jerome. Yes, I realised that the problem is probably simply that we are not closing the SampleData file.

I'm assuming that the way we do that is simply to call [sample_data.close()](https://github.com/tskit-dev/tsinfer/blob/7442f6ff5b98497061d9729a40b953c9bc41910c/tsinfer/formats.py#L336) - I'll try testing this.

jeromekelleher: Or use the context manager for the SampleData file --- that's what it's for.

jeromekelleher: Ah, strike that, there may not be a context manager for the copy method.

hyanwong: Yep, right. I'll go through the other tests that are commented out and see if I can use a context manager for those, though.

PR #202, opened 2019-10-30
Allow compression to be forced to none

hyanwong:
The documentation was misleading here. There was no way to force no compression (and no test for the default compression when a path was specified). I think it's sensible to force no compression when `compressor=FALSE`

### DIFF
--- a/tsinfer/formats.py
+++ b/tsinfer/formats.py
@@ -240,10 +240,9 @@ class DataContainer(object):
     FORMAT_VERSION = None
 
     def __init__(
-            self, path=None, num_flush_threads=0, compressor=None, chunk_size=1024):
+            self, path=None, num_flush_threads=0, compressor=DEFAULT_COMPRESSOR,
+            chunk_size=1024):
         self._mode = self.BUILD_MODE
-        if path is not None and compressor is None:
-            compressor = DEFAULT_COMPRESSOR
         self._num_flush_threads = num_flush_threads
         self._chunk_size = max(1, chunk_size)
         self._metadata_codec = numcodecs.JSON()
@@ -614,7 +613,7 @@ class Individual(object):
 class SampleData(DataContainer):
     """
     SampleData(sequence_length=0, path=None, num_flush_threads=0, \
-    compressor=None, chunk_size=1024)
+    compressor=DEFAULT_COMPRESSOR, chunk_size=1024)
 
     Class representing input sample data used for inference.
     See sample data file format :ref:`specifications <sec_file_formats_samples>`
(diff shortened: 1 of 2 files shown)


jeromekelleher:
I don't see what the difference is here: aren't you still passing a ``compressor=None`` value ultimately?

Using ``something is False`` is pretty nasty, and I'm not sure it's guaranteed to work (is False guaranteed to be a singleton by the language?). It doesn't make sense to me, anyway, as an interface choice. It's a bug if the compressor wasn't being set to no compression as defined, but I don't think we should change the interface.

hyanwong:
(I guess if this isn't obvious, I should nevertheless code up the other bug correction in a different PR)

jeromekelleher:
> If we want to use None as the "force-no-compression" option, then to keep the existing behaviour we'd need to change the default to some other value - something which means "use None if this object is memory stored, but use the default if storing in a file". Happy to take any suggestions...

Ah, now I see what I was thinking. We don't want to turn on compression if we're using memory. We should probably just not worry about it; if we're using memory it's probably a tiny example anyway and most likely just used in tests. We can update some of the heavier tests to set ``compressor=None`` if it turns out to be slowdown. So yeah, let's just change it like I described above.

hyanwong:
You mean we set `compressor=DEFAULT_COMPRESSOR` in __init__, and if we set it delibrately to `None`, it means "don't compress"? If we do that, then when we use `path=None`, we will end up using DEFAULT_COMPRESSOR too, unless we specify otherwise. Is that OK then, do you think?

FWIW, this is one thing that works better in R: you can set the default using logic and other parameters, but in python you can't, AFAIK, use

`def __init__(
            self, path=None, compressor=DEFAULT_COMPRESSOR if path else None):

hyanwong:
OK, this is now done, but is based on #203 which is needed to pass the new tests, so I'll rebase once that is merged,

hyanwong:
@jeromekelleher: I think this is ready to merge now (I can't see why/where the coverage is failing)

tsinfer/formats.py
```diff
@@ -693,9 +692,8 @@ class SampleData(DataContainer):
         to use for compressing data. Any codec may be used, but
         problems may occur with very large datasets on certain codecs as
         they cannot compress buffers >2GB. If None, do not use any compression.
-        By default, use the :class:`numcodecs.zstd.Zstd` codec is used
-        when data is written to a file, and no compression when data is
-        stored in memory.
+        Default=``DEFAULT_COMPRESSOR``, currently defined as the
```
jeromekelleher: I don't think DEFAULT compressor is actually defined externally, so just leave it as "By default, use the :class:`numcodecs.zstd.Zstd` codec is used".

hyanwong: OK, just changed to simply 
```
Default=:class:`numcodecs.zstd.Zstd`.
```

PR #203, opened 2019-10-30
Don't close the data if this is not a file

hyanwong:
Ported from #202 : this avoids bombing out with `AttributeError: 'DictStore' object has no attribute 'close'` when using a context manager with path==None.

It mirrors what is already done on [line 389](https://github.com/tskit-dev/tsinfer/blob/19a5204dcdbdb9fa55dcc81ca45412750cef67b8/tsinfer/formats.py#L389)

### DIFF
--- a/tests/test_formats.py
+++ b/tests/test_formats.py
@@ -130,10 +130,10 @@ class TestSampleData(unittest.TestCase, DataContainerMixin):
 
     def test_defaults_no_path(self):
         ts = self.get_example_ts(10, 10)
-        input_file = formats.SampleData(sequence_length=ts.sequence_length)
-        self.verify_data_round_trip(ts, input_file)
-        for _, array in input_file.arrays():
-            self.assertEqual(array.compressor, None)
+        with formats.SampleData(sequence_length=ts.sequence_length) as sample_data:
+            self.verify_data_round_trip(ts, sample_data)
+            for _, array in sample_data.arrays():
+                self.assertEqual(array.compressor, None)
 
     def test_from_tree_sequence(self):
         ts = self.get_example_ts(10, 10)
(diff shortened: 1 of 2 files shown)


hyanwong:
The other (and possibly more Pythonic) way to do this is simply to call `.close()` and catch the exception if a `close()` method doesn't exist. I slightly prefer this. Would you prefer me to code it up that way, @jeromekelleher ?

jeromekelleher:
> The other (and possibly more Pythonic) way to do this is simply to call .close() and catch the exception if a close() method doesn't exist. I slightly prefer this. Would you prefer me to code it up that way, @jeromekelleher ?

I prefer it this way --- it's more readable and robust. Catching exceptions gets tricky --- if you caught an AttributeException there, for example, it would be very easy to add in bugs to the code by accident.

hyanwong:
OK. This can be merged then, I think.

jeromekelleher:
The branching is a bit of a mess [here](https://github.com/tskit-dev/tsinfer/network), I can't tell if this is forking from master or something else. Can you rebase this off upstream/master please?

hyanwong:
Yes, sorry - too many interdependent branches. I thought I had rebased off upstream/master, but will try again in case I forgot.

hyanwong:
I think I may have rebased before, but forgotten to force push it back. Done now. Let me know if you prefer both the (trivial) change and the adjusted test to be merged into a single commit for this PR, rather than 2 commits.

hyanwong:
> Let me know if you prefer both the (trivial) change and the adjusted test to be merged into a single commit for this PR, rather than 2 commits.

Scratch that - I've squashed them together as the network graph still looks weird even after rebasing off upstream/master (something I don't understand there).  Looks fine now that it's squashed, though.

jeromekelleher:
Great, thanks.

PR #204, opened 2019-10-30
wrap finalising in try/finally

### LINKED ISSUE #201: On error, SampleData context manager does not close the data stream
One thing stopping us testing on windows is that, when `__exit__()` is called on a SampleData object, and it fails (e.g. because we are out of memory), then we don't end up calling `.close()` on the `data.store` within the object, so the filehandle remains open. I guess it would be neater to close it on error, somehow?

hyanwong:
Fixes #201

### DIFF
--- a/tsinfer/formats.py
+++ b/tsinfer/formats.py
@@ -384,24 +384,26 @@ class DataContainer(object):
         provenance for the current operation. The specified 'command' is used
         to fill the corresponding entry in the provenance dictionary.
         """
-        self._check_write_modes()
-        self.data.attrs[FINALISED_KEY] = True
-        if self.path is not None:
-            store = self.data.store
-            store.close()
-            logger.debug("Fixing up LMDB file size")
-            with lmdb.open(
-                    self.path, subdir=False, lock=False, writemap=True) as db:
-                # LMDB maps a very large amount of space by default. While this
-                # doesn't do any harm, it's annoying because we can't use ls to
-                # see the file sizes and the amount of RAM we're mapping can
-                # look like it's very large. So, we fix this up so that the
-                # map size is equal to the number of pages in use.
-                num_pages = db.info()["last_pgno"]
-                page_size = db.stat()["psize"]
-                db.set_mapsize(num_pages * page_size)
-            # Remove the lock file as we don't need it after this point.
-            remove_lmdb_lockfile(self.path)
+        try:
+            self._check_write_modes()
+            self.data.attrs[FINALISED_KEY] = True
+        finally:
+            # Always close the file if necessary
+            if self.path is not None:
+                self.data.store.close()
+                logger.debug("Fixing up LMDB file size")
+                with lmdb.open(
+                        self.path, subdir=False, lock=False, writemap=True) as db:
+                    # LMDB maps a very large amount of space by default. While this
+                    # doesn't do any harm, it's annoying because we can't use ls to
+                    # see the file sizes and the amount of RAM we're mapping can
+                    # look like it's very large. So, we fix this up so that the
+                    # map size is equal to the number of pages in use.
+                    num_pages = db.info()["last_pgno"]
+                    page_size = db.stat()["psize"]
+                    db.set_mapsize(num_pages * page_size)
+                # Remove the lock file as we don't need it after this point.
+                remove_lmdb_lockfile(self.path)
         self._open_readonly()
 
     def _check_format(self):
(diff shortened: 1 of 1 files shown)


hyanwong:
#198 will create tests for this

tsinfer/formats.py
```diff
@@ -1183,22 +1185,23 @@ def add_site(
...
+                elif self._build_state == self.ADDING_SITES:
+                    self._sites_writer.flush()
+                if self.num_sites == 0:
+                    raise ValueError("Must add at least one site")
+                self._build_state = -1
+                if self.sequence_length == 0:
+                    # Need to be careful that sequence_length is JSON serialisable here.
+                    self.data.attrs["sequence_length"] = float(self._last_position) + 1
+        finally:
+            super(SampleData, self).finalise()
```
jeromekelleher: I don't think this is the way to do it --- we should only be finalising (marking the file as complete and well-formed) if things succeed.

jeromekelleher:
I've had a look at this, and it's tricky to do. The logic is around closing/finalising is pretty convoluted (driven in part by the semantics of the different Zarr stores, I think). It probably needs to be rethought fully to do it well. The best thing to do for now, I think, is to figure out exactly where the problem is occuring for #201, and wrap just that line with a try/except or something.

jeromekelleher:
> Or do you mean look for the place it fails in the format.py file, which is what I think I did (the failure happens during finalise())

Yes, we wrap whatever *external* function is failing. Finalise is quite complicated, and we *don't* want to mark something as "OK, this file is good" if some failure has happened. Putting in too broad exception catching can result in really nasty bugs. If you paste in the exception back trace I can tell you where to put the try/except.

hyanwong:
> If you paste in the exception back trace I can tell you where to put the try/except.

Will do. Note, however, that what I put in was a `try/finally` (not try/except) clause, so shouldn't (I hope) disguise many bugs.

hyanwong:
So an example of the test that fails is:

```
    def test_too_small_max_file_size_add(self):
        with tempfile.TemporaryDirectory(prefix="tsinf_format_test") as tempdir:
            base_size = 2**16  # Big enough to allow the initial file to be created
            # Fail during adding a large amount of data
            with self.assertRaises(lmdb.MapFullError):
                filename = os.path.join(tempdir, "samples.tmp")
                with formats.SampleData(path=filename, sequence_length=1,
                                        max_file_size=base_size) as small_sample_file:
                    small_sample_file.add_site(
                        0, alleles=['0', '1'],
                        genotypes=np.zeros(base_size, dtype=np.int8))
```

I can stop the failure by adding

```
            # Work around https://github.com/tskit-dev/tsinfer/issues/201
            small_sample_file.data.store.close()
```

If the line is absent, the failure (windows only) is something like

```
Traceback (most recent call last):
  File "C:\Users\nicky\yan\tsinfer-hyanwong\tests\test_formats.py", line 161, in test_too_small_max_file_size_add
    genotypes=np.zeros(base_size, dtype=np.int8))
  File "C:\Users\nicky\yan\Miniconda3\lib\tempfile.py", line 805, in __exit__
    self.cleanup()
  File "C:\Users\nicky\yan\Miniconda3\lib\tempfile.py", line 809, in cleanup
    _shutil.rmtree(self.name)
  File "C:\Users\nicky\yan\Miniconda3\lib\shutil.py", line 516, in rmtree
    return _rmtree_unsafe(path, onerror)
  File "C:\Users\nicky\yan\Miniconda3\lib\shutil.py", line 400, in _rmtree_unsafe
    onerror(os.unlink, fullname, sys.exc_info())
  File "C:\Users\nicky\yan\Miniconda3\lib\shutil.py", line 398, in _rmtree_unsafe
    os.unlink(fullname)
PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Users\\nicky\\AppData\\Local\\Temp\\tsinf_format_test8vktklxn\\samples.tmp'
```

The problem here is that the *test error* is not raised during the __exit__ method of the SampleData object, but when we finally try to delete the file. We (presumably) want to catch the `lmdb.MapFullError` (and in fact, any non-completion errors), e.g. during the __exit__ method, and force a `self.data.store.close()` (and maybe remove the lockfile)

jeromekelleher:
This is tricky, and I think the ``__close__`` logic need to be reworked. Let's just put in the workaround into the tests as you suggest above, and keep the issue open.

hyanwong:
Right. I've just realised that the problem might remain after that anyway, as even with the context manager, we (deliberately) leave the file open (in read-only mode) after exiting the `with` block. So on Windows we need to close the file (and I guess it's good practice even on other OSes). At the moment I would do this using e.g. `sample_data.close()`, but that fails if this is a memory mapped file. I can either check whether `path is None` in the tests, or I can stick that test in the [`close()` method](https://github.com/tskit-dev/tsinfer/blob/65f3e70d4935665d241b6766560d51085ed33cee/tsinfer/formats.py#L336) of the DataContainer, so that it is no longer an error to call close() on a memory-mapped SampleData or AncestorData file. Which would you prefer?

jeromekelleher:
Let's work around it in the test case and keep the issue open.

hyanwong:
> Let's work around it in the test case and keep the issue open.

OK. Most of the discussion is in this PR. I guess I should move some of it into the issue itself, and close this PR

PR #206, opened 2019-10-31
Allow non-biallelic alleles to be input and mapped

hyanwong:
This simply marks them as not for inference, and uses the map_mutations method to place the mutations so as to re-create the correct haplotypes

### DIFF
--- a/docs/inference.rst
+++ b/docs/inference.rst
@@ -23,10 +23,10 @@ Data requirements
 Input haplotype data for tskit must satisfy the following requirements:
 
 - Data must be *phased*.
-- For each site we must know the *ancestral state*. Note that this is
+- For each site used for inference we must know the *ancestral state*. Note that this is
   not necessarily the same as the REF column from VCF, and ancestral
   states must be computed using some other method.
-- Only biallelic sites are currently supported.
+- Only biallelic sites can be used for inference.
 
 
 .. _sec_inference_data_model:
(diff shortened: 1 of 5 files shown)


hyanwong:
Closing and reopening because of a temporary appveyor glitch

hyanwong:
This is ready for review. There is only one added test, but that might be considered enough?

jeromekelleher:
Very nice, but could use a bit more testing. What are the limits here, i.e. what happens if there's lots of alleles? There must be some other corner cases that need exploring.

hyanwong:
Yep, as well as the limit of large numbers of alleles, we should also check interaction with `from_tree_sequence` (c.f. #211) I suspect. Ping other suggestions in this thread and I can try to implement something.

Note, however, that since we are simply applying `map_mutations`, I would hope that some of the cases such as large allele numbers would be mostly checked in the tests for that method.

jeromekelleher:
> Note, however, that since we are simply applying map_mutations, I would hope that some of the cases such as large allele numbers would be mostly checked in the tests for that method.

Sure, it's just good to locally characterise the behaviour so we don't run into suprises later.

hyanwong:
I've added a test for 64 alleles (passes) and 65 alleles (expected to fail because `map_mutations()` only allows 64 alleles), a test if we only have a subset of the alleles indexes in `genotypes`, and also a test that `from_tree_sequence()` works where we have multiple mutations at a site. Happy to add other suggestions, otherwise this can probably be merged.

jeromekelleher:
This is excellent, merging.

PR #216, opened 2019-11-25
Allow pip install of dev requirements even on non-linux

hyanwong:
Since prctl and numa are optional, and prctl doesn't even work on non-linux systems, only install these on linux

### DIFF
--- a/requirements/development.txt
+++ b/requirements/development.txt
@@ -13,9 +13,9 @@ lmdb
 attrs
 # Only needed for the Python implementation.
 sortedcontainers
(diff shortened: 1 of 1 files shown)


hyanwong:
If we want to test this, I could add `python3 -m pip install -r requirements/development.txt` to the `.travis.yml` file, and e.g. build the docs there too, although that might slow down CI testing.

jeromekelleher:
LGTM, didn't know about the marker syntax in pip, that's handy.

PR #219, opened 2020-01-16
Remove kc distance

jeromekelleher:
Use upstream version in tskit instead.


hyanwong:
Great that it's now in tskit. LGTM

PR #235, opened 2020-03-02
Add tests when there are different positions for a breakpoint

hyanwong:
As discussed in https://github.com/tskit-dev/tsinfer/pull/231#issuecomment-593362196, but there's obviously something I don't understand here, as all 3 tests I've implemented are failing, and I'd expect at least one of the first two tests to pass (the third test won't pass while we prune out the non-inference sites to make an ancestors TS, but might be expected to pass when we remove the idea of an ancestors TS entirely).

Essentially I'm sticking a (non informative) site (site 1) between two conflicting sites, so that the breakpoint could either be before site 1 or after site 1. I would hope that the new algorithm would place the recombination in the longer section of genome, but it doesn't. Perhaps I need to do this by setting the new `mutation_rate=0.0`, or I don't properly understand the likelihood compression on trees.

I've made this into a PR so at least we have an example to talk about, @jeromekelleher

### DIFF
--- a/tests/test_inference.py
+++ b/tests/test_inference.py
@@ -2158,3 +2158,45 @@ class TestSequentialAugmentedAncestors(TestAugmentedAncestors):
             self.assertEqual(expected_sample_ancestors, num_sample_ancestors)
             tsinfer.verify(samples, final_ts.simplify())
             ancestors_ts = augmented_ancestors
+
+
+class TestAlgorithmResults(unittest.TestCase):
+    """
+    Some features of the algorithm have expected outcomes in simple cases. Test these.
+    """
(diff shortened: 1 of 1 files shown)


jeromekelleher:
Ah. We're not actually weighting by distance here @hyanwong (which we should, I'll open an issue). Can you do these tests by manually setting the recombination rates? rho[j] is the recomb rate between site j and the previous site (rho[0] is ignored/must be 0).

hyanwong:
OK. Set the recombination rate manually and it works (I've skipped the one with a non-inference site in, as I think it should work once we have non-inference sites in the ancestors TS). Hurrah.

jeromekelleher:
Looks good, thanks.

Issue #238, opened 2020-03-04
Add documentation link to tskit for analysis


hyanwong:
At the moment the docs give examples of how to do inference, but then a novice user has no way to know how to analyse their data. I suggest it might be good to have a (very short) `analysis.rst` file that gives a single example (e.g. with GNN or Fst) and points to the statistics page in tskit for more information. What do you think @jeromekelleher ?

jeromekelleher:
Good idea, go for it!

hyanwong:
Great, will do, and consult with @markravinet about what he might find a useful example, as a new user.

jeromekelleher:
Closed in #239

PR #239, opened 2020-03-05
Add example of reading a VCF to the tutorial

hyanwong:
Would be nicer to add the VCF to this repo, rather than rely on another source. @markravinet is producing once for us with ancestral allele states (although the URL and the code as-written should be OK still). Perhaps hold off merging until I change the URL to whatever we decide upon.


docs/conf.py
```diff
@@ -196,7 +196,7 @@ def __getattr__(cls, name):
 # Example configuration for intersphinx: refer to the Python standard library.
 intersphinx_mapping = {
     'https://docs.python.org/3/': None,
-    'https://tskit.readthedocs.io/en/stable/': None,
```
hyanwong: Need to temporarily change to the latest tskit docs to have the "trees_and_tree_sequences" section link.

hyanwong:
This is now ready for review @jeromekelleher, although until it is merged, the file `https://github.com/tskit-dev/tsinfer/raw/master/docs/_static/P_dom_chr24_phased.vcf.gz` won't exist, so you'll have to substitute a different URL (e.g. `https://github.com/hyanwong/tsinfer/raw/add-analysis-docs/docs/_static/P_dom_chr24_phased.vcf.gz`) into the demonstration script to get it working.

It's taken a while to write, but I think it's worth it. I can ask @MarkRavinet if it seems helpful from a user POV.

docs/tutorial.rst
```diff
@@ -306,17 +318,281 @@ trees. In this example we incorrectly inferred that 0 coalesces with the
...
+    import tsinfer
+    
+    def add_diploid_sites(vcf, samples):
+        pos = None
+        for variant in vcf:
+            if pos == variant.POS:
+                raise ValueError("Duplicate positions for variant at position", pos)
+            else:
+                pos = variant.POS
+            if any([row[2]==False for row in variant.genotypes]):
```
jeromekelleher: Comparing things to False is a bad idea. A more idiomatic thing to do is ``any(not row[2] for row in ...]). It's not obvious why this would mean it's unphased though. An explanation or a link to the docs would be good (people will ask us questions otherwise...)

docs/tutorial.rst
```diff
@@ -306,17 +318,281 @@ trees. In this example we incorrectly inferred that 0 coalesces with the
...
+                raise ValueError("Unphased genotypes for variant at position", pos)
+            alleles = [variant.REF] + variant.ALT
+            # Sort with the ancestral allele first if we have one (otherwise use REF)
+            ancestral = variant.INFO.get('AA', variant.REF)
```
jeromekelleher: Why not
```python
ancestral = variant.INFO.get('AA', variant.REF)
all_alleles = set(variant.ALT + [variant.REF])
# Ancestral state must be first in the allele list.
alleles = [ancestral] + list(all_alleles - {ancestral})
# Map allele values to their indexes in the alleles list for each sample.
genotypes = [alleles.index(allele) for row in variant.genotypes for allele in row[0:2]]
```
All the sorting stuff is obscuring the point of the operations, to me.

docs/tutorial.rst
```diff
@@ -306,17 +318,281 @@ trees. In this example we incorrectly inferred that 0 coalesces with the
...
+        # In this VCF, the first letter of the sample name refers to the population
+        pop_lookup = {'8': 'Norway', 'F': 'France'}
+        sample_populations = [pop_lookup[sample_name[:1]] for sample_name in vcf.samples]
+        name_to_pop_id = {}
+        for name in set(sample_populations):
+            pop_id = samples.add_population(metadata={'country': name})
+            name_to_pop_id[name] = pop_id
+        return {i: name_to_pop_id[name] for i, name in enumerate(sample_populations)}
+    
+    def add_diploid_individuals(vcf, samples, populations={}):
```
jeromekelleher: Never use a mutable object as a function parameter [default](https://florimond.dev/blog/articles/2018/08/python-mutable-defaults-are-the-source-of-all-evil/).

This code should **only** work for the example we're giving and be absolutely no more general, so I don't think it should have any default.

docs/tutorial.rst
```diff
@@ -306,17 +318,281 @@ trees. In this example we incorrectly inferred that 0 coalesces with the
...
-    1. Add documentation links for msprime above so we can explain tree
-       sequences there.
+.. code-block:: python
 
-++++++++++++
-Data example
-++++++++++++
+    import json
+
+    def add_populations(vcf, samples):
```
jeromekelleher: We already know what the populations are, so this is way more complicated than it needs to be:
```
def add_populations(vcf, samples):
    """
    Adds tsinfer Population objects and returns a list of IDs corresponding to the VCF samples.
    """
    # In this VCF, the first letter of the sample name refers to the population
    pop_lookup['8'] = samples.add_population(metadata={'country': 'Norway'})
    pop_lookup['F'] = samples.add_population(metadata={'country': 'France'})
    return [pop_lookup[sample_name[0]] for sample_name in vcf.samples]
```

docs/tutorial.rst
```diff
@@ -306,17 +318,281 @@ trees. In this example we incorrectly inferred that 0 coalesces with the
...
+        name_to_pop_id = {}
+        for name in set(sample_populations):
+            pop_id = samples.add_population(metadata={'country': name})
+            name_to_pop_id[name] = pop_id
+        return {i: name_to_pop_id[name] for i, name in enumerate(sample_populations)}
+    
+    def add_diploid_individuals(vcf, samples, populations={}):
+        for i, name in enumerate(vcf.samples):
+            samples.add_individual(
+                ploidy=2, metadata={"name": name}, population=populations.get(i, None))
```
jeromekelleher: Using the code above with the lists:
```
for name, population in zip(vcf.samples, populations):
     samples.add_individual(ploidy=2, metadata={"name": name}, population=population)
```

docs/tutorial.rst
```diff
@@ -306,17 +318,281 @@ trees. In this example we incorrectly inferred that 0 coalesces with the
...
+    for n in sparrow_ts.samples():
+        population_data = sparrow_ts.population(sparrow_ts.node(n).population)
+        colours_for_node[n] = colours[json.loads(population_data.metadata)['country']]
+    
+    individual_for_node = {}
+    for n in sparrow_ts.samples():
+        individual_data = sparrow_ts.individual(sparrow_ts.node(n).individual)
+        individual_for_node[n] = json.loads(individual_data.metadata)['name']
+    
+    sparrow_ts.at(1e6).draw(path="tree_at_1Mb.svg", height=700, width=1200,
```
jeromekelleher: Clearer:
```
tree = sparrow_ts.at(1e6)
tree.draw(...)

File: docs/tutorial.rst [hunk 8]
```diff
@@ -306,17 +318,281 @@ trees. In this example we incorrectly inferred that 0 coalesces with the
...
+the allele or site frequency spectrum (SFS) can be calculated using
+:meth:`tskit.TreeSequence.allele_frequency_spectrum` and the allelic diversity ("Tajima's
+:math:`{\pi}`") using :meth:`tskit.TreeSequence.diversity`, both of which can also be
+calculated locally (e.g. :ref:`per tree or in genomic windows <sec_stats_windows>`). As
+a basic example, here's how to calculate genome-wide :math:`F_{st}` between the Norwegian
+and French (sub)populations:
+
+.. code-block:: python
+
+    sample_node_objs = [sparrow_ts.node(n) for n in sparrow_ts.samples()]
```
jeromekelleher: Simpler:
```
samples_by_pop = [sparrow_ts.samples(population=pop_id) for pop_id in range(sparrow_ts.num_populations)]
```

docs/tutorial.rst
jeromekelleher: If cyvcf2 just returns a tuple, then something like
``any(not phased for _, _, phased in variant.genotypes)`` would be better. Hopefully it returns a namedtuple or something though?

hyanwong: Sadly, in the version of cyvcf2 that I have, the `.genotypes` attribute for a set of N diploid samples simply contains an N length list of size 3 lists. I'll use your `_,_,phased` notation, thanks.

docs/tutorial.rst
hyanwong: Yes, this is nicer. Thanks.

hyanwong: Although the VCF (and CYVCF2) seems to list `0`s and `1`s, not e.g. `A`s and `T`s so we don't want to be using `.index`. Rather, we want to make a list of the variants, and reorder somehow. I'll see what I can do.

hyanwong: So I think this is how you need to do it using your approach:

```
            all_alleles = [variant.REF] + variant.ALT
            # Ancestral state must be first in the allele list.
            alleles = [ancestral] + list(set(all_alleles) - {ancestral})
            allele_index = {old_index: alleles.index(allele) for old_index, allele in enumerate(alleles)}
            # Map original allele indexes to their indexes in the new alleles list.
            genotypes = [allele_index[old_index] for row in variant.genotypes for old_index in row[0:2]]
```

docs/tutorial.rst
hyanwong: OK. Good tip, thanks.

docs/tutorial.rst
hyanwong: Of course, I forgot you could provide a `population` param to `samples()`. But I do use the list of node objects later, in e.g.

`sample_pops = [json.loads(sparrow_ts.population(n.population).metadata)['country'] for n in sample_node_objs]`

I'll need to rework these.

hyanwong:
> I've made a good few comments on the code, as I think it's important it's as simple and clear as possible. 

Absolutely. These are really useful suggestions, thanks. I've made the changes and pushed as a separate commit, so you can see what I've done (basically all you suggested, with one extra var for clarity, and some logic corrections around allele indices, as I noted in a reply to a comment above).

I also added a tqdm bar to the vcf reader, as I think this is a useful thing for people to see (and as the VCF contains the sequence_length, we can use that as the total (and report the prograssbar in base pairs), rather than the hacky way we previously uses where we had to use bcftools to get the number of sites.

I reckon this is mergable now.

docs/tutorial.rst
```diff
@@ -306,17 +318,303 @@ trees. In this example we incorrectly inferred that 0 coalesces with the
...
+
+    def add_diploid_sites(vcf, samples):
+        """
+        Loop over sites in the vcf and add them to the samples object, reordering the
+        alleles to put the ancestral allele first, if it is available.
+        """
+        progressbar = tqdm.tqdm(total=samples.sequence_length, desc="Read VCF", unit='bp')
+        pos = 0
+        for variant in vcf:
+            if pos and pos == variant.POS:
```
jeromekelleher: Why would pos be False here? I don't get the first condition.

docs/tutorial.rst
```diff
@@ -306,17 +318,303 @@ trees. In this example we incorrectly inferred that 0 coalesces with the
...
+                raise ValueError("Duplicate positions for variant at position", pos)
+            else:
+                progressbar.update(variant.POS - pos)
+                pos = variant.POS
+            if any([not phased for _, _, phased in variant.genotypes]):
+                raise ValueError("Unphased genotypes for variant at position", pos)
+            alleles = [variant.REF] + variant.ALT
```
jeromekelleher: The ``alleles`` variable is unused now, and the comment doesn't make sense anymore I think. _

docs/tutorial.rst
```diff
@@ -306,17 +318,303 @@ trees. In this example we incorrectly inferred that 0 coalesces with the
...
+            allele_index = {old_index: alleles.index(allele)
+                for old_index, allele in enumerate(all_alleles)}
+            # Map original allele indexes to their indexes in the new alleles list.
+            genotypes = [allele_index[old_index]
+                for row in variant.genotypes for old_index in row[0:2]]
+            samples.add_site(pos, genotypes=genotypes, alleles=alleles)
+        progressbar.close()
+
+    def chromosome_length(vcf):
+        try:
```
jeromekelleher: This is too general --- we're only writing a script to support just this one VCF file. Every bell and whistle that's added makes it that bit more difficult for people to understand, and we want to be totally and absolutely explicit that this is just a **guide** for writing your own parser.

docs/tutorial.rst
```diff
@@ -306,17 +318,303 @@ trees. In this example we incorrectly inferred that 0 coalesces with the
...
+            if len(vcf.seqlens) > 1:
+                raise ValueError("More than one contig length listed in the VCF file")
+            return vcf.seqlens[0]
+        except (AttributeError, IndexError):
+            print("No chromosome length in VCF file, percent progress hidden", flush=True)
+            return 0
+
+    # URL for the VCF
+    url = ("https://github.com/tskit-dev/tsinfer/raw/master/docs/_static/P_dom_chr24_phased.vcf.gz")
```
jeromekelleher: What are the brackets for?

docs/tutorial.rst
```diff
@@ -306,17 +318,303 @@ trees. In this example we incorrectly inferred that 0 coalesces with the
...
+    import cyvcf2
+    import tsinfer
+    import tqdm
+
+    def add_diploid_sites(vcf, samples):
+        """
+        Loop over sites in the vcf and add them to the samples object, reordering the
+        alleles to put the ancestral allele first, if it is available.
+        """
+        progressbar = tqdm.tqdm(total=samples.sequence_length, desc="Read VCF", unit='bp')
```
jeromekelleher: I don't think we need the progressbar --- it's not necessary here for reading this VCF.

jeromekelleher:
Looks good, but I think we're still making the VCF parsing too complicated. This should be ruthlessly simple so that people grasp the basic ideas. The code won't work on arbitrary VCFs anyway without modifications, so it's better to keep it simple so they understand the important bits like allele mapping, ancestral states, etc. If they are going to write a VCF to suit the format here, then they might as well include all the stuff that's needed.

docs/tutorial.rst
hyanwong: Oh, sorry, yes.

docs/tutorial.rst
hyanwong: We should allow the first `variant.POS` to be 0 without raising an error. Perhaps clearer if I say `if variant.POS != 0 and pos == variant.POS:` ?

docs/tutorial.rst
hyanwong: OK. Although I think that the version with a progressbar should be described or demoed *somewhere*. Both Mark and Wenhua are testing their files using some derivation of this script, and will try to apply it to their own VCFs, when the progressbar will come in handy. I added it because I wanted to test on the same samples using chromosome 8, and I couldn't work out if the script had hung or not. So I think most people would benefit from some sort of feedback. 

We could, of course, just *state* in the text that you can add a progressbar by getting the sequence length out of the VCF and updating the value based on the genomic position. That way we would keep it out of the demo code...

docs/tutorial.rst
hyanwong: Ah, left in when I had to multi-line the URL for length reasons. At the moment this line is over the 89 char limit (94 chars, actually), but it looks OK in the docs. Do we want to enforce the 89 char limit in Python code in the docs too?

hyanwong:
> Looks good, but I think we're still making the VCF parsing too complicated. This should be ruthlessly simple so that people grasp the basic ideas. The code won't work on arbitrary VCFs anyway without modifications, so it's better to keep it simple so they understand the important bits like allele mapping, ancestral states, etc. If they are going to write a VCF to suit the format here, then they might as well include all the stuff that's needed.

Do you mean the sequence length + progressbar stuff? Presumably we do want to add the sequence length to the SampleData call, so the simplest would be

```
with tsinfer.SampleData(sequence_length=vcf.seqlens[0]) as samples:
```

But then if they have more than one sequence length in the VCF, the script will work, rather than fail noisily. So I thought you might want to assert that `len(vcf.seqlens) == 1` somewhere, although we could do that without a new function, I guess.

docs/tutorial.rst
hyanwong: See below re comments about setting the sequence_length, which I do think we want to do (and encourage), even in this basic guide version.

docs/tutorial.rst
jeromekelleher: 94 is fine for the docs I think.

hyanwong:
P.s. it seems a bit odd that the default sequence_length value for a SampleData file is 0 rather than `None`. Surely `None` better reflects the "don't know" state?

docs/tutorial.rst
jeromekelleher: Or we could add a separate snippet showing this.

jeromekelleher:
> But then if they have more than one sequence length in the VCF, the script will work, rather than fail noisily. So I thought you might want to assert that len(vcf.seqlens) == 1 somewhere, although we could do that without a new function, I guess.

What would it mean if there was more than one seqlen in the VCF? We can't possibly cover all quirks of input VCFs.

jeromekelleher:
> P.s. it seems a bit odd that the default sequence_length value for a SampleData file is 0 rather than None. Surely None better reflects the "don't know" state?

That sounds right. Perhaps it's something to do with the numeric encoding or something?

docs/tutorial.rst
jeromekelleher: Well, why not make pos = -1 or None initially rather than 0 then?

I'm not sure that pos can be zero here anyway, VCF is a 1 based coordinate system.

docs/tutorial.rst
jeromekelleher: I disagree - I think we should be explicit about what we expect the VCF to contain and to write code to process only that. Having extra stuff in there just gives a false sense of generality.

hyanwong:
> What would it mean if there was more than one seqlen in the VCF? We can't possibly cover all quirks of input VCFs.

The VCF can (I think) list different lengths for different contigs in the same file. I agree that we can't cover the quirks, but think we should at least try to fail if something looks wrong.

hyanwong:
> > P.s. it seems a bit odd that the default sequence_length value for a SampleData file is 0 rather than None. Surely None better reflects the "don't know" state?
> 
> That sounds right. Perhaps it's something to do with the numeric encoding or something?

Should I raise a GitHub issue about it?

docs/tutorial.rst
hyanwong: But that's what I mean: this particular VCF contains the sequence length (which is different from the last variant position), and I think we want to pass this as the `sequence_length` parameter when we make the SampleData file. It's merely a question of whether we have a function that does this, or if we pass it in direct as vcf.seqlen[0] and don't bother about ensuring that our script "fails noisily" on a bad (multi-length) VCF.

docs/tutorial.rst
hyanwong: Didn't know that VCF was 1-based. In which case this can be nixed.

docs/tutorial.rst
hyanwong: Can't see an easy way to add a snippet. I suggest:

```
.. note::
    Larger VCF files will take some time to parse, and you may wish to add some sort of
    progress indicator to this script, for example by inserting the line
    `progressbar = tqdm.tqdm(total=samples.sequence_length, desc="Read VCF", unit='bp')`
    at the top of the `add_diploid_sites` function, followed by
    `progressbar.update(variant.POS - pos)` as the first line of the variant loop within
    that function.
```

hyanwong: (p.s. I do note that you use a tqdm progressbar in the simulation example higher up in the file, so it's not unfamiliar territory in this tutorial)

hyanwong:
NB closes #238 - ping me when you want me to squash the commits into one, @jeromekelleher

jeromekelleher:
LGTM, thanks @hyanwong. Can you rebase and squash please?

hyanwong:
Done.

PR #240, opened 2020-03-10
Use tables natively throughout inference

jeromekelleher:
WIP


jeromekelleher:
Backing out of these changes as the low-level library updates and communication problems are very tricky. We can build new features on the existing infrastructure.

PR #244, opened 2020-03-12
Relax input restrictions on AncestorData.

jeromekelleher:
WIP

Allow for arbitrary patterns of variation in ancestral haplotypes and arbitrary mutations in the ancestors tree sequence.

### DIFF
--- a/dev.py
+++ b/dev.py
@@ -113,22 +113,46 @@ def tsinfer_dev(
         print("num_sites = ", ts.num_sites)
     assert ts.num_sites > 0
 
-    ts = msprime.mutate(ts, rate=1e-8, random_seed=seed,
-            model=msprime.InfiniteSites(msprime.NUCLEOTIDES))
+    # ts = msprime.mutate(ts, rate=1e-8, random_seed=seed,
+    #         model=msprime.InfiniteSites(msprime.NUCLEOTIDES))
 
     samples = tsinfer.SampleData.from_tree_sequence(ts)
     rho = recombination_rate
-    mu = 1e-3
+    mu = 1e-3 #1e-15
+
+
+#     num_alleles = samples.num_alleles(inference_sites=True)
+#     num_sites = samples.num_inference_sites
+#     with tsinfer.AncestorData(samples) as ancestor_data:
+#         t = np.sum(num_alleles) + 1
+#         for j in range(num_sites):
+#             for allele in range(num_alleles[j]):
+#                 ancestor_data.add_ancestor(j, j + 1, t, [j], [allele])
+#                 t -= 1
 
     ancestor_data = tsinfer.generate_ancestors(
         samples, engine=engine, num_threads=num_threads)
+
     ancestors_ts = tsinfer.match_ancestors(
         samples, ancestor_data, engine=engine, path_compression=True,
         extended_checks=False, recombination_rate=rho,
-        # FIXME mu must be zero for ancestor matching at the moment.
-        mutation_rate=0)
+        mutation_rate=mu)
+    # print(ancestors_ts.tables)
+    # print("ancestors ts")
+    # for tree in ancestors_ts.trees():
+    #     print(tree.draw_text())
+    #     for site in tree.sites():
+    #         if len(site.mutations) > 1:
+    #             print(site.id)
+    #             for mutation in site.mutations:
+    #                 print("\t", mutation.node, mutation.derived_state)
 
-    print(ancestors_ts.tables)
+
+
+    # for var in ancestors_ts.variants():
+    #     print(var.genotypes)
+
+    # print(ancestors_ts.tables)
 
     # ancestors_ts = tsinfer.augment_ancestors(samples, ancestors_ts,
     #         [5, 6, 7], engine=engine)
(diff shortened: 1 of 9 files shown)


hyanwong:
It's fantastic to see all this work on tsinfer. Let me know if any of it is relevant to my (presumably now defunct) PR for incorporating missing data into the inference pathway, and if so, if there is any help I can give (e.g. I have various unit tests written for this)

jeromekelleher:
> Let me know if any of it is relevant to my (presumably now defunct) PR for incorporating missing data into the inference pathway, and if so, if there is any help I can give (e.g. I have various unit tests written for this)

Your PR should still be relevant, and I'm planning to pull it in once I have support for matching on missing data in the copying process (trivial addition I think).

jeromekelleher:
This is ready to merge I think @hyanwong. It's switching the ancestor matching process to using the copying process natively to generate mutations and ignoring the focal sites (for now). The default values of rho and mu should reproduce the old behaviour, but maybe not. 

I'm planning on fixing the normalisation/precision issue (#246), adding support for missing data and then looping back to seeing how all this affects actual inference. It'd be great if you can take a look at how it's working at the moment.

hyanwong:
Will get onto this today. Just setting up a framework to test multiple versions of tsinfer against each other (think I've figured it out).

hyanwong:
Sorry I've taken so long to get onto this. In https://github.com/hyanwong/tsinfer-benchmarking I now have a very rough framework for being able to load different versions of tsinfer (by commit hash) into a running python session, to make it easier to compare results.

The take-home message (although only from some simple testing at the moment, with msprime 30-sample simulations, Ne=1 length=1, mu=10, rho=10), is that the new code (commit 03ad4bd) is *slightly worse* than the old (0.1.4 = efbafff) version when run in a vanilla way, but better as soon as the recombination scaled to genetic distance is used. I haven't yet incorporated seq error (inference "mutation_rate"), or done more complex models, although this should now be easy.

If you run `python3 simulation.py` in the repo, it should (if it hasn't been run before) download 2 copies of the tsinfer github repo into folders in `_versions`, check out different git commits, and compile. Then it runs a comparison on a set of replicate tree sequences from `msprime.simulate(30, recombination_rate=10, mutation_rate=10)` with the following results (the important thing is the first column in each case, where negative means worse performance by the new code):

```
== efbafff-03ad4bd diffs with positional info ==
           mean_diff  stderr_diff  mean_efbafff  stderr_efbafff  mean_03ad4bd  stderr_03ad4bd
time       -0.030388     0.009440      0.103121        0.005079      0.133509        0.007354
num_edges  29.593220     1.843072    254.067797        6.345353    224.474576        4.938805
kc          1.163697     0.209559     26.506895        0.398050     25.343198        0.365243

== efbafff-03ad4bd diffs without positional info ==
           mean_diff  stderr_diff  mean_efbafff  stderr_efbafff  mean_03ad4bd  stderr_03ad4bd
time       -0.013907     0.009765      0.135445        0.008355      0.149352        0.008590
num_edges -14.372881     1.736246    254.067797        6.345353    268.440678        6.600315
kc         -0.629668     0.213411     26.506895        0.398050     27.136563        0.433874
```

hyanwong:
Actually, clearer if we simply show the percentage change of new codebase vs old (100 = no change)

```
== efbafff vs 03ad4bd with positional info ==
           percent_change    stderr
time            78.280656  1.306972
num_edges      112.843469  0.634854
kc             104.721798  0.838316
== efbafff vs 03ad4bd without positional info ==
           percent_change    stderr
time            73.149012  1.893676
num_edges       94.756844  0.638864
kc              97.919000  0.751962
```

jeromekelleher:
Great, thanks Yan, excellent info.

hyanwong:
Should I look in more detail about *why* we do worse when using the vanilla params (i.e. not accounting for distance). I had expected that we should do exactly the same. I can easily check performance of all the previous commits.

jeromekelleher:
> Should I look in more detail about why we do worse when using the vanilla params (i.e. not accounting for distance). I had expected that we should do exactly the same. I can easily check performance of all the previous commits.

Please, yes, it would be helpful to know where the regression happened and what we're doing wrong.

hyanwong:
You recent commit aa44116685d13b76f51753deef38ea432f159a13 seems to have more-or-less restored the original behaviour. Is that expected?

```
== efbafff vs aa44116 without positional info ==
           percent_change    stderr
time            67.699667  1.553598
num_edges       99.009031  0.569107
kc              99.340182  0.805254
```

jeromekelleher:
> You recent commit aa44116 seems to have more-or-less restored the original behaviour. Is that expected?

No, but, great!

hyanwong:
Bisecting the issue reveals that bb0bbddcf47e71f41c9c48ed021d995251f42b38 is the first bad commit. Not sure if that's very helpful, @jeromekelleher ? Do you want me to drill down more?

jeromekelleher:
Let's not worry about it, better to see how we're doing more generally and get infrastructure in place.

hyanwong:
Yep - it was useful to get the machinery in place for doing `git bisect` though.

Issue #209, opened 2019-11-06
Create new samples file with a subset of sites


hyanwong:
It's useful in a few instances to subset the sites in a SampleData file down to a limited number. Since we haven't got methods to iterate through e.g. the populations in a SampleData file, it's hard to recreate the file exactly (but with a subset of sites). We *can* do this, I think, by resizing the underlying zarr arrays, something like this:

```
def subset(sample_data, keep_sites_bool):
    site_shape = np.sum(keep_sites_bool).item(),
    subset = sample_data.copy()
    for attr in [a for a in dir(sample_data) if a.startswith('sites_')]:
        zarr_name = attr.replace('_', '/', 1)
        subset.data[zarr_name].resize(*(site_shape +  sample_data.data[zarr_name].shape[1:]))
        subset.data[zarr_name].set_basic_selection(..., sample_data.data[zarr_name][:][select])
    subset.finalise()
    return subset
```

Is this worth adding to tsinfer?

hyanwong:
Or perhaps more to Jerome's taste, we could `clear()` the sites arrays in the copy and add sites back in incrementally using add_sites - presumably slower by less hacky (and doesn't subset a copy of the genotypes array in memory)

hyanwong:
FWIW running `subset()` on chr20, 1000G, subsetting doen to about 1/3 of the 800,000 sites takes about 40 sec, whereas running `subset2()` takes about 3 mins.

hyanwong:
I've just needed to do this again, without realising that I coded a solution previously. The code is neater if the zarr groups are used directly, and np.take is used. The version below creates a new SampleData object, instead of acting on an old one:


```
def keep_sites(sample_data, keep_sites, sequence_length=None, **kwargs ):
    """
    Return a new sample data object keeping only the specified sites.
    
    :param arraylike keep_sites: An array of sites to keep, either a list of
        site indices or a numpy boolean array of size sample_data.num_sites
    :param float sequence_length: A new sequence length for the returned
        SampleData object. If None (the default), take the existing sequence_length
    :return: A new :class:`.SampleData` object with a reduced number of sites
    :rtype: :class:`.SampleData`.
    """
    new_sample_data = sample_data.copy(**kwargs)
    new_sample_data._check_edit_mode()
    for name, arr in sample_data.data.sites.arrays():
        new_data = np.take(arr[:], keep_sites, axis=0)
        arr.resize(*new_data.shape)
        arr[:] = new_data
[...]

PR #251, opened 2020-03-19
Document inherited methods in SampleData

benjeffery:
@hyanwong Here's the option in sphinx to document inherited members.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -12,11 +12,12 @@ File formats
 
 .. autoclass:: tsinfer.SampleData
     :members:
+    :inherited-members:
 
 .. todo::
 
     1. Add documentation for the data attributes in read-mode.
-    2. Document copy() and define copy mode.
+    2. Define copy mode.
     3. Provide example of updating inference_sites
 
 .. autoclass:: tsinfer.AncestorData


hyanwong:
Cool, thanks!

hyanwong:
(Needed this for tsinfer, not tksit, but I can make the change there)

benjeffery:
I thought this was tsinfer?

hyanwong:
Oops. dunno how I mistook this for tskit. Sorry. Brain fart.

Actually, it's because the header says "merge 1 commit into tskit-dev:master from ...", and doesn't list the actual project within tskit-dev.

hyanwong:
@jeromekelleher I'm trying to document `SampleData.copy()` and friends (marked as a TODO, and relevant to #209 . This PR does that, but the docs for SampleData then end up referring to DataContainer (which isn't documented), e.g. under the SampleData API docs we have:

image

Is that OK, or if not, how and where do you want to write the docstrings? it would be nice if we could programmatically refer to the child class name itself in the docstring, but I can't see any option to do that in Sphinx.

hyanwong:
NB:  I guess we could work around it by changing the language, e.g. using the word "class" rather than "DataContainer" where necessary.

jeromekelleher:
Looks good to me, thanks @benjeffery.

PR #252, opened 2020-03-25
Switch to times as frequencies rather than counts

### LINKED ISSUE #210: Path compressed ancestor time increment is hardcoded: can cause errors for user-input variant times
I'm trying to re-infer a sample data file produced from the trees file on the wellcome server at

`/well/mcvean/ukbb12788/wohns/tsinfer_geva/tgp_geva_chr20.trees`

After installing the latest (github) version of tsinfer:

`python3 -m pip install git+https://github.com/tskit-dev/tsinfer --user`

I then use the from_tree_sequence method to create a sample data file within python:

```
import tskit
ts = tskit.load("/well/mcvean/ukbb12788/wohns/tsinfer_geva/tgp_geva_chr20.trees")
sd = tsinfer.SampleData.from_tree_sequence(
    ts, path="tgp_geva_chr20.samples", use_times=True)
sd.close()
```

then when I try inference, I fail at the match_samples stage, with error -5:

```
yan@holly:~$ python3 -m tsinfer infer tgp_geva_chr20.samples -p -t 6
ga-add   (1/6)100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 860k/860k [00:17, 50.5kit/s]
ga-gen   (2/6)100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 739k/739k [02:53, 4.25kit/s]
ma-match (3/6)  7%|█████████▊                                                                                                                                           | 48.7k/739k [14:45, 33.0it/s]2019-11-07 17:21:19,423 [6387] CRITICAL root: Traceback (most recent call last):
  File "/usr/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/yan/.local/lib/python3.7/site-packages/tsinfer/__main__.py", line 9, in <module>
    main()
  File "/home/yan/.local/lib/python3.7/site-packages/tsinfer/__main__.py", line 5, in main
    cli.tsinfer_main()
  File "/home/yan/.local/lib/python3.7/site-packages/tsinfer/cli.py", line 479, in tsinfer_main
    args.runner(args)
  File "/home/yan/.local/lib/python3.7/site-packages/tsinfer/cli.py", line 187, in run_infer
    sample_data, progress_monitor=progress_monitor, num_threads=args.num_threads)
  File "/home/yan/.local/lib/python3.7/site-packages/tsinfer/inference.py", line 171, in infer
    path_compression=path_compression, progress_monitor=progress_monitor)
  File "/home/yan/.local/lib/python3.7/site-packages/tsinfer/inference.py", line 249, in match_ancestors
    return matcher.match_ancestors()
  File "/home/yan/.local/lib/python3.7/site-packages/tsinfer/inference.py", line 995, in match_ancestors
    self.__match_ancestors_multi_threaded()
  File "/home/yan/.local/lib/python3.7/site-packages/tsinfer/inference.py", line 981, in __match_ancestors_multi_threaded
    self.__complete_epoch(j)
  File "/home/yan/.local/lib/python3.7/site-packages/tsinfer/inference.py", line 922, in __complete_epoch
    extended_checks=self.extended_checks)
_tsinfer.LibraryError: Error occured: -5
```

Where can I look up the meaning of a `-5` error, and how can I slim this down to debug?

### LINKED ISSUE #227: Should probably switch to true frequencies for variant ages, rather than using counts
In #169 we treat variants which have a missing datapoint as contributing 0.5 to the count of derived alleles, which we then use as a proxy for age. However, it would be better to use the sample frequency instead, rather than the count, so that the "time" of variants ranges from 0-1. However, this is likely to cause some issues down the line, so is not implemented yet.

### LINKED ISSUE #277: Read large VCFs in parallel
One of the slow stages in inference is actually reading in the VCF file. For example, @awohns finds that importing the 1000G vcf as a SampleData file takes 6 hours! One easy way to speed this up is to chunk up the VCF file by region, read in all N chunks in parallel, and munge them all together at the end.

The easiest way I can see to do this is to actually create N SampleData files for each region, with identical data apart from the sites. This then requires a function to append them all together. I've got something working, that looks like this:

```
def append_sites(self, *additional_samples):
    """
    For a writeable sample data file, take a set of sample data objects 
    with exactly the same populations, individuals, and samples, but with
    additional sites, and append these sites in order after the existing sites.
    The additional sample data objects must contain sites in order.
    """
    self._check_write_modes()
    last_pos = self.sites_position[-1]
    for other in additional_samples:
        if other.sites_position[0] <= last_pos:
            raise ValueError("sample data files must be in ascending order of genome position")
        last_pos = other.sites_position[-1]
        if not self.sequence_length == other.sequence_length:
            raise ValueError(f"sample data files must have identical sequence lengths")
        if not self.formats_equal(other):
            raise ValueError(f"sample data files must be of the same format")            
        if not self.populations_equal(other):
            raise ValueError(f"sample data files must have identical populations")
        if not self.individuals_equal(other):
            raise ValueError(f"sample data files must have identical individuals")
        if not self.samples_equal(other):
            raise ValueError(f"sample data files must have identical samples")
    for other in additional_samples:
        for name, arr in self.arrays():
            if name.startswith("sites/"):
                arr.append(other.data[name])
```

(This requires some additional checking routines that I think are a useful addition to tsinfer anyway).

I've got this working, and Wilder has an alternative version too, and it reduces the read time to <20 mins from 6 hours, so I think is worth it. If it is deemed a sensible approach, I can also add it to the docs.

This is part of a general theme of wanting to slice, dice, and past SampleData files, so it may be that we want a more general routine that can (for example) add extra haplotypes to a SampleData file too. But perhaps we want to do this easy step first?

hyanwong:
Fixes #227.

We had expected this to be more involved than simply changing the values from counts to freqs, but this simple change seems to pass all the tests. However, we might want to adjust the delta value for PC_ANCESTOR_INCREMENT (see e.g. https://github.com/tskit-dev/tsinfer/issues/210#issuecomment-551303076) as the current increment of 1.0/65536 will be too large when using a frequency timescale. To replicate the current behaviour we should set it to  `1.0/65536/num_samples`, but that won't fix #210. What would you like me to do @jeromekelleher ?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,7 +6,8 @@
 
 - The ancestors tree sequence now contains the real alleles and not
   0/1 values as before.
-
+- The time value for a node is now measured on a 0..1 (frequency) scale, not as a count
+  from 1..num_samples
 
 ********************
 [0.1.5] - 2019-09-25
--- a/tests/test_formats.py
+++ b/tests/test_formats.py
@@ -1239,7 +1239,8 @@ class TestSampleData(unittest.TestCase, DataContainerMixin):
         with formats.SampleData() as data:
             for j in range(4):
                 data.add_site(position=j, alleles=["0", "1"], genotypes=[0, 1, 1, 0])
-        self.assertEqual(list(data.sites_time), [2.0, 2.0, 2.0, 2.0])  # Freq == 2.0
+        # Freq == 0.5
+        self.assertAlmostEqual(list(data.sites_time), [0.5, 0.5, 0.5, 0.5])
 
         with tempfile.TemporaryDirectory(prefix="tsinf_format_test") as tempdir:
             filename = os.path.join(tempdir, "samples.tmp")
@@ -1254,7 +1255,7 @@ class TestSampleData(unittest.TestCase, DataContainerMixin):
                     copy.sites_time = time
                 self.assertFalse(copy.data_equal(data))
                 self.assertEqual(list(copy.sites_time), time)
-                self.assertEqual(list(data.sites_time), [2.0, 2.0, 2.0, 2.0])
+                self.assertAlmostEqual(list(data.sites_time), [0.5, 0.5, 0.5, 0.5])
                 if copy_path is not None:
                     copy.close()
 
@@ -1290,7 +1291,7 @@ class TestSampleData(unittest.TestCase, DataContainerMixin):
         for j in range(4):
             data.add_site(position=j, alleles=["0", "1"], genotypes=[0, 1, 1, 0])
         data.finalise()
-        self.assertEqual(list(data.sites_time), [2.0, 2.0, 2.0, 2.0])
+        self.assertAlmostEqual(list(data.sites_time), [0.5, 0.5, 0.5, 0.5])
         copy = data.copy()
         for bad_shape in [[], np.arange(100, dtype=np.float64), np.zeros((2, 2))]:
             self.assertRaises((ValueError, TypeError), set_value, copy, bad_shape)
--- a/tsinfer/formats.py
+++ b/tsinfer/formats.py
@@ -1381,10 +1381,10 @@ class SampleData(DataContainer):
             number of samples carrying the derived state is greater than
             1 and less than the number of samples.
         :param float time: The time of occurence (pastwards) of the mutation to the
-            derived state at this site. If not specified or None, the count of the
-            derived alleles (i.e., the number of non-zero values in the genotypes) is
-            used instead, with missing values contributing 0.5 to the count. For
-            biallelic sites this count should provide a reasonable estimate
+            derived state at this site. If not specified or None, the frequency of the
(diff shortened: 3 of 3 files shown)


hyanwong:
Also not sure why the lint / pre-commit step is failing on running `Black` - is the line length set wrongly?

hyanwong:
I think this can be merged, @jeromekelleher - I can't see why it is failing coverage, as there are no extra bits of code added, only removed.

jeromekelleher:
LGTM. Can you update the CHANGELOG with this please? It's easier to document these changes as they happen.

hyanwong:
Added to the changelog. Should now be OK to merge.

jeromekelleher:
Merged, thanks @hyanwong!

PR #254, opened 2020-03-25
Mark sites whose time is unspecified as -np.inf, implying "use freq"

hyanwong:
I think this is a reasonably neat and non-invasive way to store the information that a site has been marked as "use freq as a proxy for time". The `Variant` object gains an additional attribute, `inference_time`, which is the site.time if the user defined it, otherwise it is the frequency. This seems a reasonable attribute of a *variant* rather than a *site* to me, as it changes depending on the samples that are passed in.

~~Although the tests pass as-is, one thing which is a bit of a hack is the `from_tree_sequence` method. Here, for a non-variable site (no mutations), we set the `site.time` to -np.inf (i.e. `USE_FREQ_AS_TIME`). I would prefer it, I think, if we set them to `np.inf` (or `np.nan`, although that seems to raise other issues), as a marker that the concept of a variant time is nonsensical for a non-variable site. However, that leads to `test_formats.py:TestSampleData.test_from_tree_sequence_variable_allele_number` failing (can't round-trip), so I haven't implemented it until I get some feedback.~~ (corrected in latest push)

### DIFF
--- a/tests/test_formats.py
+++ b/tests/test_formats.py
@@ -1239,10 +1239,10 @@ class TestSampleData(unittest.TestCase, DataContainerMixin):
     def test_copy_update_sites_time(self):
         with formats.SampleData() as data:
             for j in range(4):
-                data.add_site(position=j, alleles=["0", "1"], genotypes=[0, 1, 1, 0])
-        for v in data.variants():
-            self.assertAlmostEqual(v.site.time, data.USE_FREQ_AS_TIME)
-            self.assertAlmostEqual(v.inference_time, 0.5)  # Freq == 0.5
+                data.add_site(
+                    position=j, alleles=["0", "1"], genotypes=[0, 1, 1, 0], time=0.5
+                )
+        self.assertAlmostEqual(list(data.sites_time), [0.5, 0.5, 0.5, 0.5])
 
         with tempfile.TemporaryDirectory(prefix="tsinf_format_test") as tempdir:
             filename = os.path.join(tempdir, "samples.tmp")
@@ -1257,9 +1257,7 @@ class TestSampleData(unittest.TestCase, DataContainerMixin):
                     copy.sites_time = time
                 self.assertFalse(copy.data_equal(data))
                 self.assertEqual(list(copy.sites_time), time)
-                for v in data.variants():
-                    self.assertAlmostEqual(v.site.time, data.USE_FREQ_AS_TIME)
-                    self.assertAlmostEqual(v.inference_time, 0.5)  # Freq == 0.5
+                self.assertAlmostEqual(list(data.sites_time), [0.5, 0.5, 0.5, 0.5])
                 if copy_path is not None:
                     copy.close()
 
@@ -1293,10 +1291,11 @@ class TestSampleData(unittest.TestCase, DataContainerMixin):
 
         data = formats.SampleData()
         for j in range(4):
-            data.add_site(position=j, alleles=["0", "1"], genotypes=[0, 1, 1, 0])
+            data.add_site(
+                position=j, alleles=["0", "1"], genotypes=[0, 1, 1, 0], time=0.5
+            )
         data.finalise()
-        for v in data.variants():
-            self.assertAlmostEqual(v.inference_time, 0.5)
+        self.assertAlmostEqual(list(data.sites_time), [0.5, 0.5, 0.5, 0.5])
         copy = data.copy()
         for bad_shape in [[], np.arange(100, dtype=np.float64), np.zeros((2, 2))]:
             self.assertRaises((ValueError, TypeError), set_value, copy, bad_shape)
--- a/tests/test_formats.py
+++ b/tests/test_formats.py
@@ -40,7 +40,6 @@ import tskit
 import tsinfer
 import tsinfer.formats as formats
 import tsinfer.exceptions as exceptions
-import tsinfer.constants as constants
 
 
 class DataContainerMixin(object):
@@ -171,11 +170,9 @@ class TestSampleData(unittest.TestCase, DataContainerMixin):
                     samples_metadata=[json.loads(n.metadata or "{}") for n in nodes],
                 )
         for v in ts.variants():
-            t = None
+            t = np.nan  # default is that a site has no meaningful time
             if len(v.site.mutations) == 1:
                 t = ts.node(v.site.mutations[0].node).time
-            else:
-                t = constants.TIME_MEANINGLESS
             input_file.add_site(
                 v.site.position,
                 v.genotypes,
--- a/tests/test_formats.py
+++ b/tests/test_formats.py
@@ -173,6 +173,8 @@ class TestSampleData(unittest.TestCase, DataContainerMixin):
             t = None
             if len(v.site.mutations) == 1:
                 t = ts.node(v.site.mutations[0].node).time
+            else:
+                t = input_file.MEANINGLESS_TIME
             input_file.add_site(
                 v.site.position,
                 v.genotypes,
@@ -369,7 +371,7 @@ class TestSampleData(unittest.TestCase, DataContainerMixin):
 
     def test_from_tree_sequence_variable_allele_number(self):
         ts = self.get_example_ts(10, 10)
-        # Create > 2 alles by scattering mutations on the tree nodes at the first site
+        # Create > 2 alleles by scattering mutations on the tree nodes at the first site
         tables = ts.dump_tables()
         focal_site = ts.site(0)
         tree = ts.at(focal_site.position)
@@ -393,7 +395,6 @@ class TestSampleData(unittest.TestCase, DataContainerMixin):
         self.verify_data_round_trip(ts, sd1)
         self.assertFalse(sd1.sites_inference[0])
         self.assertFalse(sd1.sites_inference[sd1.num_sites - 1])
-
         num_alleles = sd1.num_alleles()
         for var in ts.variants():
             self.assertEqual(len(var.alleles), num_alleles[var.site.id])
@@ -1239,8 +1240,9 @@ class TestSampleData(unittest.TestCase, DataContainerMixin):
         with formats.SampleData() as data:
             for j in range(4):
                 data.add_site(position=j, alleles=["0", "1"], genotypes=[0, 1, 1, 0])
-        # Freq == 0.5
-        self.assertAlmostEqual(list(data.sites_time), [0.5, 0.5, 0.5, 0.5])
+        for v in data.variants():
+            self.assertAlmostEqual(v.site.time, data.USE_FREQ_AS_TIME)
+            self.assertAlmostEqual(v.inference_time, 0.5)  # Freq == 0.5
 
         with tempfile.TemporaryDirectory(prefix="tsinf_format_test") as tempdir:
             filename = os.path.join(tempdir, "samples.tmp")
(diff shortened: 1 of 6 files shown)


hyanwong:
This fixes #255, which describes the justification for the change

tests/test_formats.py
```diff
@@ -403,6 +403,7 @@ def test_from_tree_sequence_variable_allele_number(self):
             ):
                 self.assertEqual(len(var.alleles), num_alleles)
         sd2 = formats.SampleData.from_tree_sequence(ts)
+        print(sd1.sites_time[:], sd2.sites_time[:])
```
jeromekelleher: stray print

tsinfer/formats.py
```diff
@@ -69,6 +69,16 @@ def remove_lmdb_lockfile(lmdb_file):
         os.unlink(lockfile)
 
 
+def allele_counts(genotypes):
+    """
+    Return summary counts of the number of different allele types for a genotypes array
+    """
+    n_known = np.sum(genotypes != tskit.MISSING_DATA)
```
jeromekelleher: Returning this as a namedtuple would be nice. Client code could then write ``counts.known``, etc.

tsinfer/formats.py
```diff
@@ -634,11 +644,15 @@ class Site(object):
...
     A single variant. Mirrors the definition in tskit but with some extra fields.
+    The inference_time field returns the time used for inference: where a site
+    is using frequency as a proxy for time, this will contain the frequency, otherwise
+    it will simply contain the site.time value.
     """
 
     # TODO document properly.
     site = attr.ib()
     genotypes = attr.ib()
+    inference_time = attr.ib()
```
jeromekelleher: This is a property of the Site isn't it?

tsinfer/formats.py
```diff
@@ -766,6 +780,10 @@ class SampleData(DataContainer):
     ADDING_SAMPLES = 1
     ADDING_SITES = 2
 
+    # Marker value for site times that will use frequency as a proxy for time
+    # We use -inf rather than NaN for simplicity
+    USE_FREQ_AS_TIME = -np.inf
```
jeromekelleher: Good call.

tsinfer/formats.py
```diff
@@ -1185,9 +1203,13 @@ def from_tree_sequence(cls, ts, use_times=True, use_metadata=True, **kwargs):
                     samples_metadata=[sample_metadata],
                 )
         for v in ts.variants():
-            variant_time = None
```
jeromekelleher: Can't this be done by just changing ``variant.time = USE_FREQ_AS_TIME`` on the first line? As in, there's no need to change the logic, we just change the marker value.

tsinfer/formats.py
```diff
@@ -1495,7 +1513,7 @@ def finalise(self):
 
     def sites(self, inference_sites=None):
         """
-        Returns an iterator over the Site objects.
```
jeromekelleher: Full stop seems fine?

tsinfer/formats.py
```diff
@@ -1563,7 +1584,17 @@ def variants(self, inference_sites=None):
         for (j, genotypes), site in zip(
             self.genotypes(inference_sites), self.sites(inference_sites)
         ):
-            variant = Variant(site=site, alleles=site.alleles, genotypes=genotypes)
+            if site.time == self.USE_FREQ_AS_TIME:
```
jeromekelleher: Maybe it's just me, but I find it clearer to assign the value for the nominal case like this:
```python
time = site.time
if site.time == self.USE_FREQ_AS_TIME:
    counts = allele_counts(genotypes)
    time = counts.derived / counts.known
```

It just seems a little clearer what's the nominal and exceptional cases here, and saves a little vertical space.

tsinfer/inference.py
```diff
@@ -469,7 +469,7 @@ def add_sites(self):
         logger.info("Starting addition of {} sites".format(self.num_sites))
         progress = self.progress_monitor.get("ga_add_sites", self.num_sites)
         for j, variant in enumerate(self.sample_data.variants(inference_sites=True)):
-            self.ancestor_builder.add_site(j, variant.site.time, variant.genotypes)
+            self.ancestor_builder.add_site(j, variant.inference_time, variant.genotypes)
```
jeromekelleher: Not sure why we need the variant.inference_time field here?

tests/test_formats.py
hyanwong: Thanks

tsinfer/formats.py
hyanwong: Good idea

tsinfer/formats.py
hyanwong: This has to be a property of the Variant, because it it is a function of the genotypes (which the Site doesn't have access to). The Site only has access to the passed-in time, which (when time==freq) is the marker value `-inf`. We can't add it to the Site without slowing down the Sites iterator by grabbing the genotypes on iteration.

tsinfer/formats.py
hyanwong: Sorry - stray delete. Will change

tsinfer/inference.py
hyanwong: Variant.site.time will be `-inf` if no time was specified by the user. It is only the variant.inference_time that contains the frequency value of that genotype during iteration.

tsinfer/formats.py
hyanwong: I'm not sure what you mean here. I don't think a `ts.Variant` object has a `time` attribute?

hyanwong:
I've also added `MEANINGLESS_TIME = np.inf` to the marker values that can be put into a time field. This helped catch a bug in the test suite, so I think it is worth it (buxfig is now the first commit in this PR 8731a1d09a92617f3ef881baaec20b9354a9201b)

tsinfer/formats.py
hyanwong: Ping @jeromekelleher in case this conversation is muted

tsinfer/formats.py
hyanwong: Done

tsinfer/formats.py
jeromekelleher: I meant ``variant_time``

jeromekelleher:
I'm not sure this is the right approach Yan - maybe we should decouple the default time computation from the storage format entirely, and move it into ``inference.py`` (i.e., put the logic in AncestorGenerator.add_sites)? This seems more flexible and removes the need for most of the code here.

tsinfer/formats.py
hyanwong: @jeromekelleher - Oh, I see. You mean the same as the others, take the stuff out of the `else` clause(s) as before. Something like:

```
        for v in ts.variants():
            variant_time = USE_FREQ_AS_TIME
            if use_times:
                variant_time = self.MEANINGLESS_TIME
                if len(v.site.mutations) == 1:
                    variant_time = ts.node(v.site.mutations[0].node).time
        ...
```

hyanwong:
I can do that, it just seemed less invasive to do it like this (I wasn't sure if the inference time was used elsewhere). But thinking about it, you are probably right. The logic is inference logic, not storage logic. I'll see how it looks when moved into `inference.py`, now that I know it's only used in one place.

tsinfer/formats.py
jeromekelleher: Yes - see point below about taking this out of SampleData entirely though.

jeromekelleher:
I think that's the only place it's used, could be wrong though!

hyanwong:
Yep, easy to move into `inference.py` and only used in one place, as you thought. I've done this as a separate commit so that you can see the difference. Let me know when you want me to squash the commits together @jeromekelleher

tsinfer/formats.py
```diff
@@ -766,6 +781,10 @@ class SampleData(DataContainer):
     ADDING_SAMPLES = 1
     ADDING_SITES = 2
 
+    # Marker constants for time value
+    USE_FREQ_AS_TIME = -np.inf  # Placeholder to mark freq of derived alleles as time
```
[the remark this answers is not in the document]
hyanwong: Done in latest push

tsinfer/formats.py
```diff
@@ -766,6 +781,10 @@ class SampleData(DataContainer):
     ADDING_SAMPLES = 1
     ADDING_SITES = 2
 
+    # Marker constants for time value
+    TIME_UNSPECIFIED = -np.inf  # Placeholder to mark freq of derived alleles as time
+    TIME_MEANINGLESS = np.inf  # Placeholder for e.g. freq value for a nonvariable site
```
jeromekelleher: We're still making judgements here about what the time **means**. I think we should just have TIME_UNSPECIFIED and nothing else. Leave the semantics to the consumer.

tsinfer/inference.py
```diff
@@ -469,7 +469,16 @@ def add_sites(self):
         logger.info("Starting addition of {} sites".format(self.num_sites))
         progress = self.progress_monitor.get("ga_add_sites", self.num_sites)
         for j, variant in enumerate(self.sample_data.variants(inference_sites=True)):
-            self.ancestor_builder.add_site(j, variant.site.time, variant.genotypes)
+            time = variant.site.time
+            if time == self.sample_data.TIME_UNSPECIFIED:
```
jeromekelleher: Would be better to make this a global constance, like ``tsinfer.TIME_UNSPECIFIED`` rather than a class constant.

tsinfer/formats.py
```diff
@@ -1185,9 +1204,11 @@ def from_tree_sequence(cls, ts, use_times=True, use_metadata=True, **kwargs):
                     samples_metadata=[sample_metadata],
                 )
         for v in ts.variants():
-            variant_time = None
-            if use_times and len(v.site.mutations) == 1:
-                variant_time = ts.node(v.site.mutations[0].node).time
+            variant_time = self.TIME_UNSPECIFIED
+            if use_times:
+                variant_time = self.TIME_MEANINGLESS
```
jeromekelleher: To be clear, I don't think we should have TIME_MEANINGLESS at all.

tsinfer/formats.py
hyanwong: I see "unspecified by user" as quite different to "algorithmically meaningless". The first case is used to store the user preference. The second is to flag up times when we wrongly try to use a meaningless time. It's demonstrated by the fact that by adding `TIME_MEANINGLESS` in addition to `TIME_UNSPECIFIED`, I managed to pick up a long standing bug in the test suite.

tsinfer/formats.py
hyanwong: See my [comment above](https://github.com/tskit-dev/tsinfer/pull/254#discussion_r399219938). I think it is distinct and useful...

tsinfer/formats.py
hyanwong: Also, if we programatically set `TIME_UNSPECIFIED` when we really mean `TIME_MEANINGLESS`, we override any information about what the user originally fed in to the algorithm. I'm prepared to say that we should kill off `TIME_MEANINGLESS` and simply put `np.nan` in there though, which is what we would get anyway when all samples at a site are marked as missing data, and the freq is 0/0.

hyanwong:
@jeromekelleher the latest commit in this PR, e324e3f382c71c12d539c6e1b72ebd700d561de6, uses np.nan instead of constants.TIME_MEANINGLESS, but that means we then need to adjust `SampleData.data_equal` to use something like `np.allclose(..., equal_nan=True)` for comparisons.

jeromekelleher:
This looks good to me thanks @hyanwong. We can work out the kinks on this new approach as we hit them but I think it's basically the right approach.

PR #256, opened 2020-03-27
Sampledata subset/delete

### LINKED ISSUE #209: Create new samples file with a subset of sites

hyanwong:
To be considered after #254 is merged. Submitted here for initial review, fixes #209 and allows us to test inference against subsets of large datasets, such as 1000G (see comments at https://github.com/tskit-dev/tsinfer/issues/209#issuecomment-601259473). Rather than `SampleData.subset(sites=[], samples=[])`, I have reversed the logic and require the user to pass in sites or samples to *omit* from the new, returned SampleData file, rather than samples and sites to *keep*. This makes more sense with the default values of `None`, meaning do not remove any samples. I'm not sure about the name though. Do you have any better ideas than `delete`, @jeromekelleher ?

### DIFF
--- a/tests/test_formats.py
+++ b/tests/test_formats.py
@@ -147,7 +147,7 @@ class TestSampleData(unittest.TestCase, DataContainerMixin):
             random_seed=100,
         )
 
-    def verify_data_round_trip(self, ts, input_file):
+    def verify_data_round_trip(self, ts, input_file, use_times=True):
         self.assertGreater(ts.num_sites, 1)
         for pop in ts.populations():
             input_file.add_population(metadata=json.loads(pop.metadata or "{}"))
@@ -161,7 +161,7 @@ class TestSampleData(unittest.TestCase, DataContainerMixin):
                     input_file.add_individual(
                         ploidy=1,
                         population=node.population,
-                        time=node.time,
+                        time=node.time if use_times else 0,
                         samples_metadata=[json.loads(node.metadata or "{}")],
                     )
             else:
@@ -170,13 +170,15 @@ class TestSampleData(unittest.TestCase, DataContainerMixin):
                     population=nodes[0].population,
                     metadata=json.loads(ts.individual(i).metadata or "{}"),
                     location=ts.individual(i).location,
-                    time=nodes[0].time,
+                    time=nodes[0].time if use_times else 0,
                     samples_metadata=[json.loads(n.metadata or "{}") for n in nodes],
                 )
         for v in ts.variants():
-            t = np.nan  # default is that a site has no meaningful time
-            if len(v.site.mutations) == 1:
-                t = ts.node(v.site.mutations[0].node).time
+            t = None
+            if use_times:
+                t = np.nan  # default when setting times is no meaningful time
+                if len(v.site.mutations) == 1:
+                    t = ts.node(v.site.mutations[0].node).time
             input_file.add_site(
                 v.site.position,
                 v.genotypes,
@@ -207,7 +209,7 @@ class TestSampleData(unittest.TestCase, DataContainerMixin):
             self.assertEqual(variant.site.position, position[j])
             self.assertTrue(np.all(variant.genotypes == genotypes[j]))
             self.assertEqual(alleles[j], list(variant.alleles))
-            if len(variant.site.mutations) == 1:
+            if use_times and len(variant.site.mutations) == 1:
                 the_time = ts.node(variant.site.mutations[0].node).time
                 self.assertEqual(the_time, site_times[j])
             if variant.site.metadata:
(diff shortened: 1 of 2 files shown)


hyanwong:
I've also started documenting parts of the SampleData API, as I had to refer to them when documenting the new function.

hyanwong:
I think, pending a suggestion for name changes, this is ready for review and (hopefully) merge, @jeromekelleher ? Then I can code up the new inference tests on small fractions of larger datasets.

Perhaps @benjeffery has an idea for a nicer name that essentially means "make a copy of this data, but delete the following sites and/or samples from it"?

benjeffery:
How about `copy_and_remove`? I liked `copy_except` but it seemed too cryptic.

hyanwong:
@benjeffery I'm wondering if it's cleared to have something like: `subset(delete_samples, delete_sites)`. Or if we use `copy_and_remove`, I wonder if the current `copy` should be `copy_and_edit`?

benjeffery:
I'd be happier with `subset` if the argument was the things you were keeping.

jeromekelleher:
I think using ``subset()`` and stating the things you want to keep is the right way to go.

jeromekelleher:
Can you pull the documentation stuff out of this PR please @hyanwong? It helps to have small, focused PRs when multiple people are working on the code. Should be easy enough to cherry pick off the last commit.

hyanwong:
> Can you pull the documentation stuff out of this PR please @hyanwong? It helps to have small, focused PRs when multiple people are working on the code. Should be easy enough to cherry pick off the last commit.

Yep, no problem

hyanwong:
Docs in another branch now. Just to check that we are happy with keeping `SampleData.copy()` with that name, as it is now being documented. The reason I ask is that it doesn't *quite* align with the normal semantics of "object.copy", as the created SampleData object is not exactly identical to the original, but instead opened in write-mode, needing to be `finalise`d before use. So I wondered if a clearer name might be e.g. `copy_and_edit()` or something similar?

jeromekelleher:
Or maybe just `edit`, since it returns a file in edit mode? Let's not do that here though, one thing at a time.

hyanwong:
> Or maybe just `edit`, since it returns a file in edit mode? Let's not do that here though, one thing at a time.

Sure. It's just if we want to call the function in this PR `copy_and_remove`, then it would argue for calling the other one `copy_and_edit`, I guess?

hyanwong:
@jeromekelleher - as discussed, I tested without re-setting `sites_inference[:]`, and it seemed to work, but I before updating this PR I thought I should check that (a) there should be no problems with sites marked for inference at time 0 (which would be at the same time as samples, so perhaps weird) or sites marks for inference at time `np.nan` (if the freq is 0/0 because all sites have become missing). I'm wondering if it would be better to implement the "keep inference state as null" idea first, then all this goes away.

jeromekelleher:
Yes, that does sounds like a good idea. I'll take a look at the ``sites_inference`` thing today.

awohns:
A `subset` function seems like a really handy one to have, I'd certainly find it useful in a week or two with the big sampledata file I'm harmonizing now!

awohns:
Just checking in on the status of this PR. Does it seem ok to merge? I've been using the delete function in my iterative script and it's been working fine.

hyanwong:
I think the only thing remaining to discuss here is (a) the name of the function and (b) whether we should specify the sites/samples to *remove* or to *keep*.

jeromekelleher:
We're still making semantic choices here, which #266 is removing. I've given a quick look through again, and I also don't understand a few things, so unless having this in master is blocking forward progress I don't want to get caught up in it right now.

awohns:
No problem from my perspective leaving unmerged, working with @hyanwong 's branch installed has been fine for me

hyanwong:
Replaced with #293

PR #258, opened 2020-03-27
Support for missing data in LS copying model.

jeromekelleher:
WIP


jeromekelleher:
@hyanwong, it would be good to get your eyes on this at this point. I've added a new ``impute_missing`` option to ``infer`` and friends. If this is True (default) we impute any missing data in the sample file in the output tree sequence. If False, we guarantee round-tripping of the missing data in the output. We do this by snipping out edges that cover any of the affected sites.

The default case is working now, but the impute_missing=False case will take a bit more work. It might be quite a bit more work, I'm not sure, as it now means that we violate the assumption that edges for a given child are always contiguous. I don't know yet how deeply that assumption is built in. We'll also have to do something similar for the sites that we ``map_mutations`` on, I haven't thought that through properly yet. It might turn out that it's simpler to do all the edge surgery afterwards as a post-processing step. (Actually, it almost certainly is, and I can reuse the machinery built here...)

Anyway, good to get your thoughts.

hyanwong:
Will get onto this tomorrow (Tues) morning. Had internet problems today, apologies.

jeromekelleher:
@hyanwong, I've decided to back out of supporting round-tripping missing data. It seems like a fairly niche requirement right now and is really just a bit of fiddly tree sequence surgery imposed on the output. It's reasonably expensive to do, as we have to resort the output edges, and iterate over all the sample genotypes, so would take a bit of time to do properly. 

It seems to me the most important feature is to check how we're doing with a small amount of missing data, so I'll write some evaluation checks to see how we're doing and update the PR.

hyanwong:
> @hyanwong, I've decided to back out of supporting round-tripping missing data. It seems like a fairly niche requirement right now and is really just a bit of fiddly tree sequence surgery imposed on the output.

I agree with this. We decided to implement it as a post-inference surgery step anyway. The only issue is that we need to skip any tests that try round-tripping with missing data. We should probably note it somewhere in the docs too.

> It seems to me the most important feature is to check how we're doing with a small amount of missing data, so I'll write some evaluation checks to see how we're doing and update the PR.

Great. I agree.

jeromekelleher:
I think this is ready to go - I've added a new evaluation tool "inference-accuracy" which sees how well we impute randomly missing data. At the moment we just naively mark a fraction of the genotypes as missing, which might well be unfair and we might need to use something more like the error generation approach.

Nonetheless, the accuracy is reasonably good so it's clear we're not doing anything majorly wrong. I'll post an example of the output here in a minute after the current run is done.

hyanwong:
Do you want me to review?

jeromekelleher:
Here's the output for n=100 and L=10Mb:

imputation-accuracy_n=100_L=10 0_mu=1e-08_rho=1e-08_num

jeromekelleher:
That seems reasonable, right? Linear drop off in accuracy, and ~99% accuracy on imputation for 1% missing data rate.

jeromekelleher:
> Do you want me to review?

yes please @hyanwong!

hyanwong:
Graph looks reasonably to me, although I'll need to think about it a bit. Will look at the code tonight.

evaluation.py
```diff
@@ -659,6 +659,103 @@ def run_ancestor_properties(args):
...
+    G = ts.genotype_matrix()
+    missing = np.random.rand(ts.num_sites, ts.num_samples) < missing_proportion
+    G[missing] = tskit.MISSING_DATA
+    with tsinfer.SampleData(ts.sequence_length) as sample_data:
+        for var in ts.variants():
+            sample_data.add_site(
+                var.site.position, alleles=var.alleles, genotypes=G[var.site.id]
+            )
+
+    # FIXME need to turn off path compression when running this with n=1000
```
hyanwong, line 675: This is weird? Is this because of the recent changes that convert time to freq rather than to counts? I think the PC_ANCESTOR_INCREMENT value is probably too large now, which might be at the root of this problem (see also https://github.com/tskit-dev/tsinfer/issues/210#issuecomment-553875914

jeromekelleher: Yes, that's the issue all right. I'll follow up on it.

jeromekelleher:
OK, I'm going to merge this and follow up then on the ancestor increment thing.

hyanwong:
Yes, do, if you are happy with it. Sorry I didn't have time to review properly. Juggling kids.

jeromekelleher:
> Juggling kids.

Sounds dangerous! Don't be tempted to mix in any chainsaws too...

hyanwong:
> > Juggling kids.
> 
> Sounds dangerous! Don't be tempted to mix in any chainsaws too...

IMG_4394

PR #264, opened 2020-04-02
Move away from determining inference sites at import time.

jeromekelleher:
WIP

First pass is to refactor the main inference code to stop using the ``sample_data.sites_inference`` array. This is fine because both the AncestorData and ancestors tree sequence store the site positions that are actually used.

### DIFF
--- a/dev.py
+++ b/dev.py
@@ -6,8 +6,9 @@ import tqdm
 import pprint
 import numpy as np
 import json
-import tskit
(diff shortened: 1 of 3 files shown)


jeromekelleher:
Going to merge this much first before I change any of the inference semantics, as this is just a refactor up to this point, and we'll be able to see better what the actual changes are if we merge it.

PR #266, opened 2020-04-07
Determine 'inference sites' at ancestor build time.

jeromekelleher:
Deciding what sites to use as inference at import time is inflexible and creates unwanted linkage between the SampleData file and the AncestorBuilder. The SampleData file should not  be making any semantic judgements on what the AncestorBuilder things is good input for it to work with.

This changes the semantics so that the decision about whether to use a site for inference or not is decided when we add the site to the AncestorBuilder. The user can still state that the definitely want the site to be used for inference or not, but by default the AncestorBuilder will decide.

The basic update worked easily enough and the core inference is working well. Predictably though, this has broken a bunch of stuff in the tests and the ancillary tools. We use the idea of ``inference_sites`` quite a lot.

Since the term ``inference_sites`` isn't particularly well chosen anyway (we still do inference on the other sites, we just don't use them in building ancestral haplotypes), I'm tempted to change the terminology entirely and raise and error if someone uses the ``inference_sites`` approach. Code that depends on ``SampleData.sites_inference`` is most likely broken now anyway, so it may be the safest thing to do.

Any thoughts here @hyanwong?

### DIFF
--- a/tests/test_inference.py
+++ b/tests/test_inference.py
@@ -409,15 +409,10 @@ class TestSparseAncestorsRoundTrip(TestRoundTrip):
                 site_alleles = None if alleles is None else alleles[j]
                 sample_data.add_site(positions[j], genotypes[j], site_alleles, time=t)
 
-        # Everything is an inference site
-        with sample_data.copy() as sample_data:
-            sample_data.sites_inference[:] = True
-
-        num_alleles = sample_data.num_alleles(inference_sites=True)
-        num_sites = sample_data.num_inference_sites
+        num_alleles = sample_data.num_alleles()
         with tsinfer.AncestorData(sample_data) as ancestor_data:
             t = np.sum(num_alleles) + 1
-            for j in range(num_sites):
+            for j in range(sample_data.num_sites):
                 for allele in range(num_alleles[j] - 1):
                     ancestor_data.add_ancestor(j, j + 1, t, [j], [allele])
                     t -= 1
@@ -476,15 +471,23 @@ class TestNonInferenceSitesRoundTrip(unittest.TestCase):
     of inference and non inference sites.
     """
 
-    def verify_round_trip(self, genotypes, inference):
-        self.assertEqual(genotypes.shape[0], inference.shape[0])
+    def verify_round_trip(self, genotypes, exclude_sites):
+        self.assertEqual(genotypes.shape[0], exclude_sites.shape[0])
         with tsinfer.SampleData() as sample_data:
             for j in range(genotypes.shape[0]):
-                sample_data.add_site(j, genotypes[j], inference=inference[j])
+                sample_data.add_site(j, genotypes[j])
+        exclude_positions = sample_data.sites_position[:][exclude_sites]
         for simplify in [False, True]:
-            output_ts = tsinfer.infer(sample_data, simplify=simplify)
+            output_ts = tsinfer.infer(
+                sample_data, simplify=simplify, exclude_positions=exclude_positions
+            )
             for tree in output_ts.trees():
                 for site in tree.sites():
+                    inf_type = json.loads(site.metadata)["inference_type"]
+                    if exclude_sites[site.id]:
+                        self.assertEqual(inf_type, tsinfer.INFERENCE_FITCH_PARSIMONY)
+                    else:
+                        self.assertEqual(inf_type, tsinfer.INFERENCE_FULL)
                     f = np.sum(genotypes[site.id])
                     if f == 0:
                         self.assertEqual(len(site.mutations), 0)
@@ -499,7 +502,7 @@ class TestNonInferenceSitesRoundTrip(unittest.TestCase):
         self.assertGreater(ts.num_sites, 2)
         genotypes = ts.genotype_matrix()
         inference = np.sum(genotypes, axis=1) > 1
-        self.verify_round_trip(genotypes, inference)
+        self.verify_round_trip(genotypes, ~inference)
 
     def test_half_sites_single_tree(self):
         ts = msprime.simulate(10, mutation_rate=5, random_seed=20)
@@ -507,7 +510,7 @@ class TestNonInferenceSitesRoundTrip(unittest.TestCase):
         genotypes = ts.genotype_matrix()
         inference = np.sum(genotypes, axis=1) > 1
         inference[::2] = False
-        self.verify_round_trip(genotypes, inference)
+        self.verify_round_trip(genotypes, ~inference)
 
     def test_simple_many_trees(self):
         ts = msprime.simulate(10, mutation_rate=5, recombination_rate=4, random_seed=10)
@@ -515,7 +518,7 @@ class TestNonInferenceSitesRoundTrip(unittest.TestCase):
         self.assertGreater(ts.num_sites, 2)
         genotypes = ts.genotype_matrix()
         inference = np.sum(genotypes, axis=1) > 1
-        self.verify_round_trip(genotypes, inference)
+        self.verify_round_trip(genotypes, ~inference)
 
     def test_half_sites_many_trees(self):
         ts = msprime.simulate(10, mutation_rate=5, recombination_rate=4, random_seed=11)
@@ -524,7 +527,7 @@ class TestNonInferenceSitesRoundTrip(unittest.TestCase):
         genotypes = ts.genotype_matrix()
         inference = np.sum(genotypes, axis=1) > 1
         inference[::2] = False
-        self.verify_round_trip(genotypes, inference)
+        self.verify_round_trip(genotypes, ~inference)
 
     def test_zero_inference_sites(self):
         ts = msprime.simulate(10, mutation_rate=5, recombination_rate=4, random_seed=21)
@@ -532,13 +535,13 @@ class TestNonInferenceSitesRoundTrip(unittest.TestCase):
         genotypes = ts.genotype_matrix()
         inference = np.sum(genotypes, axis=1) > 1
         inference[:] = False
-        self.verify_round_trip(genotypes, inference)
+        self.verify_round_trip(genotypes, ~inference)
 
     def test_random_data(self):
         genotypes, _ = get_random_data_example(20, 50, seed=12345)
         inference = np.sum(genotypes, axis=1) > 1
         inference[::2] = False
-        self.verify_round_trip(genotypes, inference)
+        self.verify_round_trip(genotypes, ~inference)
 
 
 class TestZeroNonInferenceSites(unittest.TestCase):
@@ -597,13 +600,19 @@ class TestZeroInferenceSites(unittest.TestCase):
         m = genotypes.shape[0]
         with tsinfer.SampleData(sequence_length=m + 1) as sample_data:
             for j in range(m):
-                sample_data.add_site(j, genotypes[j], inference=False)
-        self.assertEqual(sample_data.num_non_inference_sites, m)
-        self.assertEqual(sample_data.num_inference_sites, 0)
+                sample_data.add_site(j, genotypes[j])
+        exclude_positions = sample_data.sites_position
         for path_compression in [False, True]:
-            output_ts = tsinfer.infer(sample_data, path_compression=path_compression)
+            output_ts = tsinfer.infer(
+                sample_data,
+                path_compression=path_compression,
+                exclude_positions=exclude_positions,
+            )
             for tree in output_ts.trees():
                 self.assertEqual(tree.num_roots, 1)
+            for site in output_ts.sites():
+                inf_type = json.loads(site.metadata)["inference_type"]
+                self.assertEqual(inf_type, tsinfer.INFERENCE_FITCH_PARSIMONY)
 
     def test_many_sites(self):
         ts = msprime.simulate(10, mutation_rate=5, recombination_rate=4, random_seed=21)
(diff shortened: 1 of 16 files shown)


hyanwong:
Break early, break hard. I think that now is the time to make the change, before people start using tsinfer more widely. I agree that raising an error when accessing `.sites_inference[:]`, `num_inference_sites`, etc is the right thing to do. We should probably ping @awohns too, as some of his previous code relied on the concept of inference sites.

By the way, the new `tsinfer.infer()` code requires recombination_rate and mutation_rate to be arrays of length `num_inference_sites`, so this will need to be fixed asap if we are to test the variable rate code.

hyanwong:
Two more points:

1) we might want to work out if a site was used for inference *after* having run through the inference framework. At the moment, sites in a TS don't have flags, so we might need to store this in the metadata? Or implement site flags.

2) if we want the user to be able to mark a site as "not for inference", we could either do this during `add_site()` as we do now, or as a parameter passed to `tsinfer.infer` and friends. I'm inclined to the latter. That way we don't store anything like a `sites_inference[:]` array in the SampleData file. My preference would be for a parameter something like `skip_inference_sites=None` which would be a list of site IDs or a boolean array of length `num_sites`.

jeromekelleher:
OK, thanks Yan. We're on the same page here I think.

hyanwong:
Just re-reading this and thinking about flags. I first thought that we might want to mark a site as "used for inference", but now I'm wondering if it's actually a property of a *mutation* - e.g. if there are multiple mutations at a site, we could conceivably use one for inference but not the other. Is there any plan / possibility / other use for adding flags to the mutations table, do you think @benjeffery ? Or would we be happy placing this sort of thing in the site/mutation metadata, using a tsinfer-specific schema? It feels like a flag to me, but others may well disagree.

benjeffery:
The rule of thumb in my mind is that unless or until core tskit C API methods need to know which mutations are used for inference this belongs in metadata.

hyanwong:
Just putting a link in to issue #268 which is think is highly relevant. We probably want to implement a "mask" option for inference sites, so that we can pass in a list of `(start, end)` tuples, like we do for `delete_intervals`, which indicate the regions to be masked out during ancestral inference. Note that a similar "mask" idea may be required for `tsdate` (https://github.com/awohns/tsdate/issues/129), where we want to ignore certain regions for dating, but sill want to leave in singletons, which are normally skipped in the topology inference step.

I imagine that we might want to *combine* such a mask with setting `skip_inference_sites=[list, of, site, ids]`. We could either do this by having a few parameters to the inference functions, or by abstracting the function to a class (e.g. `class ts_mask`) which implements the combining capability as methods within the class.

@jeromekelleher - do you see a role for a (tskit) class which defines regions of a tree sequence, such that we could pass a "mask" object to `delete_intervals` and friends, as well as to `tsinfer.infer()` and the like? There's a nasty complication here that may only be relevant to inference, where we might want to combining regions of a TS as defined by positions, along with masking individual site positions. This could be an argument *for* a class that deals with any such nastiness, or an argument not to pollute tskit with extra intricacies. I'm not sure which.

I'm bringing this up now because it may be that something like this will be the basis of a decent ancestral inference pipeline . Given the time frame, it's probably best for  @awohns to hack round this for the time being by creating a mask of sites_inference-like booleans and `logical_and`ing it with the current inference sites if he wants to skip using these for inference. But it would be good to think about this now so that we can convert the hack to the "approved" way of doing things later.

hyanwong:
I guess we will want to check at some point which sites have been used for inference and which ones have not, which we would do by finding those in the sample data file that are not present in the ancestors tree sequence? It would be good to have a function that does this, I think.

hyanwong:
Actually, I was being dense. It's easy to do

```
sites_used = np.isin(sample_data.sites_position[:], ancestor_data.sites_position[:])
```

*although* - in the final tree sequence, I do think it would be useful to have an extra flag either on sites that were used for LS matching, or alternatively, on those that were placed via `map_mutations`. Is this OK, and have you any preference for which, and what the flag name should be, @jeromekelleher ?

jeromekelleher:
> Oh, but we don't have a flag column on the site or mutation tables, only the node table. So I guess this would have to go in metadata.

Putting it into metadata seems like a good idea. We could define a key "inference_type", which would be either "full" or "parsimony", or something along those lines.

jeromekelleher:
I've added a "inference_type" metadata key here to the output sites in the tree sequence - this has helped a lot actually, great idea. We can systematise this once tskit 0.3 is out, and maybe add in some more standard metadata.

I think this mostly works now, I just need to do some testing on the accessor methods in the SampleData file which used to have parameters for ``inference_sites``. I think we can just break these, tbh, and put in more sensible methods where we specify a set of sites we're interested in instread.

hyanwong:
> I've added a "inference_type" metadata key here to the output sites in the tree sequence - this has helped a lot actually, great idea. We can systematise this once tskit 0.3 is out, and maybe add in some more standard metadata.

Glad that helped. I'm wondering if the metadata is more properly on the *mutation* than the site. I could imagine us using one mutation for inference and putting another at the same site on using parsimony.

tests/test_inference.py
```diff
@@ -1964,6 +1959,20 @@ def test_many_trees(self):
...
 
+def get_default_inference_sites(sample_data):
+    """
+    Returns the site positions that would be used for inference by
+    default.
+    """
+    inference_sites = []
+    for var in sample_data.variants():
+        counts = tsinfer.allele_counts(var.genotypes)
+        assert len(var.site.alleles) == 2
```
hyanwong, line 1970: I don't get why this is an assert. Surely we allow triallelic etc sites, but just don't adde them to the returned inference sites.

tsinfer/formats.py
```diff
@@ -1597,24 +1585,13 @@ def add_site(
...
-                inference = False
-            if inference:
-                raise ValueError("Only biallelic sites supported for inference")
-            if n_alleles > 64:
-                # This is mandated by tskit's map_mutations function.
-                raise ValueError("Cannot have more than 64 alleles")
-        if counts.derived > 1 and counts.derived < counts.known:
-            if inference is None:
-                inference = True
+        if inference is None:
```
hyanwong, line 1588: Do we want to entirely remove the concept of setting inference sites when we build the sample data file, and only specify them (if we want to override the default) when matching ancestors instead? We can work out from the ancestors_ts which sites were used for inference to create that file, so we wouldn't need to re-specify them when doing match_samples.

The way I might think of doing this is to provide an array of genome positions to match_ancestors, giving the sites that should *not* be used for inference. All others would use the tsinfer default.

tests/test_inference.py
jeromekelleher: It's just a testing function that never gets called on anything more complicated. This seemed better than duplicating the logic of the main function.

jeromekelleher:
@hyanwong, you've convinced me to break this one hard and remove the idea of ``inference_sites`` entirely. There's no mention of it in the SampleData file. We now have an new option to ``infer``, called ``exclude_sites``. This lets us specify the sites we definitely don't want to use in the full inference algorithm. What do you think?

@awohns, is this going to break a bunch of stuff for you? How do you know how to mark an inference or non-inference site at the moment? Presumably this decision is made while the VCF is being processed?  I hate having to store stuff in external files, but maybe this is the least bad option for now until sgkit is ready for use.

jeromekelleher:
Note that I've bumped the major file version number here which will (in principle) mean a full rebuild of all existing files. This isn't strictly necessary, since all we're doing is ignoring part of the information in the old files. Maybe we'll back of that one.

hyanwong:
> @hyanwong, you've convinced me to break this one hard and remove the idea of `inference_sites` entirely. There's no mention of it in the SampleData file. We now have an new option to `infer`, called `exclude_sites`. This lets us specify the sites we definitely don't want to use in the full inference algorithm. What do you think?

I think if we want to do this, we should do it now, before it gets used by lots of people. And I think we probably *do* want to do this because (a) it's logical and (b) we might want to re-run inference excluding different sites. We can currently do (b) by editing the sample_data file, but that's very messy IMO. It will also make it very obvious that we are excluding stuff for a particular inference pathway. However, I wonder if we might want to be able to specify sites by *position* as well as (or instead of) the index into the sample_data file?

> @awohns, is this going to break a bunch of stuff for you? How do you know how to mark an inference or non-inference site at the moment? Presumably this decision is made while the VCF is being processed? I hate having to store stuff in external files, but maybe this is the least bad option for now until sgkit is ready for use.

I can probably answer here - it's decided if the ancestral state is of low quality. This is currently done by comparing to the ensembl FASTA file. Possibilities are either to save an exclude_positions file when creating the same_data file, or to zip through the FASTA file before inference and create the exclude-list like that.

hyanwong:
> Note that I've bumped the major file version number here which will (in principle) mean a full rebuild of all existing files. This isn't strictly necessary, since all we're doing is ignoring part of the information in the old files. Maybe we'll back of that one.

We are recreating them now anyway. The released version is already a major number behind, so I don't think it matters.

jeromekelleher:
> However, I wonder if we might want to be able to specify sites by position as well as (or instead of) the index into the sample_data file?

Yes, I agree, this'll be less error prone if we're bringing different data sources together. I'll finish this up so, unless @awohns starts shouting.

awohns:
Sorry to get in on this one late, looks like it'll be a big help!

> @awohns, is this going to break a bunch of stuff for you? How do you know how to mark an inference or non-inference site at the moment? Presumably this decision is made while the VCF is being processed? I hate having to store stuff in external files, but maybe this is the least bad option for now until sgkit is ready for use

As @hyanwong  said, I mark a site as inference or non-inference using the `inference` parameter in `add_site`. Singletons, n-1 tons, invariant get automatically marked as not for inference, so I pass `None`, low-quality ancestral state I manually mark as not for inference. Indels look like they're pretty rubbish when used for inference in 1000g so I might mark those as not for inference too.

As far as using inference_sites in my pipelines, I've simplified the way I do things so I don't think this will break anything actually.

hyanwong:
> As @hyanwong said, I mark a site as inference or non-inference using the `inference` parameter in `add_site`. 

So the breaking part of this PR is that there will no longer be an "inference" parameter in add_site. You will need to collect a list of the n-1 tons (I don't think these get marked as not-for-inference by default, actually) and low quality ancestral states, as well as indels if you want, and then pass this list on to generate_ancestors, rather than embed the information in the sample_data file.

> As far as using inference_sites in my pipelines, I've simplified the way I do things so I don't think this will break anything actually.

Great!

jeromekelleher:
OK, I think this is ready to go now guys; can you have a last look at the new interface please and see if you agree?

We might want to reconsider making (n-1)-tons non inference by default too. But, I guess it's hard to imagine doing this QC without making a second pass through the data. But, at least if we have a single "master" SampleData file with all the data from the VCF, then it's easy enough to go through using ``variants()`` it and make some choices about which sites we like and don't like.

hyanwong:
> OK, I think this is ready to go now guys; can you have a last look at the new interface please and see if you agree?

Will do.

> We might want to reconsider making (n-1)-tons non inference by default too. 

If anything, I think that's a bespoke option to generate_ancestors. It depends on the quality of the ancestral sites: n-1 tons do logically contain information, we tend to exclude them only because we think they are just places where we think ancestral allele is wrong. In fact, it would be quite reasonable to argue that we should exclude n-2 tons on this basis too.

In the future, I think it would be reasonable to have a parameter like `exclude="singletons"|"n-1-tons"|"something_else"` as a shorthand for the most likely types of sites we might want to exclude - the default being singletons, of course (I guess we *could* theoretically include them, but we expect it to make the inference slower for no gain)

> But, I guess it's hard to imagine doing this QC without making a second pass through the data. But, at least if we have a single "master" SampleData file with all the data from the VCF, then it's easy enough to go through using `variants()` it and make some choices about which sites we like and don't like.

Yes, thanks to zarr, even for large files I think it's quick to rip through a sample data file and collect a list of the the n-1 tons.

jeromekelleher:
> If anything, I think that's a bespoke option to generate_ancestors. It depends on the quality of the ancestral sites: n-1 tons do logically contain information, we tend to exclude them only because we think they are just places where we think ancestral allele is wrong. In fact, it would be quite reasonable to argue that we should exclude n-2 tons on this basis too.

OK, let's leave things as they are so. We'll need to make a pass through the data anyway, so no point in complicating things more at this point.

hyanwong:
If we are marking the full inference vs fitch parsimony sites with metadata, it would be pretty useful for us to summarise a tsinferred tree sequence in terms of the number of mutations at each sort of site. E.g.

```
python -m tsinfer ls my_inferred_ts.trees
path      = my_inferred_ts.trees
size      = 8.3 MiB
edges     = 118139
trees     = 5857
total sites = 40169
full inference sites     = xxxx
fitch inference sites     = yyyy
total mutations = 91265
full inference mutations     = XXXX
fitch inference mutations     = YYYY
```

As we currently allow `tsinfer list` to work on inferred tree sequences as well as sample data files and ancestors tree sequences, this could be a feature of `tsinfer list` that is different from plain old `tskit info`. In fact, perhaps this could be [what `tsinfer info` does](https://github.com/tskit-dev/tsinfer/issues/274), while `tsinfer list` simply lists the storage as it does currently.

jeromekelleher:
Good idea @hyanwong, let's mark this as something to do in a follow up. Can you open an issue to track please?

Since no one has disagreed, I'm going to merge this much so we can move on.

hyanwong:
> Good idea @hyanwong, let's mark this as something to do in a follow up. Can you open an issue to track please?

I'll add it to #274 

> Since no one has disagreed, I'm going to merge this much so we can move on.

Perfect, thanks.

PR #280, opened 2020-06-29
Check sd equality split

hyanwong:
The first commit in this PR splits the checking of equality between two sample data instances into a set of smaller tests. Although a little more wordy, I think it's both clear and more useful for debugging to have them written explicitly.

The second commit is more provisional, as previously discussed with @jeromekelleher , and provides a simple (undocumented) function which will be very helpful for reading in VCF files in parallel, something @awohns needs. I haven't provided any tests for this, but think I probably should do if the method is given the go ahead.

### DIFF
--- a/tests/test_formats.py
+++ b/tests/test_formats.py
@@ -108,11 +108,15 @@ class TestSampleData(unittest.TestCase, DataContainerMixin):
             tables.populations.add_row(metadata=pop_meta)  # One pop for each individual
 
         node_metadata = []
+        node_populations = tables.nodes.population
         for node in ts.nodes():
+            if node.is_sample():
+                node_populations[node.id] = node.id // ploidy
             if node.id % 3 == 0:  # Scatter metadata into nodes: once every 3rd row
                 node_metadata.append('{{"node id":{}}}'.format(node.id).encode())
             else:
                 node_metadata.append(b"")
+        tables.nodes.population = node_populations
         tables.nodes.packset_metadata(node_metadata)
 
         site_metadata = []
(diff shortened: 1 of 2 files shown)


jeromekelleher:
LGTM, thanks @hyanwong. A quick test for the ``append_sites`` function and we should be good. I think it's enough to test a few simple cases, and make sure we handle the case where one of the arguments has zero sites (this will happen in practise).

jeromekelleher:
Also, it might be handy to add a method ``assert_equal(self, other)`` which asserts the various functions you've just made. This would then tell you where the problem when writing tests.

hyanwong:
> make sure we handle the case where one of the arguments has zero sites (this will happen in practise).

I don't think this can happen because it is impossible to create a sample data file with zero sites (I just tried it), so we can't even create it to pass as an arg.

hyanwong:
> Also, it might be handy to add a method `assert_equal(self, other)` which asserts the various functions you've just made. This would then tell you where the problem when writing tests.

Should this be in format.py or in the test suite? If in format.py, in the tests we'd then have to do `self.assertRaises(ValueError, tc.assert_equal(other))`, right?

jeromekelleher:
> Should this be in format.py or in the test suite? If in format.py, in the tests we'd then have to do self.assertRaises(ValueError, tc.assert_equal(other)), right?

Can put it in format.py I guess and literally put in assert statements. It doesn't matter so much if it doesn't fit in with the unittest way of doing things, we don't expect it to fail much. No need for anything elaborate, and just don't bother if you don't think it's useful. Just a quick idea.

hyanwong:
> Can put it in format.py I guess and literally put in assert statements. It doesn't matter so much if it doesn't fit in with the unittest way of doing things, we don't expect it to fail much. No need for anything elaborate, and just don't bother if you don't think it's useful. Just a quick idea.

Right. I think it is a good idea, and in fact I just hit this use-case when writing the new tests.

hyanwong:
@jeromekelleher - would you like me to change some of the `self.assertTrue(tc1.data_equal(tc2))` asserts in other parts of the test suite to simply `tc1.assert_data_equal(tc2)`, which gives a better breakdown of *how* two table collections differ, if the test fails (but is not so unit-test-like)? Or, since everything passes now, shall I just let sleeping dogs lie, in which case I think this PR can be briefly reviewed and merged?

jeromekelleher:
Looks good, thanks @hyanwong. Let's let sleeping dogs lie!

Issue #276, opened 2020-06-15
Long ancient ancestors / slowness in HGDP analysis


hyanwong:
Because of the way we build ancestors, the oldest ancestors end up being extremely long (as nearly all sites are younger, and hence are skipped over when deciding when to stop building to left and right.

So far we have managed to get away with this because in the datasets we've previously been using there are few high frequency sites, but that's changing. For instance, in the 1000 genomes data there are 2120 sites at frequency N-2, but in the HGDP  there are over twice that number (4863). That means in HGDP there are a lot of ancient ancestors that span the entire genome (~17,000 out of 65000, versus 10,000 such sites out of 700,000 in the TGP). Here's a plot of the median length of ancestors binned by time of the ancestor (i.e. frequency of the focal site). You can see that in the HGDP there are a lot more full-length ancient ancestors - the pattern is the same regardless of whether we include or excluding missing sites, although because there are a sixth of the number of sites when excluding missing data, the actual number of long ancestors in that case is *relatively* fewer.

<img width="407" alt="Screenshot 2020-06-15 at 12 30 19" src="https://user-images.githubusercontent.com/4699014/84652717-fd709d00-af03-11ea-838a-6337ba107fe6.png">

Moreover, in the 1000G, the 10,000 full length ancestors are distributed over only 87 time values, which allows easy parallelisation. In HGDP with missing data the 17,000 full length ancestors occur over 2136 separate times, and about half these time points have only one ancestor. Note that if we exclude missing data from HGDP it's much more manageable: we only have 2987 full length ancestors over  70 time slices. 

Both the large number of long ancestors and the few of these ancestors per time point is causing inference to be *extremely* slow for HGDP with missing data. I've been running it using all 40 processes on holly for 40 hours days and we're only 2% through match ancestors. This is basically the same as @awohns found. It also explains why I didn't find any effect of including missing sites when I tried inference with smaller subsets of the HGDP, because by subsetting I end up with very few high freq sites, so that inferences with vs without missing data are comparable when normalising by the number of sites (i.e. there's nothing in the missing sites code *per se* that makes it slower).

We have several possibilities to speed this up:
1. (crude) exclude any sites from inference that have a freq greater that, say 99.9%.
2.  (slightly more sophisticated) exclude from inference only those at high frequency that also have some missing data (this would reduce the number of full length ancestor time points).
[...]

PR #283, opened 2020-07-02
Count lower freq conflicts

hyanwong:
Here's the alternative python code that reduces the sample set size when building ancestors even if the site is at lower freq (I think this is essentially performing the four gamete test between the focal site and the adjacent ones). It will probably make ancestor building slower, as we are checking every site now, not just the lower freq ones. But ancestor building is the quick part of inference anyway. Since this builds shorter ancestors, matching should become way faster, and we don't need the cutoff_exponent hack described in https://github.com/tskit-dev/tsinfer/issues/276. 

This seems to produce reasonable inference on both OOA simulations with injected error (sequencing error and ancestral state polarity error), and also with a subset of 10000 sites of HGDP chromosome 20. So I think it's worth coding in C to try on a larger example.

The excess length of ancestors produced by the C code (blue points below) is reduced substantially in this python version (red points):

image

### DIFF
--- a/docs/inference.rst
+++ b/docs/inference.rst
@@ -110,7 +110,7 @@ with other individuals.
 
 In ``tsinfer``, metadata can be stored by providing a JSON encodable
 mapping. This information is then stored as JSON, and embedded in the
-final tree sequence object and can be recovered using the ``msprime``
+final tree sequence object and can be recovered using the ``tskit``
 APIs.
 
 .. _sec_inference_import_samples:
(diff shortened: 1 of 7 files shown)


hyanwong:
An additional point: this code requires us to iterate not just through the set of active samples, S, but also through its inverse (to check if there are any derived mutations outside the set). At the C level, we might therefore want some sort of specific structure that allows us to transfer sample node IDs from one set to the other. There might be some short cuts that we can do to work this out, however, especially if we can maintain a list of derived counts for each inference site. Optimising this in C is going to require a little thought...

hyanwong:
I've just realised a better way to do this. Don't review yet.

hyanwong:
Closing this as we have realised that looking at lower freq conflicts means that we may be wrongly cutting down the older haplotypes on the basis of more recent recombinations. #467 is a simple alternative that seems to work.

PR #287, opened 2020-07-10
Change to keyword-only params

hyanwong:
For any more complicated params in the standard calls.

Although backwards-incompatible, forcing people to use keywords for e.g. num_threads and friends allows us to rationalise the order of params. We have already thought this was a good idea in tskit, so I'm starting with tsinfer as low-hanging fruit


jeromekelleher:
LGTM, thanks @hyanwong !

PR #288, opened 2020-07-12
Restrict samples

hyanwong:
This simply exposes a parameter to the user to restrict sample matching to a subset of samples. The functionality was already there, so I have not implemented any extra tests, but could do if required.

This would be useful for @awohns and the pipeline he has for matching only contemporaneous samples from a file. Would it be possible to merge this and the trivial associated [previous PR](#287) @jeromekelleher ?

### DIFF
--- a/tsinfer/inference.py
+++ b/tsinfer/inference.py
@@ -388,10 +388,11 @@ def match_samples(
     stabilise_node_ordering=False,
     engine=constants.C_ENGINE,
     progress_monitor=None,
+    samples=None,
 ):
     """
     match_samples(sample_data, ancestors_ts, *, num_threads=0, path_compression=True,\
-        simplify=True)
+        simplify=True, samples=None)
 
     Runs the sample matching :ref:`algorithm <sec_inference_match_samples>`
     on the specified :class:`SampleData` instance and ancestors tree sequence,
@@ -411,6 +412,8 @@ def match_samples(
         on a path between the root and any of the samples. To do so, the final tree
         sequence is simplified by appling the :meth:`tskit.TreeSequence.simplify` method
         with ``keep_unary`` set to True (default = ``True``).
+    :param array_like samples: An array of sample IDs from the sample_data file to
+        match, or None for all samples. Default: None.
     :return: The tree sequence representing the inferred history
         of the sample.
     :rtype: tskit.TreeSequence
@@ -428,7 +431,7 @@ def match_samples(
         engine=engine,
         progress_monitor=progress_monitor,
     )
-    manager.match_samples()
+    manager.match_samples(indexes=samples)
     ts = manager.finalise(
         simplify=simplify, stabilise_node_ordering=stabilise_node_ordering
     )
@@ -990,7 +993,7 @@ class SampleMatcher(Matcher):
         self.ancestors_ts_tables = ancestors_ts.dump_tables()
         super().__init__(sample_data, self.ancestors_ts_tables.sites.position, **kwargs)
         self.restore_tree_sequence_builder()
-        self.sample_ids = np.zeros(self.num_samples, dtype=np.int32)
+        self.sample_ids = np.full(self.num_samples, tskit.NULL, dtype=np.int32)
 
     def restore_tree_sequence_builder(self):
         tables = self.ancestors_ts_tables
@@ -1101,6 +1104,8 @@ class SampleMatcher(Matcher):
             indexes = np.arange(self.num_samples)
         # Add in sample nodes.
         for j in indexes:
+            if j < 0:
+                raise ValueError("Sample indexes must be positive")
             self.sample_ids[j] = self.tree_sequence_builder.add_node(0)
         logger.info("Started matching for {} samples".format(len(indexes)))
         if self.num_sites > 0:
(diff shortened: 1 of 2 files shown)


jeromekelleher:
LGTM, happy to merge after a rebase.

I just noticed that your commits aren't being associated with your GitHub account. Can you check your local git setup please? We want the contributions page to be accurate, and these commits won't count if they don't have the right email address in them. You can edit the email address during rebase for the second commit here. The first one has been merged already, so that one is lost I'm afraid.

hyanwong:
Thanks @jeromekelleher - Looks like it might be because I set up my new computer to use my BDI email, which wasn't listed as an email address in my GitHub account. I've added it, rebased, and it looks OK from my end now. Let me know if it still isn't associating correctly, and I'll have more of a dig into what's going on.

hyanwong:
Oh - don't merge this yet @jeromekelleher - I can't seem to get it working properly, so will stick a test case in to make sure I understand it properly.

hyanwong:
Commit now revised - I think I revealed a bug whereby the list of samples-to-use (`SampleMatcher.sample_ids`) was of length `num_samples` and initialised with zeros (which are a valid sample id). If only a few sample indexes were passed, we therefore had an array with the required IDs, plus lots of zeros, and ended up visiting sample 0 multiple times. Presumably we didn't spot it before because we always kept sample 0 when matching (and it was only in augment_ancestors anyway).

I've fixed it so that the array is filled with `tskit.NULL` instead, and made sure that we only use indices >=0 when using the values in `self.sample_ids`. I implemented a few tests, which now work, so I think this is good to merge, @jeromekelleher.

tsinfer/inference.py
```diff
@@ -990,7 +993,7 @@ def __init__(self, sample_data, ancestors_ts, **kwargs):
         self.ancestors_ts_tables = ancestors_ts.dump_tables()
         super().__init__(sample_data, self.ancestors_ts_tables.sites.position, **kwargs)
         self.restore_tree_sequence_builder()
-        self.sample_ids = np.zeros(self.num_samples, dtype=np.int32)
+        self.sample_ids = np.full(self.num_samples, tskit.NULL, dtype=np.int32)
```
jeromekelleher, line 996: Wouldn't it be simpler to have ``self.sample_ids = []`` here, and then fill these out below?

hyanwong: Yes, I can certainly do this. I was trying to retain the same approach as you had implemented previously, but it's certainly more obvious to create a dynamically sized array and fill it with what you need, rather than mark unused elements with tskit.NULL. Should I re-write using this approach then?

hyanwong:
> Is this urgent?

Well, I'm trying to sort out an approach for Wilder's iterative inference that doesn't use a bespoke version of tsinfer. At the moment he has to use a branch off my fork so that he can subset down the sample data files. He has stuff working, but it's a hack. So it's not blocking, but it is stopping us getting a sensible pipeline developed.

tsinfer/inference.py
hyanwong: I've just tried going here @jeromekelleher , and realise why it's not so easy. The sample_ids is a mapping between the row number in the sample_data file (`0..(num_samples-1)`) and the node ID in the tree sequence builder (which usually have IDs towards then end of the list of nodes, so are much larger numbers).

Instead of `self.sample_ids = []` I could use `self.sample_ids = {}` and fill it with a dictionary mapping row_id->node_id, but since it's usually the case that we are iterating over *all* the samples, that's probably a bit wasteful. In conclusion, I think that my suggestion of filling this with tskit.NULL values at the start, and ignoring any that are <0, is after all, the right one.

awohns:
Perhaps we should take some time to discuss how we want the iterative approach to coexist with tsinfer at some point. The iteration involves a variety of methods to glue together tsinfer, tsdate, and ancient samples: restricting which samples match in match_samples is one, relatively small, part of it. I'd really like to get the rest of the results out for the preprint, so my vote is that we develop tsinfer-related functions for the iterative approach with a slightly lower priority. On the other hand, Yan makes the important point that having the tsinfer-related functionality actually living in tsinfer makes for more robust, reproducible code, so it's definitely agree something I think should be done sometime soon-ish.
Also, just to clarify, I'm using the branch with the @hyanwong 's delete function and the improvements to ancestor generation. Those two things are quite separate from this PR though.

hyanwong:
> Perhaps, at some point, we should take some time to discuss how we want the iterative approach to coexist with tsinfer. 

Yes, this should be another issue, I think.

I would like this and #291  merged because that I think they are of more general use too, not just for non-contemporaneous samples. Note that once they are in, however, I do have a simple script to get some of the ancient ancestors stuff working.

I don't think we can really publish Wilder's paper if it relies upon my `sample data.delete()` branch, or the ancestor generation improvements (until they are merged, that is). We can probably get away with it for the preprint, but I'm wary of putting lots of work in to a bespoke pipeline that we will then have to change substantially before publication.

jeromekelleher:
OK, I'll push this branch through and check the details.

jeromekelleher:
I've updated the branch (although GitHub doesn't see it for some reason). I changed

1. The name to ``indexes`` for consitency with augment ancestors, and clarity within the code
2. Changed ``self.sample_ids`` to ``self.sample_id_map`` as a dict (if it becomes a performance issue we can change to array)
3. Fixed bugs in metadata handling.

jeromekelleher:
Great, merging. Thanks @hyanwong, this will be a useful utility for people until the tskit interfaces are done.

hyanwong:
Did you see my comments above @jeromekelleher ? I think we need `max(sample_ids)` and `min(sample_ids)`, not `sample_ids[-1]` and `sample_ids[0]`. This seems to have been merged though.

jeromekelleher:
No, didn't see any comments (GitHub has been very weird today). Can you log an issue if you think there's a bug there please (or a PR fixing it?)

hyanwong:
Ah - might have been that I forgot to click the "submit review" button. Doh. I'll make a PR. Fixes are simple

PR #290, opened 2020-07-14
Return the new ID from add_ancestor

hyanwong:
This is a useful addition if we are adding ancestors by hand, and want to keep track of the ID of the just-added one. We can't use `anc.num_ancestors` because the writing is buffered, so the array size may not be up to date. I'm guessing that `self.item_writer.total_items - 1` should always give the correct ID, even when using multiple threads for writing, and it seems to be fine in the single threaded case. Am I right here @jeromekelleher ?

If we can get this merged, then along with https://github.com/tskit-dev/tsinfer/pull/288 it's basically the last piece of the puzzle for getting a simple alternative to `match_ancestors` working that inserts non-contemporaneous samples as ancestors in the right place, so it would be good to get this merged.

### DIFF
--- a/tsinfer/formats.py
+++ b/tsinfer/formats.py
@@ -1959,7 +1959,8 @@ class AncestorData(DataContainer):
         """
         Adds an ancestor with the specified haplotype, with ancestral material
         over the interval [start:end], that is associated with the specified timepoint
-        and has new mutations at the specified list of focal sites.
+        and has new mutations at the specified list of focal sites. The id of the added
+        ancestor is returned
         """
         self._check_build_mode()
         haplotype = tskit.util.safe_np_int_cast(haplotype, dtype=np.int8, copy=True)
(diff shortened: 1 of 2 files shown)


tsinfer/formats.py
```diff
@@ -1987,6 +1988,7 @@ def add_ancestor(self, start, end, time, focal_sites, haplotype):
             focal_sites=focal_sites,
             haplotype=haplotype,
         )
+        return self.item_writer.total_items - 1
```
jeromekelleher, line 1991: Would be better for the item_writer to return the index, I think? Keeps encapsulation better, plus resolves any threading issues since you'd expect the writer to hold any locks needed during add.

jeromekelleher:
Yep, good idea and it's idiomatic. Minor comment above.

tsinfer/formats.py
hyanwong: Looks like it [already returns the index](https://github.com/tskit-dev/tsinfer/blob/8877a8f76af1bf5329bcd3a9bcb36ab2d7221e8b/tsinfer/formats.py#L214), so it's even more trivial than I thought!

jeromekelleher:
Great, ready to merge.

PR #291, opened 2020-07-14
Add a population object, and accessors

hyanwong:
We previously had no way to get a population from a sample data file. This implements functions for populations in the same way as currently exist for individuals (but more simply, as there is only metadata, and the definition of a population matches that in tskit)

### DIFF
--- a/tests/test_formats.py
+++ b/tests/test_formats.py
@@ -327,6 +327,30 @@ class TestSampleData(unittest.TestCase, DataContainerMixin):
         with formats.SampleData(sequence_length=ts.sequence_length) as sample_data:
             self.verify_data_round_trip(ts, sample_data)
 
+    def test_access_individuals(self):
+        ts = self.get_example_individuals_ts_with_metadata(5, 2, 10, 1)
+        sd = tsinfer.SampleData.from_tree_sequence(ts)
+        self.assertGreater(sd.num_individuals, 0)
+        has_some_metadata = False
+        for i, individual in enumerate(sd.individuals()):
+            if individual.metadata:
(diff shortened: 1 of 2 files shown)


jeromekelleher:
LGTM, thanks.

Needs a rebase though. In fact, the recent history is [a mess](https://github.com/tskit-dev/tsinfer/network). Can you please check that PRs are rebased against upstream? We will want to set up mergify here to automate this, but until then we need to manually rebase PRs so that we keep a linear history. (Non linear history is bad for a number of reasons).

hyanwong:
Rebased. I try to rebase the PRs against upstream when it changes, so not sure what in particular you are meaning? PR #256 is on a branch used by Wilder, but we are unsure whether to actually do anything with it, as it might be better left until we have a non-bespoke sample data format. #283 is draft, so I haven't kept it up to date with the changes. There are some other branches on my repo not in SPRs that Wilder is using too. Since I have loads of branches, I'm unlikely to keep them all rebased until they are ready to test/merge, I think.

jeromekelleher:
This is what the history looks like:

Screenshot from 2020-07-15 13-20-21

This is what is should look like: 

Screenshot from 2020-07-15 13-22-29

Each branch that we merge should have its base on the tip of master, and it should merge directly back in. This is a linear history. We have branches forking off and joining in all sorts of orders currently in tsinfer.

hyanwong:
Right - but those branches that look weird are not going to be merged (and many are not in SPRs anyway), so is this an issue?

jeromekelleher:
It's not the branches that aren't merged but the ones that are. Disregard the unmerged branches and compare the two screenshots.

PR #293, opened 2020-07-15
Subset function for SampleData.

jeromekelleher:
Here's an alternative approach to solving the underlying problem in #256 (letting us slice and dice the data). I think this basically works now, but it needs a good bit more testing. I'm hoping to get a bit of feedback on whether think this is a good idea.

Inference sites is a PITA, but I think the best thing to do is just kick the can down the road. We'll need to do a bit of an update to push the automatic behaviour into the tsinfer inference run-time rather than make this decision at data set build time, so we might as well leave it for now.

Any thoughts @hyanwong, @awohns?

### DIFF
--- a/tests/test_formats.py
+++ b/tests/test_formats.py
@@ -42,6 +42,9 @@ import tsinfer.formats as formats
 import tsinfer.exceptions as exceptions
 
 
+IS_WINDOWS = sys.platform == "win32"
+
+
 class DataContainerMixin(object):
     """
     Common tests for the the data container classes."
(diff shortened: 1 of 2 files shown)


awohns:
Thanks @jeromekelleher, why did you choose to subset by individuals rather than by samples? Works for real data, but in a lot of the simulations I've been doing I don't use individuals (could add them of course after the simulation, but I don't see it documented in msprime how to simulate with individuals).

hyanwong:
More generally, there is often confusion between indexes for individuals and for samples (there have been several msprime/SLiM mailing list posts where one was expected, but the other was provided. Since this is a source of error/confusion, I think we need to be pretty careful here.

Since sample indices are used more often, I wonder if one possibility is for any functions that take a list of individuals to require the indices to be provided in some sort of wrapping, e.g. a class of type `IndividualsList`, and to raise an error if not. Seems heavy handed, so perhaps there are better suggestions? I'll ask on the slack channel too.

jeromekelleher:
> First Q: why did you choose to subset by individuals rather than by samples? Works for real data, but in a lot of the simulations I've been doing I don't use individuals (could add them of course after the simulation, but I don't see it documented in msprime how to simulate with individuals). 

I started out with lists of samples, but it ended up being quite awkward. The way that the SampleData API is organised, you can't add samples without adding individuals first. Since individuals are at the top of the chain, it seems to make more sense to specify those. When we have haploid individuals (i.e., coming from simulations), the sample and individual IDs will be the same, so you can use the IDs directly.

For real data anyway, it's surely on a per-individual you'll want to work. It will be a bit limiting if you want to pull out say one haplotype for a given individual, but that's a bit of a corner case. Longer term, I'm hoping to drop the majority of this code and move to using [sgkit](https://github.com/pystatgen/sgkit/) as our input data format. This will solve a whole raft of problems around data ingestion and manipulation.

Re individuals, I think the main problem is that tskit's API support for individuals is weak at the moment, and that's something to build up.

jeromekelleher:
I think this is ready to go now. I also added a little bit of infrastructure around the getting Individuals to make it a bit less tedious. We can now get a list of sample IDs for the individual in an instance variable rather than have to look it up in a map.

We can add support for a ``samples`` argument later if we need it, but I think this should satisfy the vast majority of our requirements. It would be nice to nuke the ``inference_sites`` thing soon so that we're determining this at runtime rather than build-time, so let's not worry about the semantics of what happens to the ``inference_sites`` values during subsetting.

hyanwong:
Just had a look. Seems fine to me, although when I implemented it I tested it slightly differently, by comparing against the original tree sequences with `ts.delete_sites` to remove sites, and `ts.simplify` to remove samples. I guess your way is equivalent, though, so go ahead and merge unless you want me to stick in my approach to the tests too.

PR #294, opened 2020-07-15
use max and min rather than relying on order of sample_ids

hyanwong:
With the previous PR we assumed that the `sample_id_map.values()` are returned with the lowest first, but that's not necessarily the case. In particular, we are allowing sample indexes to be provided in any order, and it's only in Python3.7 that the `.values()` method returns the original order anyway. I'm pretty sure that it's max and min that we want.

### DIFF
--- a/tsinfer/inference.py
+++ b/tsinfer/inference.py
@@ -1314,7 +1314,7 @@ class SampleMatcher(Matcher):
             metadata_offset=tables.nodes.metadata_offset,
         )
         sample_ids = list(self.sample_id_map.values())
-        assert len(tables.nodes) == sample_ids[0]
+        assert len(tables.nodes) == min(sample_ids)
         samples_metadata = self.sample_data.samples_metadata[:]
         samples_population = self.sample_data.samples_population[:]
         samples_individual = self.sample_data.samples_individual[:]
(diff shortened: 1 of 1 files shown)


jeromekelleher:
Dicts are insertion order preserving since Python 3.6, and we are guaranteed to insert values in increasing order. So, what's the difference here? 

I see we're currently listing 3.6 as our latest version, but we'll change that to the oldest version before release. The rest of the tskit world is on minimum of 3.6, so that's the effective minimum anyway.

hyanwong:
Well, I thought that we didn't insist on indices being in increasing order when passed to the match_samples function, so I wanted to make sure. But I now see that if they are out of order, we bomb out in the `haplotypes()` call, so at least this isn't a bug if Python>=3.6. But I find the intention clearer using max and min, and I can imagine that the haplotypes call might in the future allow out-of-order indices, in which case the match_samples function will appear to allow unordered indices, but create a buggy response.

Not super vital, I guess, but it seemed to catch future potential problems and make the intention clear.

jeromekelleher:
The *values* in the ``sample_id_map`` (node IDs) must always be inserted in increasing order, regardless of the order in which input sample indexes (IDs in the sample data file) are in:

```
      for j in sample_indexes:                                                              
            self.sample_id_map[j] = self.tree_sequence_builder.add_node(0)
``` 

Adding a max and min just seems confusing when the list is guaranteed to be in increasing order. It would probably be better to store an explicit list of these at the same time, though, rather than making a list of of the dict values later on.

hyanwong:
Oh, right. I read that wrong. Sorry.

jeromekelleher:
Can we close this so?

PR #297, opened 2020-07-16
Check time of added ancestors

### LINKED ISSUE #286: Ancestors time order required? Or not documented
We [never seem to enforce](https://github.com/tskit-dev/tsinfer/blob/ba418b7aad37d2d4f8befce1b6378c1dc3427e7e/tsinfer/formats.py#L1971) the order of ancestors added to ancestors_data file, but inspecting one created, it looks like they are always in time order, with the oldest first. I assume this is required for later inference.

Should we 
a. check when we `add_ancestor` that the time of an added ancestor is less than or equal to the last added ancestor? OR
b. sort the ancestors on `finalise()`

a. is easier, b. would allow us to add ancestors in any order, e.g. adding additional ones into an existing file. I'm happy to implement b. if this is thought to be a good idea, @jeromekelleher

hyanwong:
Fixes #286

### DIFF
--- a/tsinfer/formats.py
+++ b/tsinfer/formats.py
@@ -1801,6 +1801,8 @@ class AncestorData(DataContainer):
         # Cache the num_sites value here as it's expensive to compute.
         self._num_sites = self.sample_data.num_inference_sites
         self._num_alleles = self.sample_data.num_alleles(inference_sites=True)
+        self._last_time = 0
+
         assert self._num_alleles.shape[0] == self._num_sites
         self.data.attrs["sample_data_uuid"] = sample_data.uuid
         if self.sample_data.sequence_length == 0:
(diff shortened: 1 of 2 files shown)


tsinfer/formats.py
```diff
@@ -1999,6 +1999,8 @@ def add_ancestor(self, start, end, time, focal_sites, haplotype):
             raise ValueError("haplotypes incorrect shape.")
         if np.any(haplotype >= self._num_alleles[start:end]):
             raise ValueError("haplotype values must be < num_alleles.")
+        # Ancestors must be added in decreasing time order, but buffering means we have
```
jeromekelleher, line 2002: I think it'd be more helpful to check at addition-time. We can do this in a similar way to what's done for sites - we keep the last site position in an instance variable, and check against that.

hyanwong: Good call. Amended.

PR #300, opened 2020-07-16
Ancestors as samples

hyanwong:
Here's a draft first pass at a `match_ancestors_including_noncontemporanous_samples`. It's just the general idea, so no tests implemented, but I've tried it out on some code in a Jupyter notebook, which I'll port into a test.

One question is whether to make this an option in the normal `match_ancestors` function or not. Another is whether it should all be in one long function like this (but I've tried splitting it up and putting it into other classes, and I can't quite get it to sit anywhere else).

### DIFF
--- a/tsinfer/inference.py
+++ b/tsinfer/inference.py
@@ -29,6 +29,7 @@ import threading
 import json
 import heapq
 
+import attr
 import numpy as np
 import humanize
 import tskit
@@ -97,6 +98,13 @@ def allele_counts(genotypes):
     return AlleleCounts(
         known=n_known, ancestral=n_ancestral, derived=n_known - n_ancestral
     )
+
+
+def exclude_id(attribute, value):
+    """
+    Used to filter out the id field from attrs objects such as Ancestor
+    """
+    return attribute.name != "id"
 
 
 class DummyProgress(object):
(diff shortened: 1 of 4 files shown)


tsinfer/inference.py
```diff
@@ -452,6 +460,136 @@ def match_samples(
...
+    current_anc = next(inferred_ancestors)
+    with formats.AncestorData(sample_data=sample_data) as new_data:
+        for insert_time, sample_ids in sorted(samples_by_timepoint.items()):
+            while current_anc is not None and current_anc.time > insert_time:
+                new_data.add_ancestor(**attr.asdict(current_anc, filter=exclude_id))
+                current_anc = next(inferred_ancestors, None)
+            for s in sample_ids:
+                logger.debug(f"Adding sample {s} as an ancestor")
+                haplotype = sample_data.sites_genotypes[:, s][sites_used]
+                if np.any(haplotype == tskit.MISSING_DATA):
```
awohns, line 521: Almost certain that this can be removed. @jeromekelleher , missing data in ancestors is supported, is that right? I can't find a test case in test_inference.py though

jeromekelleher: Should work fine, I probably just didn't test it explicitly.

hyanwong: I'm working through it in #306

hyanwong:
On some test data, when setting `create_sample_nodes=True`, this fails when trying to map mutations, and for an interesting reason. If a sample has been added as an ancestor, we only keep track of the state of sites marked for full inference, and discard those sites which will be later inferred using Fitch parsimony (e.g. we discard triallelic sites). *However*, when we get to `match_samples`, we need to know the state of the sample ancestor at these Fitch_parsimony sites. Since we haven't kept it from the match_ancestors stage, we can't work it out unless it's provided to match_samples somehow.

This would be a problem, for example, if we tried to match UKB samples, including some triallelic (or other non-full-inference sites) against a TGP ancestors tree sequence with Afansievo individuals included as real ancestors. In this case we would need to pass the full sample information on the Afansievo individuals along with the UKB individuals during match_samples.

There's no easy way around this, unless somehow the ancestors tree sequence can record all of the Fitch_parsimony sites too, but ignore them somehow during matching (and this is a *big* change). We could require the sample ancestors to be present in the sample_data object passed to match_samples, but then we would need a way to map the sample node in the ancestors_ts to the sample in the sample_data file.

This needs a bit of thought, so pinging both @jeromekelleher and @awohns to discuss while I'm away for the next few days.

awohns:
Can't we just go back to the sampledata file and use the states from that?

tsinfer/inference.py
```diff
@@ -490,6 +498,136 @@ def match_samples(
     return ts
 
 
+def match_ancestors_including_noncontemporanous_samples(
```
jeromekelleher, line 501: An obvious question here is, do we really need another function for this? Could we do the same thing as this via ``match_ancestors``? Or, more to the point, why aren't we doing this as a method of the AncestorData file, since the majority of what we're doing is just splicing in some extra haplotypes to create a new ancestor data file.

jeromekelleher:
cc @awohns - what do you think of the approach I've suggested above?

awohns:
That general approach sounds sounds to me @jeromekelleher, the `extract_ancestors` thing is basically the subset function isn't it, and then turning that `SampleData` subset into an `AncestorData` file. 

Re: terminology, I personally think 'ancient' instead of 'non-contemporaneous' is fine, but it does bias towards human data instead of e.g. HIV where I think they say time-series more than ancient

tsinfer/inference.py
hyanwong: I was thinking we could either (a) roll this into the match_ancestors function, or as you suggest (b) have an ancestor data merge function. The slight catch with (b) is that if we are using "true" sample ancestors, then we perhaps want to think of the thing in the ancestor_data file as a true genome, rather than an inferred ancestor - for example, we want an individual associated with the ancestor (we don't have a concept of individuals in the ancestor_data file), and we probably don't want the "ancestor_data_id=X" metadata field.

hyanwong:
> This looks good, but I think we're taking the wrong basic approach. More generally, we want to be able to update the AncestorData file with some extra haplotypes, spliced in at the right time. This is different to the augmented ancestors thing, as we're thinking about adding actual ancient data here. If we add a "flags" field to the AncestorData file we can differentiate these from run-of-the-mill ancestors and deal with them appropriately downstream in the standard inference pipeline.

See my [comment](https://github.com/tskit-dev/tsinfer/pull/300#discussion_r461452927) above - I can see the logic to what you suggest, but a "ancestor_is_real_sample" node has stuff associated with it that's not in the ancestor_data file, such as an individual with a population ID, etc. This is all taken directly from the sample_data file, so would either need to be stored in ancestors_data somehow (quite a big change) or we would need to provide the sample data to the routine that creates the ancestors_ts anyway (as I do in this PR)

> I find the "non-contemporaneous" terminology a bit wordy - is there a more concise way we can say this?

I completely agree. Many somewhat amusing options at https://www.wordhippo.com/what-is/the-opposite-of/present-day.html : the short options that struck me as not immediately ridiculous include "old", "past", "prior", or "dated", as in "ANCESTOR_IS_OLD_SAMPLE", "ANCESTOR_IS_PAST_SAMPLE", "ANCESTOR_IS_PRIOR_SAMPLE", "ANCESTOR_IS_DATED_SAMPLE". However, these alternatives all seem to have other meanings that risk confusion. Please do suggest other options, or argue for one of the above.

hyanwong:
Another thought: perhaps we should think of the ancestors file as containing only *inferred ancestors* (hence they have no individuals, populations, etc). Anything that creates an inferred ancestor does so by putting something into an ancestors file. We can think of creating a "proxy" sample ancestors (i.e. just the full inference sites, no associated individual) as a sort of inference, and hence proxy sample ancestors can be included in ancestor_data files.

On the other hand, putting a sample genome as a "real" ancestor is not a form of inference - it's simply sticking the genome in as-is. Hence real sample ancestors are not stored in an ancestor_data file, but in a sample_data file.

This suggests the following scheme to me: 

1. `generate_ancestors` creates an ancestor data file. We can add a parameter such as `include_historic` which creates *proxy* ancestors for any samples attached to historic individuals. In the final tree sequence, these proxy sample ancestors don't get marked as sample nodes, or have associated individuals - they are just helpful additional plausible ancestral genomes.

2. `match_ancestors` creates an ancestors ts. We add a parameter (which could have the same name, `include_historic` unless that's confusing) which creates *true* sample ancestors  for any samples attached to historic individuals. These true sample ancestors get marked as sample nodes in the final TS, and the individuals with which they are associated appears as individuals in the output tree sequences too.

3. `match samples` places sample genomes into the tree sequence. We add a parameter `map_mutations` (default=True) which if false, does not place fitch parsimony mutations onto sites.

4. we create a new helper function, `map_mutations(ts, sample_data)`, which can be used to place the fitch parsimony sites if it is not done in step 3. This allow us to run step 2 with some true sample ancestors, match only the modern samples in step 3 (using [`indexes=moderns`](https://github.com/tskit-dev/tsinfer/pull/288)), then place the Fitch sites using the full sample data file.

This is all starting to make logical sense to me. What do you think @awohns , @jeromekelleher ?

tsinfer/inference.py
```diff
@@ -490,6 +498,136 @@ def match_samples(
...
+                    end=ancestor_data.num_sites,
+                    time=insert_time,
+                    focal_sites=[],
+                    haplotype=haplotype,
+                )
+                ancestor_to_sample_map[anc_id] = s
+
+    logger.info("Matching ancestors including noncontemporaneous samples")
+    ancestor_tables = match_ancestors(
+        sample_data, ancestor_data, **kwargs
```
hyanwong, line 576: I think this should be `match_ancestors(
         new_data, ancestor_data, **kwargs)`

awohns:
Fixing the bug mentioned by @hyanwong on line 576 introduces new problems. The new ancestordata file has the wrong number of sites, a slighly hacky way to fix this is to use sampledata.subset to subset down to only the full inference sites

awohns:
@hyanwong, what do you think the best way to finish up this PR is? There was talk earlier about a different approach, alternatively I think we can fix the bug I mentioned above and it would be about ready to go

hyanwong:
Currently closed in favour of the suggestion at https://github.com/tskit-dev/tsinfer/issues/328

Issue #248, opened 2020-03-18
Support missing data in LS copying process


jeromekelleher:
We want to be able to support small amounts of missing data in the LS copying process. This will be indicated by a -1 in the genotype for a given sample at a give site. As implemented in https://github.com/tskit-dev/tskit/pull/489, we can update the copying process quite simply to accommodate missing data. We can then either impute (the default, I'd imagine) or preserve the missingness by snipping the edges where the haplotype for a given sample is missing.

Any thoughts @hyanwon?

Related to #169

hyanwong:
We discussed the idea of imputing everything (easiest), then having a separate routine that masks out the missing sites by snipping out edges. This seems sensible to me. One question is whether the normal inference process has an "impute_missing" flag which runs this routine as a final step.

hyanwong:
To do the snipping, we need access to the original SampleData file as well as the inferred TS. Given that the SampleData file is a necessary input to `match_samples`, it seems sensible to have a flag on that method that also runs the "missingness-snipping" routine as a final step.

jeromekelleher:
We can do it inline as well during ``match_samples`` --- it's quite straightforward to do actually, so probably simplest to add it as an option ``impute_missing=True`` to ``match_samples`` I think.

hyanwong:
A follow-up to this issue: we might like to allow missing data within an ancestor (e.g. if we are using an ancient sample as an ancestor). From [this comment](https://github.com/tskit-dev/tsinfer/blob/264861c1497ddb21f1f4d39634b8f8d265155d60/tests/test_inference.py#L993) in the test suite, it looks as if a site is all missing when reconstructing normally (i.e. without ancients), we set then site value to `0` rather than `-1`. Yet @awohns thinks that sites marked as missing in the middle of an ancestor *are* allowed, in which case I assume we should be marking these as `-1`? Can you confirm whether sites marked as `tskit.MISSING_DATA` within the middle of an otherwise intact ancestor are just skipped over (or equivalent) by the matching process,  @jeromekelleher ?

jeromekelleher:
Missing data should be treated the same in samples and within ancestor haplotypes (ie., in the middle), so I don't see any reason this wouldn't work. I think we can close this issue all right, this functionality was implemented a while ago.

hyanwong:
So it looks like we should make the following change in the ancestor builder, @jeromekelleher:
 
[...]

PR #306, opened 2020-07-22
Allow missing data within an ancestor

hyanwong:
Following [this comment](https://github.com/tskit-dev/tsinfer/issues/248#issuecomment-662301218), this sticks `tskit.MISSING_DATA` as the value within a haplotype if we really do have no clue as to the possible haplotype value. @awohns is testing it with some real data right now.

### DIFF
--- a/_tsinfermodule.c
+++ b/_tsinfermodule.c
@@ -164,6 +164,7 @@ static PyObject *
 AncestorBuilder_make_ancestor(AncestorBuilder *self, PyObject *args, PyObject *kwds)
 {
     int err;
+    PyObject *ret = NULL;
     static char *kwlist[] = {"focal_sites", "ancestor", NULL};
     PyObject *ancestor = NULL;
     PyArrayObject *ancestor_array = NULL;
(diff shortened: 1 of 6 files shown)


hyanwong:
This is failing because (I think) the tests are wrong. When there is missing data, it is possible that the last site in a reconstructed haplotype will be tskit.MISSING (if all the focal haplotypes are missing at that point). So the assertions at https://github.com/tskit-dev/tsinfer/blob/93fac8ff515bb0b618d416ad6c7491a307ddf377/lib/ancestor_builder.c#L382 and https://github.com/tskit-dev/tsinfer/blob/93fac8ff515bb0b618d416ad6c7491a307ddf377/lib/ancestor_builder.c#L391 are sometimes false. Can I simply remove the assertions, do you think @jeromekelleher ?

hyanwong:
Even after removing the assertions mentioned above, the C tests fail on e.g. `test_random_data_n5_m3`, because `-1` is being added as a mutation_derived_state [here](https://github.com/tskit-dev/tsinfer/blob/93fac8ff515bb0b618d416ad6c7491a307ddf377/lib/tests/tests.c#L276). I'm not that familiar with what is going on in the C tests, so would appreciate a hand figuring this out, @jeromekelleher

hyanwong:
This is ready for review @jeromekelleher - but I haven't unit tested the line that raises an error in the python code, as I'm not sure how were would trigger it from the outside - the only way I can think of to force reconstruction of an ancestor at a focal site with all zeros is to start making a bespoke `AncestorsGenerator` object and force it to do things it normally wouldn't.

hyanwong:
(I just added a rather cumbersome test for the python error raising, as a separate commit - I can't get it to work with the C engine, oddly)

PR #307, opened 2020-07-22
Add check that singletons are marked as fitch_parsimony

hyanwong:
It wasn't clear to me that this was explicitly covered in the tests, because the TestNonInferenceSitesRoundTrip class explicitly passes in the list of sites to ignore (including singletons). This checks that with the default settings, singleton sites are fitch_parsimony and others are full_parsimony.


jeromekelleher:
LGTM, thanks.

Issue #103, opened 2018-08-30
Change sample_data and ancestor_data to samples and ancestors


jeromekelleher:
This will clean up the API a little bit. Also for the CLI.

jeromekelleher:
Let's not bother with this.

PR #309, opened 2020-07-22
Save ancestor id

### LINKED ISSUE #301: Match Ancestors: Map ancestor ids
In @hyanwong 's PR for matching noncontemporaneous samples, we need to keep track of the ID of the sample which was added as an ancestor. We can get the id of the ancestor that was added, but the ids change during match ancestors when path compression ancestors are added.

We think a extra parameter to match_ancestors which returns a mapping of node ids to ancestor ids would be helpful here, and better than simply iterating through the ancestors ts and adding +1 when we see a PC ancestor.
Thoughts, @jeromekelleher ?

hyanwong:
Fixes #301 - Note that the way I assign the IDs into the nodes seems a little cumbersome to me: see comment below. Perhaps there's a better way?

### DIFF
--- a/tests/test_inference.py
+++ b/tests/test_inference.py
@@ -1278,6 +1278,19 @@ class TestAncestorsTreeSequence(unittest.TestCase):
                 )
                 self.assertEqual(ancestor_data.provenances_timestamp[j], p.timestamp)
 
+            # Ancestors indicated in node metadata should have the same age as their node
+            ancestors_time = ancestor_data.ancestors_time[:]
+            num_ancestor_nodes = 0
+            for n in ancestors_ts.nodes():
+                md = json.loads(n.metadata) if n.metadata else {}
+                if tsinfer.is_pc_ancestor(n.flags):
+                    self.assertFalse("ancestor_data_id" in md)
+                else:
+                    self.assertTrue("ancestor_data_id" in md)
+                    self.assertEqual(ancestors_time[md["ancestor_data_id"]], n.time)
+                    num_ancestor_nodes += 1
+            self.assertEqual(num_ancestor_nodes, ancestor_data.num_ancestors)
+
     def test_no_recombination(self):
         ts = msprime.simulate(10, mutation_rate=2, random_seed=234)
         sample_data = tsinfer.SampleData.from_tree_sequence(ts)
@@ -2677,7 +2690,7 @@ class TestAugmentedAncestors(unittest.TestCase):
             self.assertEqual(node.flags, tsinfer.NODE_IS_SAMPLE_ANCESTOR)
             self.assertEqual(node.time, 1)
             metadata = json.loads(node.metadata.decode())
-            self.assertEqual(node_id, metadata["sample"])
+            self.assertEqual(node_id, metadata["sample_data_id"])
 
         t2.nodes.truncate(len(t1.nodes))
         # Adding and subtracting 1 can lead to small diffs, so we compare
@@ -2830,7 +2843,7 @@ class TestSequentialAugmentedAncestors(TestAugmentedAncestors):
             for node in final_ts.nodes():
                 if node.flags == tsinfer.NODE_IS_SAMPLE_ANCESTOR:
                     metadata = json.loads(node.metadata.decode())
-                    self.assertIn(metadata["sample"], subset)
+                    self.assertIn(metadata["sample_data_id"], subset)
                     num_sample_ancestors += 1
             self.assertEqual(expected_sample_ancestors, num_sample_ancestors)
             tsinfer.verify(samples, final_ts.simplify())
--- a/tsinfer/inference.py
+++ b/tsinfer/inference.py
@@ -73,7 +73,7 @@ def count_pc_ancestors(flags):
     NODE_IS_PC_ANCESTOR set.
     """
     flags = np.array(flags, dtype=np.uint32, copy=False)
-    return np.sum(np.bitwise_and(flags, constants.NODE_IS_PC_ANCESTOR) != 0)
+    return np.sum(is_pc_ancestor(flags))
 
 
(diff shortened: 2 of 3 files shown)


tsinfer/inference.py
```diff
@@ -1019,10 +1020,19 @@ def get_ancestors_tree_sequence(self):
...
 
+        # Add metadata for any non-PC node, pointing to the original ancestor
+        metadata = []
+        ancestor = 0
+        for is_pc in pc_ancestors:
+            if is_pc:
+                metadata.append(b"")
+            else:
+                metadata.append(self.encode_metadata({"ancestor": ancestor}))
+                ancestor += 1
```
hyanwong, line 1034: Is there a tidier way to do this?

tsinfer/inference.py
```diff
@@ -73,7 +73,7 @@ def count_pc_ancestors(flags):
     NODE_IS_PC_ANCESTOR set.
     """
     flags = np.array(flags, dtype=np.uint32, copy=False)
-    return np.sum(np.bitwise_and(flags, constants.NODE_IS_PC_ANCESTOR) != 0)
+    return np.sum(is_pc_ancestor(flags))
```
hyanwong, line 76: I added this in as it seemed a sensible reuse of the code

tsinfer/inference.py
```diff
@@ -1476,14 +1486,16 @@ def get_augmented_ancestors_tree_sequence(self, sample_indexes):
                 )
                 s += 1
             else:
+                # This is a path compressed node
                 tables.nodes.add_row(flags=flags[j], time=times[j])
+                assert (flags[j] & constants.NODE_IS_PC_ANCESTOR) != 0
```
hyanwong, line 1491: I thought better to be explicit here about what we expect

tsinfer/inference.py
jeromekelleher: Not sure there's much to be done here, this is as tidy as it'll get I'd say.

tests/test_inference.py
```diff
@@ -1273,6 +1273,17 @@ def verify(self, sample_data, mismatch_rate=None, recombination_rate=None):
                 )
                 self.assertEqual(ancestor_data.provenances_timestamp[j], p.timestamp)
 
+            # Ancestors indicated in node metadata should have the same age as their node
+            ancestors_time = ancestor_data.ancestors_time[:]
+            num_ancestor_nodes = 0
+            for n in ancestors_ts.nodes():
+                if n.metadata:
+                    md = json.loads(n.metadata)
+                    if "ancestor" in md:
```
jeromekelleher, line 1282: We should be testing whether this is present for non-pc ancestors using the flags, right? Currently this would pass if no metadata was added at all.

hyanwong: I'm not sure what you mean here, sorry. If no metadata was added at all, then at the end, num_ancestor_nodes would be 0, so the test would fail. I could assert that the nodes with metadata containing 'ancestor' are *not* PC nodes? But there isn't a flag that marks whether an node in the ancestors ts corresponds to an ancestor in the file. That's the point of the metadata, I guess.

jeromekelleher: Well, every ancestor is either a PC ancestor OR an ancestor from the file, right? So, if the PC flag is not set, then metadata must contain this "ancestor" key. If it is, it must not contain this ancestor key.

hyanwong: Right, so I've just pushed a change to check that the PC flag is not set on these ancestors. I could also check that PC nodes do *not* have that metadata field, I guess? Would you like that additional check? I think the assert at the end checks that we have caught all the ancestor nodes.

jeromekelleher: Yes - what we want is:
```
if is_pc_ancestor(node):
     assert "ancestor" not in md
else:
      assert "ancestor" in md
```
Although, looking at it now, "ancestor" is perhaps not the best metadata key. What about ``ancestor_data_id`` - it's clear then that this is it's ID in the ancestor data file.

hyanwong: Yes, I chose 'ancestor' because we use 'sample' as a metadata key in `get_augmented_ancestors_tree_sequence`, so I thought that should match. Perhaps we should change both of them to have `_id` on the end?

hyanwong: > Yes - what we want is:
> 
> ```
> if is_pc_ancestor(node):
>      assert "ancestor" not in md
> else:
>       assert "ancestor" in md
> ```

I was trying to avoid being exhaustive, in case other nodes are will be added in the future (e.g. there is some mention of SRB nodes). I really just wanted to check that the nodes with 'ancestor' were valid, rather than validate at all the nodes.

jeromekelleher: We should validate all nodes. If we change things in the future the tests *should* break - that's part of the point.

hyanwong: Hmm, OK. Do you want both "ancestors" and "samples" metadata keys changed too?

jeromekelleher: Yes, might be good to be explicit and change these to ``ancestor_data_id`` and ``sample_data_id`` to be clear that these are the IDs within the given files.

hyanwong: If we are keeping `sample_data` and `ancestor_data`, you might want to close https://github.com/tskit-dev/tsinfer/issues/103, @jeromekelleher

hyanwong:
Metadata fields changed, and nodes all checked. Should be ready for merging now.

tests/test_inference.py
jeromekelleher: True, done.

jeromekelleher:
Merged, thanks @hyanwong. Feels like the data model is pulling together much more tightly!

PR #310, opened 2020-07-22
Use the seeded rng

hyanwong:
I noticed that the tests in https://github.com/tskit-dev/tsinfer/pull/306 passed on my machine but not on CI, because the wrong RNG was being used in tests (so that test results varied from one run to another). Oops. This corrects the behaviour.

### DIFF
--- a/evaluation.py
+++ b/evaluation.py
@@ -717,7 +717,7 @@ def run_imputation_accuracy(args):
             work.append((sim_args, missing_proportion))
             # imputation_accuracy_worker((sim_args, missing_proportion))
 
-    random.shuffle(work)
+    rng.shuffle(work)
     progress = tqdm.tqdm(total=len(work), disable=not args.progress)
     results = []
     try:


hyanwong:
> I noticed that the tests in #306 passed on my machine but not on CI, because the wrong rng was being used in tests (so that test results varied from one run to another). Oops. This corrects the behaviour.

jeromekelleher:
Thanks, we should use  the seeded rng. Slightly concerning that the test failed though?

hyanwong:
Yep - discussion in https://github.com/tskit-dev/tsinfer/pull/306#issuecomment-662679906

jeromekelleher:
Looks like this broke tests in Travis (weirdly): https://travis-ci.org/github/tskit-dev/tsinfer/builds/711018860

hyanwong:
Looks like Travis was broken for a bit - will probably work if forced to re-run.

PR #313, opened 2020-07-23
Move population to Individual in SampleData.

### LINKED ISSUE #312: Rationalise SampleData model
There's a few places where the data model doesn't make sense, which we should fix while we're already breaking things with a major version bump.

The main thing that seems to not make much sense is how we're attaching data to ``samples``:
```
     ("samples/individual", zarr_summary(self.samples_individual)),
      ("samples/population", zarr_summary(self.samples_population)),
      ("samples/metadata", zarr_summary(self.samples_metadata)),
```

It definitely makes no sense to store population for each sample, and we're assuming that this is going to be the same for all samples in an individual in a number of places.

It's not clear to me why we'd want to store metadata on a per-sample basis, rather than on an individual. I think we probably kept this for compatibility at some point, and we should just drop it.

So, I vote we drop both ``population`` and ``metadata`` from samples, so that samples is just a way of keeping track of individuals with different ploidy that maps into the tskit data model nicely.

Any objections @hyanwong, @awohns? I think it'd be good to do this quickly after #308 drops, so that we can get these format changes out of the way before any more files are built.

jeromekelleher:
Closes #312.

I decided against removing the samples metadata, since it's possible there may be some application for this, and it's not really any burden to keep it.

### DIFF
--- a/tsinfer/formats.py
+++ b/tsinfer/formats.py
@@ -735,6 +735,7 @@ class Individual(object):
     location = attr.ib()
     time = attr.ib()
     metadata = attr.ib()
+    population = attr.ib()
     samples = attr.ib()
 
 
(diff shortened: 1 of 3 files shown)


jeromekelleher:
I don't think this'll be controversial, so I'm going to merge before anyone has a chance to build files that won't work.

tsinfer/formats.py
```diff
@@ -1033,8 +1039,8 @@ def individuals_time(self):
         return self.data["individual/time"]
 
     @property
-    def samples_population(self):
-        return self.data["samples/population"]
+    def individuals_population(self):
+        return self.data["individual/population"]
```
hyanwong, line 1043: This should be "individuals/population" with an "s", surely?

Issue #316, opened 2020-07-23
Fix samples iterator


jeromekelleher:
https://github.com/tskit-dev/tsinfer/commit/11cb3c033eba02216029f290576a3ee20df6874a#r40853295

Error introduced in #308

hyanwong:
I had to implement a fix and there were a few things missing from your original PR, @jeromekelleher  - fix is at https://github.com/tskit-dev/tsinfer/pull/320

hyanwong:
This can be closed now that #320 is merged

PR #320, opened 2020-07-28
Remove sample refs to population & fix plurals

### LINKED ISSUE #319: Unexpected behavior with individuals_population
After the recent change referenced in #313, I've run into unexpected behaviour (I think a bug) when making SampleData files. It threw me for a loop for a long time, but I think I've isolated the issue.

When adding > 512 individuals with populations specified to a sampledata file, the individuals_populations Zarr array begins to record populations incorrectly. See the following trivial example:

```
import tsinfer
import numpy as np
with tsinfer.SampleData(sequence_length=10) as samples:
    samples.add_population()
    samples.add_population()
    [samples.add_individual(ploidy=2, population=1) for i in range(513)]
    samples.add_site(position=2, genotypes=np.zeros(513*2).astype(int))

np.unique(samples.individuals_population[:], return_counts=True)
>>>(array([0, 1], dtype=int32), array([513, 513]))
```
If 512 individuals are used instead of 513, the entire `samples.individuals_population[:]` array is filled with population 1, as expected.
Is something going on here with the Zarr array expecting inputs < 1kb?

hyanwong:
Fixes #319 and #316

### DIFF
--- a/tests/test_formats.py
+++ b/tests/test_formats.py
@@ -1431,6 +1431,9 @@ class TestSampleDataMerge(unittest.TestCase):
             self.assertEqual(new_ind.location, old_ind.location)
             self.assertEqual(new_ind.time, old_ind.time)
             self.assertEqual(new_ind.metadata, old_ind.metadata)
+            self.assertEqual(
+                new_ind.population, old_ind.population + sd1.num_populations
+            )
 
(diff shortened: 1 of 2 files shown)


jeromekelleher:
Merging, thanks @hyanwong. That was a sloppy update from me, oops!

PR #321, opened 2020-07-30
Use consensus at terminus of ancestor

### LINKED ISSUE #126: Minor inconsistency in ancestral state computation at terminal site
At the moment, when the stopping criterion is reached in the compute_ancestral_states function (in algorithm.py) the routine stops and the ancestral state at that site is set to 0. This is slightly inconsistent: it should probably be set to the consensus state instead, by moving the line `a[l] = consensus` to above the `break` statement.

hyanwong:
Fixes #126

### DIFF
--- a/lib/ancestor_builder.c
+++ b/lib/ancestor_builder.c
@@ -276,6 +276,7 @@ ancestor_builder_compute_ancestral_states(ancestor_builder_t *self, int directio
                         sample_set[j] = -1;
                     }
                 }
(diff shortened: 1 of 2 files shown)


jeromekelleher:
Yep, LGTM. This was how it worked before, the recent updates as well, right?

hyanwong:
It's always been a bit inconsistent, like it was before this PR. After this PR is the first time tsinfer has ever used the consensus at the terminal site.

Issue #314, opened 2020-07-23
Mark non-contemporaneous samples with a flag?


hyanwong:
At the moment, when we do a "normal" inference, we put all the sample nodes at time 0, even if they are from a sample associated with an ancient individual. There's no helping this in the normal inference, but it would be useful to know in the resulting tree sequence which nodes are not from contemporaneous individuals. I propose to add a new flag, 'NODE_IS_NONCONTEMPORANEOUS_SAMPLE' for these nodes, so they can be easily identified. What do you think, @jeromekelleher ?

jeromekelleher:
Since this is something happening external to tsinfer, it's probably better to put this information as metadata in the sample data file? Unless we're going to have a function for doing this?

hyanwong:
I don't think it is external to tsinfer? When we make a SampleData file we can give times to individuals. When we infer using the standard tsinfer pathway, we then put samples from these individuals at time (=freq) 0. This is somewhat confusing, and I think flagging it up somehow is a generic part of tsinfer, rather than e.g. a tsdate specific thing?

jeromekelleher:
Right, that's true. We should probably save the time that we originally associated with them in that case though - this is a useful value, presumably? So, we add a ``sample_data_time`` or something key to the metadata, and store that value in there?

hyanwong:
Yes, I think we should have a flag *and* a little piece of metadata, as you suggest, in case people want further detail.

jeromekelleher:
closed in #322

PR #322, opened 2020-07-30
individuals_time carries through into inferred tree sequence

### LINKED ISSUE #232: Incorporate individuals_time into the inferred tree sequence sample nodes
Thanks to #190, the sampledata file has the ability to store a time for an individual (as well as times for variants). The time for an individual is, however, currently ignored.

As part of tsdate, we'd like to be able to obtain the time for a non-contemporaneous sample node, and I think the sensible thing to do is to get this out of the inferred tree sequence. In other words, I would like a tree sequence with the sample nodes placed at the times specified in the `individuals_time` field.

This mean (a) placing the sample nodes at a time in the resulting ts as specified by the individuals_time of the individual with which they are associated AND (b) ensuring that these sample node times do not conflict with the internal node times imposed by tsdate (e.g. when using the frequency of the focal variants as a proxy for time).

One slighly hacky possibility, if the frequency of focal variants is used as the node time (i.e. if  variant times are not user-specified), is to find the oldest `individuals_time` value, and add that to all the variant frequencies (or multiply the frequency values by some large value) so that all ancestors are older than the oldest sample. Of course, if the user specifies a variant age when inputting the sites, then we shouldn't alter it, but rather throw an error if a variant is younger than the age of a sample (individual) containing that derived variant.

Also see https://github.com/tskit-dev/tsinfer/issues/207

awohns:
Resolves #232  and #314 

If any individuals in a `sampledata` file are historic (their time != 0), this information currently does not carry through into the inferred tree sequence. It would be very helpful for our iteration approach if it did.

If historic individuals are detected, we add two things to the resulting inferred tree sequence:
1. A new node flag, called `NODE_IS_HISTORIC_SAMPLE`, where bit 20 is set. Bits 18 and 19 will be used in #300 
2. An individual's metadata field called `sample_data_time`, where the actual time is recorded. 

I've included a roundtrip test to ensure that this carries through, I'm sure there are other good tests to write, any suggestions @hyanwong and @jeromekelleher ?

### DIFF
--- a/tests/test_inference.py
+++ b/tests/test_inference.py
@@ -904,7 +904,6 @@ class TestMetadataRoundTrip(unittest.TestCase):
 
     def test_historic_individuals(self):
         samples = [msprime.Sample(population=0, time=0) for i in range(10)]
-        random.seed(16)
         rng = random.Random(32)
         ages = [rng.random(), rng.random()]
         historic_samples = [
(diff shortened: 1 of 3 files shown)


jeromekelleher:
This looks good @awohns. I've gone through and fixed some issues with the tests, which weren't working (if your tests pass first time, be suspicious!)

Will merge once CI passes.

PR #323, opened 2020-07-31
Map mutations func

### LINKED ISSUE #298: Mapping non-inference mutations as an independent function?
In several cases we want to [union](https://github.com/tskit-dev/tskit/pull/623) some inferred tree sequences together. For example, we might have matched UK-Biobank against an ancestors tree sequence to get a tree sequence TS_1, then matched the 1000 genomes samples against the same ancestors tree sequence to get another tree sequence, TS_2. When we union TS_1 and TS_2 together, the edges and nodes are the same, but the non-inference sites may have been mapped differently, so the union fails.

One way round this is to do the non-inference mapping separately, and make it possible to turn this off (with the warning in the docs that the haplotypes in the resulting TS will not be identical to the input if off). The function to add the non-inference sites could then be made publicly available, so it could be run after the fact. Alternatively, we could remove the non-inference-site mutations, and put them back on afterwards. Or lastly, and perhaps most sensibly, we could implement a tsinfer-compatible-union which takes not only a mapping of nodes from one TS to the other, but also list of sites whose mutations will be zapped and remapped to make consistent.

hyanwong:
Fixes #298, and the changes in https://github.com/tskit-dev/tsinfer/commit/c4e89cfe30385ec682af15d2bb2a22c34bff163f should make the iteration over variants a tad quicker.

### DIFF
--- a/tsinfer/formats.py
+++ b/tsinfer/formats.py
@@ -226,15 +226,28 @@ def zarr_summary(array):
     return ret
 
 
-def chunk_iterator(array):
+def chunk_iterator(array, indexes=None):
     """
-    Utility to iterate over the rows in the specified array efficiently
-    by accessing one chunk at a time.
+    Utility to iterate over closely spaced rows in the specified array efficiently
+    by accessing one chunk at a time (normally used as an iterator over each row)
     """
+    if indexes is None:
+        indexes = range(array.shape[0])
+    else:
+        if len(indexes) > 0 and (
+            np.any(np.diff(indexes) <= 0)
+            or indexes[0] < 0
+            or indexes[-1] >= array.shape[0]
+        ):
+            raise ValueError("ids must be positive and in ascending order")
+
     chunk_size = array.chunks[0]
-    for j in range(array.shape[0]):
-        if j % chunk_size == 0:
-            chunk = array[j : j + chunk_size][:]
+    prev_chunk_id = -1
+    for j in indexes:
+        chunk_id = j // chunk_size
+        if chunk_id != prev_chunk_id:
+            chunk = array[chunk_id * chunk_size : (chunk_id + 1) * chunk_size][:]
+            prev_chunk_id = chunk_id
         yield chunk[j % chunk_size]
 
 
@@ -1826,15 +1839,14 @@ class SampleData(DataContainer):
             num_alleles[j] = len(alleles)
         return num_alleles[sites]
 
-    def variants(self):
+    def variants(self, sites=None):
         """
         Returns an iterator over the Variant objects. This is equivalent to
         the TreeSequence.variants iterator.
         """
-        all_genotypes = chunk_iterator(self.sites_genotypes)
-        for genotypes, site in zip(all_genotypes, self.sites()):
-            variant = Variant(site=site, alleles=site.alleles, genotypes=genotypes)
-            yield variant
+        all_genotypes = chunk_iterator(self.sites_genotypes, indexes=sites)
+        for genotypes, site in zip(all_genotypes, self.sites(ids=sites)):
+            yield Variant(site=site, alleles=site.alleles, genotypes=genotypes)
 
     def __all_haplotypes(self, sites=None):
         if sites is None:
--- a/tsinfer/inference.py
+++ b/tsinfer/inference.py
@@ -43,14 +43,6 @@ import tsinfer.constants as constants
 logger = logging.getLogger(__name__)
 
 
-def update_site_metadata(current_metadata, inference_type):
-    assert inference_type in {
-        constants.INFERENCE_FULL,
-        constants.INFERENCE_FITCH_PARSIMONY,
-    }
-    return {"inference_type": inference_type, **current_metadata}
-
-
 def is_pc_ancestor(flags):
     """
     Returns True if the path compression ancestor flag is set on the specified
@@ -129,7 +121,20 @@ def _get_progress_monitor(progress_monitor):
     return progress_monitor
 
 
-def verify(samples, tree_sequence, progress_monitor=None):
+def _encode_metadata(obj):
+    return json.dumps(obj).encode()
+
+
+def _update_site_metadata(current_metadata, inference_type):
+    assert inference_type in {
+        constants.INFERENCE_NONE,
+        constants.INFERENCE_FULL,
+        constants.INFERENCE_FITCH_PARSIMONY,
+    }
+    return {"inference_type": inference_type, **current_metadata}
+
+
+def verify(sample_data, tree_sequence, progress_monitor=None):
     """
     verify(samples, tree_sequence)
 
@@ -142,14 +147,14 @@ def verify(samples, tree_sequence, progress_monitor=None):
         instance an encoding of the specified samples that we wish to verify.
     """
     progress_monitor = _get_progress_monitor(progress_monitor)
-    if samples.num_sites != tree_sequence.num_sites:
+    if sample_data.num_sites != tree_sequence.num_sites:
         raise ValueError("numbers of sites not equal")
-    if samples.num_samples != tree_sequence.num_samples:
+    if sample_data.num_samples != tree_sequence.num_samples:
         raise ValueError("numbers of samples not equal")
-    if samples.sequence_length != tree_sequence.sequence_length:
+    if sample_data.sequence_length != tree_sequence.sequence_length:
         raise ValueError("Sequence lengths not equal")
     progress = progress_monitor.get("verify", tree_sequence.num_sites)
-    for var1, var2 in zip(samples.variants(), tree_sequence.variants()):
+    for var1, var2 in zip(sample_data.variants(), tree_sequence.variants()):
         if var1.site.position != var2.site.position:
             raise ValueError(
                 "site positions not equal: {} != {}".format(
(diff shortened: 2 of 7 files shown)


tsinfer/inference.py
```diff
@@ -166,6 +171,116 @@ def verify(samples, tree_sequence, progress_monitor=None):
     progress.close()
 
 
+def verify_samples(sample_data, ts, sample_id_map=None):
```
hyanwong, line 174: This needs tests

tsinfer/inference.py
```diff
@@ -490,6 +607,111 @@ def match_samples(
...
 
 
+def map_additional_mutations(
+    sample_data,
+    tables,
+    *,
+    sample_id_map=None,
+    sites_exist=True,
+    check_consistency=True,
+    progress_monitor=False,
```
hyanwong, line 617: This is a nasty way to pass the tsinfer progress monitor to a more general function. What's the proper way to do this, @jeromekelleher ?

hyanwong:
Current a draft PR, as there are missing tests, and the way to output the progress monitor needs a little thought (both commented below)

tsinfer/inference.py
```diff
@@ -490,6 +607,111 @@ def match_samples(
...
+    new_sd_sites = np.where(np.isin(positions, tables.sites.position) == 0)[0]
+    # Create new sites and add the mutations
+    pm = progress_monitor.get("ms_parsimony_mutations", len(new_sd_sites))
+    for variant in sample_data.variants(ids=new_sd_sites):
+        site = variant.site
+        pos = site.position
+        anc_state = site.ancestral_state
+        anc_value = 0  # sample_data files always have 0 as the ancestral allele idx
+        G = variant.genotypes[sample_indexes]
+        if np.all(np.logical_or(G == tskit.MISSING_DATA, G == anc_value)):
```
jeromekelleher, line 660: It took a while to parse this - a comment would have helped. So, if the site is just a mixture of missing data of the ancestral state, or the ancestral state, we don't bother doing any inference and impute everything to be ancestral. But, wouldn't Fitch do this anyway? This seems like unnecessary complexity to me, which is (a) unrelated to the purpose of the PR and (b) probably a premature optimisation. Is this corner case actually costing us time? Do we gain anything from identifying it? I would prefer to remove it, and get the real functionality in, and follow it up later if it's needed.

tsinfer/inference.py
```diff
@@ -490,6 +607,111 @@ def match_samples(
...
+                    site=new_site_id,
+                    node=mutation.node,
+                    derived_state=mutation.derived_state,
+                    parent=mut_map[mutation.parent],
+                )
+        pm.update()
+    pm.close()
+
+    tables.sort()
+    if check_consistency:
```
jeromekelleher, line 709: We shouldn't do this here - this is  testing functionality, not user functionality.

tsinfer/inference.py
```diff
@@ -490,6 +607,111 @@ def match_samples(
...
+    contained in the sample_data file.
+    """
+    if sample_id_map is None:
+        sample_id_map = {i: i for i in range(sample_data.num_samples)}
+    if sample_data.sequence_length != tables.sequence_length:
+        raise ValueError(
+            f"The sample_data and tables must have the same sequence length"
+        )
+    ts = tables.tree_sequence()
+    if check_consistency:
```
jeromekelleher, line 645: I think this will end up being a headache - let's not do it for now.

tsinfer/formats.py
```diff
@@ -1825,15 +1833,14 @@ def num_alleles(self, sites=None):
             num_alleles[j] = len(alleles)
         return num_alleles[sites]
 
-    def variants(self):
+    def variants(self, ids=None):
```
jeromekelleher, line 1836: The parameter should be ``sites``, I think.

tsinfer/inference.py
```diff
@@ -490,6 +607,111 @@ def match_samples(
     return ts
 
 
+def map_additional_mutations(
```
jeromekelleher, line 610: What about ``insert_missing_sites`` as a more descriptive name?

tsinfer/formats.py
```diff
@@ -226,15 +226,24 @@ def zarr_summary(array):
...
-def chunk_iterator(array):
+def chunk_iterator(array, indexes=None):
     """
-    Utility to iterate over the rows in the specified array efficiently
-    by accessing one chunk at a time.
+    Utility to iterate over closely spaced rows in the specified array efficiently
+    by accessing one chunk at a time (normally used as an iterator over each row)
     """
+    if indexes is None:
+        indexes = range(array.shape[0])
```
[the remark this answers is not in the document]
hyanwong: OK. I thought `range` would be more efficient as it doesn't create the array, as np.arange would do, but merely returns a generator (in Python3, of course)?

tsinfer/formats.py
```diff
@@ -226,15 +226,24 @@ def zarr_summary(array):
...
+    else:
+        if len(indexes) > 0 and (indexes[0] < 0 or np.any(np.diff(indexes) <= 0)):
+            raise ValueError("ids must be positive and in ascending order")
```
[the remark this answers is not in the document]
hyanwong: Since we are ensuring the indices are mononically increasing, think we can simply check the top and bottom ends, so I'll add a check to make sure that `indexes[-1] >= array.shape[0]` too.

tsinfer/inference.py
hyanwong: We can't use Fitch if all the sites are missing (we previously simply skipped the site if so). I thought that we might as well skip it if there was any mix of missing or ancestral sites. But I could simply restrict it to all-missing sites, as before.

It seemed more informative to mark these sites as not having used the fitch-inference algorithm at all (since we haven't), hence the `INFERENCE_NONE` idea. I think it would be misleading to mark an entirely missing site as INFERENCE_FITCH, to be honest.

tsinfer/inference.py
hyanwong: OK. I do think we want to allow the user to check that the samples (0..N) in the sample_data file correspond to the sample nodes (0..N) in the corresponding tree sequence, don't we? I can imagine a user calling this function on an unsimplified tree sequence and getting back a completely weird result, unless some checking is done. Similarly, are we guaranteeing that the sample nodes returned from tsinfer are in the same order as the samples in the sample_data file?

tsinfer/inference.py
jeromekelleher: Ah, OK, thanks, that makes more sense. A comment to that effect would help.

jeromekelleher: So yes, INFERENCE_NONE makes sense now.

tsinfer/formats.py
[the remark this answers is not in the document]
jeromekelleher: Yes, your're right. I thought we were using this as a numpy indexing variable, but we're just iterating over so this is good.

tsinfer/inference.py
jeromekelleher: If both have n samples, then that's a pretty good check, right?

jeromekelleher: We shouldn't be relying on the samples in the ts being 0..N though.

hyanwong: Aren't the samples sometimes at the end of the nodes array, rather than at the beginning? In general, I wasn't sure we should rely on users expecting the ids of the nodes in the sample data and the tree sequence being the same and in the same order (although we are OK when using the function internally, I suppose)

tsinfer/inference.py
hyanwong: Could do. But I see this as useful to a general user, not just for our internal purposes. For instance, Wilder has clipped out the triallelic and low-quality sites from his inferences, but we could use this function to put them back in. So the sites are not "missing" as such, it's just that we suddenly find that we might want to append some new ones to the original set.

tsinfer/inference.py
jeromekelleher: The samples are just ``ts.samples()``, and they can be anywhere in the nodes array. It's part of the API contract that the user provides a tree sequence in which ts.samples() corresponds to the sample data file. The simplest check is that the sizes are equal. Let's just keep this simple.

tsinfer/inference.py
jeromekelleher: They're missing from the tree sequence, right?

hyanwong: Yes, that's a fair point.

tsinfer/formats.py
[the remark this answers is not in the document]
hyanwong: Done

tsinfer/inference.py
```diff
@@ -490,6 +607,111 @@ def match_samples(
...
+    sample_data,
+    tables,
```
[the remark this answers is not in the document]
hyanwong: OK, done.

tsinfer/inference.py
hyanwong: Good point! Done

hyanwong:
This is now ready for review & hopefully (if it passes CI) merging, @jeromekelleher , although the tests depend on #327 being merged first.

hyanwong:
> LGTM, thanks @hyanwong! Good to merge after a rebase.

Great, thanks. Rebased now.

benjeffery:
You guys need a rebasing robot. Happy to install one!

hyanwong:
> You guys need a rebasing robot. Happy to install one!

Yes please!

jeromekelleher:
> You guys need a rebasing robot. Happy to install one!

Pretty please!

Issue #328, opened 2020-08-13
Adding ancient samples to the tsinfer pipeline


hyanwong:
After discussion with @jeromekelleher and @awohns around the topic of "historical" samples (i.e. those associated with an individual with non-zero `individuals_time`), we are closing https://github.com/tskit-dev/tsinfer/pull/300, where some relevant discussion has taken place, and will open up a new PR, with goal to create a new function which given an existing ancestor_data file and a sample data file, outputs a new ancestor_data file with added "proxy sample ancestors" based on some of the samples in the sample data file.

A "proxy sample ancestor" (other name suggestions welcome) is an ancestor that has exactly the same haplotype at inference sites as a sample, but is allowed to differ at other sites: the node (`v`) that it creates in the tree sequence will *not* be marked with `NODE_IS_SAMPLE`, and will *not* be associated with an individual. Each proxy ancestor will be added at a time `T+epsilon`, where `T` is the time of the corresponding individual. During the `match_samples` phase, the true sample will be matched and  should match exactly to the proxy ancestor at full inference sites. Given this is the case, it should then be possible to change the node time of the true sample to be exactly `T`, with the parent node `v` at time  `T+epsilon`.

We agreed that this should be the default way to include historical samples, as it provides a branch for placement of unique mutations (i.e. singletons) in the historical samples. 

### Additional notes:

Adding proxy sample ancestors at a user-specified time only makes sense when the ancestor times and the individual times use the same units. For example if the ancestors are using frequency as a measure of time, we should probably bomb out when adding historical samples, as the individuals_time is likely to be measured in years or generations. It would be nice eventually to encode the meaning of time in the top level metadata associated with ancestors and sample data files.

We need extra functionality in the `match_samples` stage. This would operate after matching a sample, and would attempt to change the node time of the sample if it is associated with a non-zero-time individual. It may be that this is impossible, in which case we should probably bomb out (and for this reason, we should probably try to match the historical samples first). We should make this an option to match samples - perhaps "force_historical_times" or something? Again, we should only allow this if the time units of the ancestors_ts are the same as the time units in the sample_data file.

The pipeline would presumably go:
1) generate `Ancestors_1` from the sample_data file with all samples, including historic ones
2) match ancestors
3) match samples *excluding* historical samples (but mapping non-full-inference sites such as singletons etc, as happens by default).
[...]

PR #331, opened 2020-08-14
Add proxy sample ancestors to allow historic samples to be added

hyanwong:
As discussed in #328. No tests added yet, but the code below seems to work OK. I'm opening as a draft PR so that @awohns can see if it makes sense, and works for his use case.

```
import msprime
import tsinfer
ts = msprime.simulate(10, mutation_rate=2)
sd = tsinfer.SampleData.from_tree_sequence(ts)
anc = tsinfer.generate_ancestors(sd)
new_anc = anc.insert_proxy_samples(sd, sample_ids=[1]) # NEW FUNC
ats = tsinfer.match_ancestors(sd, new_anc)
tsinfer.match_samples(sd, ats)
```

### DIFF
--- a/tsinfer/formats.py
+++ b/tsinfer/formats.py
@@ -349,6 +349,7 @@ class DataContainer(object):
 
     def __init__(
         self,
+        *,
         path=None,
         num_flush_threads=0,
         compressor=DEFAULT_COMPRESSOR,
@@ -777,7 +778,7 @@ class Population(object):
 
 class SampleData(DataContainer):
     """
-    SampleData(sequence_length=0, path=None, num_flush_threads=0, \
+    SampleData(sequence_length=0, *, path=None, num_flush_threads=0, \
     compressor=DEFAULT_COMPRESSOR, chunk_size=1024, max_file_size=None)
 
     Class representing input sample data used for inference.
@@ -1963,7 +1964,7 @@ class Ancestor(object):
 
 class AncestorData(DataContainer):
     """
-    AncestorData(sample_data, path=None, num_flush_threads=0, compressor=None, \
+    AncestorData(sample_data, *, path=None, num_flush_threads=0, compressor=None, \
     chunk_size=1024, max_file_size=None)
 
     Class representing the stored ancestor data produced by
--- a/tsinfer/formats.py
+++ b/tsinfer/formats.py
@@ -64,6 +64,13 @@ DEFAULT_COMPRESSOR = numcodecs.Zstd()
 DEFAULT_MAX_FILE_SIZE = 2 ** 30 if sys.platform == "win32" else 2 ** 40
 
 
+def exclude_id(attribute, value):
+    """
+    Used to filter out the id field from attrs objects such as Ancestor
+    """
+    return attribute.name != "id"
+
+
 def remove_lmdb_lockfile(lmdb_file):
     lockfile = lmdb_file + "-lock"
     if os.path.exists(lockfile):
@@ -187,7 +194,7 @@ class BufferedItemWriter(object):
         """
         Add an item to each of the arrays. The keyword arguments for this
         function correspond to the keys in the dictionary of arrays provided
-        to the contructor.
+        to the constructor.
         """
         if self.num_buffered_items[self.write_buffer] == self.chunk_size:
             self._queue_flush_buffer()
@@ -613,6 +620,18 @@ class DataContainer(object):
         timestamp = datetime.datetime.now().isoformat()
         record = provenance.get_provenance_dict(command=command, **kwargs)
         self.add_provenance(timestamp, record)
+
+    def clear_provenances(self):
+        """
+        Clear all provenances in this instance
+        """
+        if self._mode not in (self.BUILD_MODE, self.EDIT_MODE):
+            raise ValueError(
+                "Invalid operation: cannot clear provenances unless in BUILD "
+                "or EDIT mode"
+            )
+        self.provenances_timestamp.resize(0)
+        self.provenances_record.resize(0)
 
     @property
     def format_name(self):
@@ -1850,8 +1869,6 @@ class SampleData(DataContainer):
             yield Variant(site=site, alleles=site.alleles, genotypes=genotypes)
 
     def __all_haplotypes(self, sites=None):
-        if sites is None:
-            sites = np.arange(self.num_sites, dtype=int)
         # We iterate over chunks vertically here, and it's not worth complicating
         # the chunk iterator to handle this.
         chunk_size = self.sites_genotypes.chunks[1]
@@ -1859,7 +1876,7 @@ class SampleData(DataContainer):
             if j % chunk_size == 0:
                 chunk = self.sites_genotypes[:, j : j + chunk_size].T
             a = chunk[j % chunk_size]
-            yield j, a[sites]
+            yield j, a if sites is None else a[sites]
 
     def haplotypes(self, samples=None, sites=None):
         """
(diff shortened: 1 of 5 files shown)


hyanwong:
The main complication here is the definition of `epsilon`, and it would be good to get some feedback on this from @jeromekelleher. In particular:

1. It seems useful to allow each sample_id to have its own epsilon, as then in a single sweep we can do the "sequential augmenting" process we used for UK Biobank, by randomly picking 2 samples and setting their epsilon to 0.09, randomly picking another 4 and setting their epsilon to 0.08, another 8, with epsilon 0.07, and so on. It would be nice to use this to (eventually) deprecate the augment_ancestors functionality, I think, as the current approach is less intricate in my mind, and naturally saves the intermediate files in a sensible way.

2. We also want a default epsilon value that will work regardless of the time units/scale, so it seems sensible to use the existing ancestors times to determine this, which is what I have implemented (i.e. pick the smallest existing time difference and divide it by (say) 100. I'm more than open to other suggestions as to how to do this, though.

tsinfer/formats.py
```diff
@@ -2174,8 +2185,182 @@ def ancestors_length(self):
...
+            raise ValueError(
+                "The number of epsilon values must equal the number of sample_ids ({})".format(
+                    len(sample_ids)
+                )
+            )
+
+        proxy_times = sample_times + epsilons
+        reverse_time_sorted_indexes = np.argsort(proxy_times)[::-1]
+
+        ancestor_data = self.copy(path=path, max_file_size=max_file_size)
```
jeromekelleher, line 2312: This seems a bit wasteful since we're writing all the data again in a minute - is the idea to avoid writing out various bits of header information again?

hyanwong: It started because I was going to keep the sites_positions array the same, but then persisted b because it means we could keep the same compressors, chunk_sizes, etc. I assumed that a straight like-for-like memory copy would be quick, and we are assured that the new file will have zarr arrays at least as big as the existing ones, so it's not like we can avoid allocating the memory itself. I hoping the resize() operation (in `clear_data()`) wouldn't try to free the zarr memory, but just use what was there previously. But you are bound to know the best way to do this.

hyanwong:
> This looks really nice, thanks @hyanwong. 

Great, thanks a lot for looking through it.

> A few bits of input:
> 
> * If we're going to be using this at biobank scale, we'll probably need to worry about performance a bit more. There's a couple of things like accessing the `genotypes` array directly which will be fine if we're doing small numbers of samples, but very inefficient if there's lots. I don't think we should worry about this on the first pass, but it's worth keeping in mind (and maybe dropping a few comments to say where these bottlenecks might be (basically, anything which accesses single rows of the underlying zarr arrays).

What's the best way to access a single haplotype at random? I'm assuming that in the biobank case, we will only be adding a few samples (e.g. 8) from an entire sample data file (e.g. UKBB). I assume that using the haplotypes iterator will be much worse in this case: is there an alternative?

> * Likewise for calling `copy`. I would imagine it's simpler to create a new AncestorData file rather than trying fiddle with the results of copy as well.

Good point (commented separately)

> * I hadn't realised that we can do the augment ancestors thing with this process, that's very exciting and well spotted!

Yep, I was pleased when I realised this too. It has the other advantage that augment_ancestors adds 1 to all the times in the ancestors tree sequence, which means it isn't terribly useful if the times are meant to retain any meaning. 

> * I don't have any suggestions on how to do epsilon better.

OK, useful to have the input here, thanks.

hyanwong:
@jeromekelleher: what do you think we should add as provenance for the newly created ancestor data file? I guess we keep the same provenance as in the original file (which is simply copied directly from the  sample data file), but we should also add in something to state the provenance of the added proxy sample ancestors. This should be easy when they are being added from the same (original) sample data file (i.e. when `require_same_sample_data=True`) but if we are adding them from somewhere else, there is another whole trail to stick into the provenances somehow.

tsinfer/formats.py
jeromekelleher: I think it would be simplest (and easiest to follow) to create a new instance - we're doing a lot of working around and fiddling with internals otherwise. I think the right approach is to pass all the extra details for creating an AncestorData file in as extra kwargs, so we get rid of 
```     
       path=None,
￼        num_flush_threads=0,
￼        max_file_size=None,
```
and replace them with ``**kwargs``, and then use 
```
with AncestorData(**kwargs) as other:
    # build the new file
return other
```
or similar.

jeromekelleher:
> What's the best way to access a single haplotype at random? I'm assuming that in the biobank case, we will only be adding a few samples (e.g. 8) from an entire sample data file (e.g. UKBB). I assume that using the haplotypes iterator will be much worse in this case: is there an alternative?

Well in the augment ancestors case for UKB we'll be accessing more than 100K haplotypes, so this is certainly in the regime where it would be best to iterate over all the haplotypes. I think a simple approach where we iterate over all the sample haplotypes and keep the ones we're interested in in a list, and then put them into the ancestors file. True, this is quite a lot worse for the small n case, but it won't be *that* bad, and we can always optimise for the small n case later. If we use the current small n algorithm, I think it would be disastrously worse to the point where it's not actually practical.

jeromekelleher:
> @jeromekelleher: what do you think we should add as provenance for the newly created ancestor data file? I guess we keep the same provenance as in the original file (which is simply copied directly from the sample data file), but we should also add in something to state the provenance of the added proxy sample ancestors

Let's not worry about this for now. Just add an extra entry with a TODO for the parameters.

hyanwong:
> Well in the augment ancestors case for UKB we'll be accessing more than 100K haplotypes, so this is certainly in the regime where it would be best to iterate over all the haplotypes. I think a simple approach where we iterate over all the sample haplotypes and keep the ones we're interested in in a list, and then put them into the ancestors file. True, this is quite a lot worse for the small n case, but it won't be _that_ bad, and we can always optimise for the small n case later. If we use the current small n algorithm, I think it would be disastrously worse to the point where it's not actually practical for UKB augment ancestors.

OK. But I see that the `haplotypes()` iterator seems to do basically the same as I an doing: it accesses row-wise via `self.sites_genotypes[:, j : j + chunk_size]` (line 1867 of formats.py). So I think the only advantage to using `haplotypes()` is that we would then be accessing the samples sequentially and buffering them, rather than pulling out rows one-by one in time order. Perhaps this is still worth doing, however?

jeromekelleher:
> So I think the only advantage to using haplotypes() is that we would then be accessing the samples sequentially and buffering them, rather than pulling out rows one-by one in time order. Perhaps this is still worth doing, however?

Yes, definitely. Accessing the zarr arrays involves decompressing the underlying chunks (which is very costly). If you access all n rows in a chunk individually, you'll decompress the chunk n times. That's why we've gone to the trouble of setting up these chunkwise iterators and hidden the details from the API users.

tsinfer/formats.py
hyanwong: Done in latest push

hyanwong:
Looks like coverage is OK now @jeromekelleher, and this can be merged. Hurrah!

jeromekelleher (approved):
Great, thanks @hyanwong!

jeromekelleher:
@benjeffery just FYI two weird things happened with this PR, which might indicate something iffy with our CI setup here if they recur:

1. We didn't get any reports from codecov, even though the reports were being uploaded OK
2. There was no AUTOMERGE-REQUESTED label available to add, even though the mergify actions were running.

benjeffery:
Sorry - should have said, I don't have permissions to add labels to this repo.

As for codecov, we've seen that on other PRs at times, other people report it but noone's got to the bottom of it.

jeromekelleher:
> Sorry - should have said, I don't have permissions to add labels to this repo.

Ah, OK. You should be an admin now.

PR #337, opened 2020-09-03
added get_sites_bound

### LINKED ISSUE #335: sampledata.site_times_bound()
It would be helpful to have a function of the `SampleData` class which returns a numpy array of length `sampledata.num_sites` with the age of the oldest individual with a derived allele at each site.

For instance, if a SampleData file has a historic sample at time 1000, the function would return 0 at all sites where the historic sample has the ancestral allele and 1000 at sites where it has the derived allele. If there were two historic samples aged 100 and 1000 and _both_ had a derived allele at a given site, 1000 would be returned.

Was just discussing this with @hyanwong. It could be a function in tsdate (it lives in an unmerged branch right now) but it feels more like a function for tsinfer. What do you think @jeromekelleher? Could easily remain in tsdate though as an auxillary function for returning "dated" SampleData files.

awohns:
Proposed method for getting the lower bound of sites_time using individuals_time. Closes #335 

Specifically seeking input on:
- the name of the function
- The efficiency seems to be acceptable, but if there are easy wins to make it faster that would be great
- I plan to call this function in `tsdate.get_sites_time(ts, samples)` which takes a dated tree sequence of only modern samples and returns a "dated" SampleData file with ancients using dates from the tree sequence and constraints from the sampledata file. I thought this particular part of the function would be more suitable for `tsinfer` though, as it's a simple operation on a `SampleData` file.

### DIFF
--- a/tsinfer/formats.py
+++ b/tsinfer/formats.py
@@ -1315,6 +1315,35 @@ class SampleData(DataContainer):
             subset.record_provenance(command="subset", **kwargs)
         return subset
 
+    def min_site_times(self, individuals_only=False):
+        """
+        Returns a numpy array of the lower bound of the time of sites in the SampleData
+        file. Each individual with a nonzero time (from the individuals_time array)
+        gives a lower bound on the age of sites where the individual carries a
+        derived allele.
+
+        :return: A numpy array of the lower bound for each sites time.
+        :rtype: numpy.ndarray(dtype=float64)
+        """
+        samples_individual = self.samples_individual[:]
+        assert np.all(samples_individual >= 0)
+        samples_time = self.individuals_time[:][samples_individual]
+        if np.any(samples_time < 0):
+            raise ValueError("Individuals cannot have negative times")
+        historic_samples = samples_time != 0
+        historic_samples_time = samples_time[historic_samples]
+        sites_bound = np.zeros(self.num_sites)
+        for var in self.variants():
+            historic_genos = var.genotypes[historic_samples]
+            derived = historic_genos > 0
+            if np.any(derived):
+                historic_bound = np.max(historic_samples_time[derived])
+                if historic_bound > sites_bound[var.site.id]:
+                    sites_bound[var.site.id] = historic_bound
+        if not individuals_only:
+            sites_bound = np.maximum(self.sites_time[:], sites_bound)
+        return sites_bound
+
     ####################################
     # Write mode
     ####################################
--- a/tests/test_formats.py
+++ b/tests/test_formats.py
@@ -1671,6 +1671,59 @@ class TestSampleDataMerge(unittest.TestCase):
             self.assertTrue(sd2.data_equal(sd3))
 
 
+class TestMinSiteTimes(unittest.TestCase):
+    """
+    Test cases for sample data's min_site_times function
+    """
+
+    def test_no_historic(self):
+        ts = get_example_ts(10, 10, 1)
+        sd1 = formats.SampleData.from_tree_sequence(ts)
+        # No arguments and individuals_only=True should give array of zeros
+        bounds_individuals_only = sd1.min_site_times(individuals_only=True)
+        self.assertTrue(
+            np.array_equal(bounds_individuals_only, np.zeros(sd1.num_sites))
+        )
+        bounds = sd1.min_site_times()
+        self.assertTrue(np.array_equal(bounds, sd1.sites_time[:]))
+
+    def test_simple_case(self):
+        sample_times = [0, 0, 0.5, 1]
+        ts = get_example_historical_sampled_ts(sample_times, 1e1)
+        sd1 = formats.SampleData.from_tree_sequence(ts)
+        time_bound_individuals_only = sd1.min_site_times(individuals_only=True)
+        # Because this is a haploid tree sequence we can use the
+        # individual and sample IDs interchangably.
+        self.assertTrue(
+            np.all(
+                np.in1d(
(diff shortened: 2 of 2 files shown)


tsinfer/formats.py
```diff
@@ -1315,6 +1315,31 @@ def subset(self, individuals=None, sites=None, **kwargs):
...
 
+    def get_sites_time_bound(self):
+        """
+        Returns a numpy array of the lower bound of the time of sites in the SampleData
+        file. Each individual with a nonzero time (from the individuals_time array)
+        gives a lower bound on the age of sites where the individual carries a
+        derived allele.
+
+        :return: A numpy array of the lower bound for each sites time.
+        :rtype: numpy.ndarray(dtype=uint32)
```
jeromekelleher, line 1326: dtype=float64, presumably?

tsinfer/formats.py
```diff
@@ -1315,6 +1315,31 @@ def subset(self, individuals=None, sites=None, **kwargs):
...
+        samples_individual = self.samples_individual[:]
+        samples_time = self.individuals_time[:][samples_individual]
+        if np.any(samples_time < 0):
+            raise ValueError("Individuals cannot have negative times")
+        historic_samples = samples_time != 0
+        historic_samples_time = samples_time[samples_time != 0]
+        sites_bound = np.zeros(self.num_sites)
+        for var in self.variants():
+            ancient_genos = var.genotypes[historic_samples]
+            if np.any(ancient_genos == 1):
```
jeromekelleher, line 1337: A bit more general to say ``ancient_genos > 0``? We might support non-biallelic *some* day.

tsinfer/formats.py
```diff
@@ -1315,6 +1315,31 @@ def subset(self, individuals=None, sites=None, **kwargs):
...
+                historic_bound = np.max(historic_samples_time[ancient_genos == 1])
```
jeromekelleher, line 1338: No point in computing this boolean array twice - something like ``derived = ancient_genos > ``

jeromekelleher (commented):
LGTM, some minor comments.

tsinfer/formats.py
hyanwong: Also perhaps switch to calling them "historical" not "ancient"?

hyanwong:
What's the precedent for using `get_XXX` as a method name? E.g. do we only use this for getters?

tsinfer/formats.py
```diff
@@ -1315,6 +1315,32 @@ def subset(self, individuals=None, sites=None, **kwargs):
...
+        """
+        Returns a numpy array of the lower bound of the time of sites in the SampleData
+        file. Each individual with a nonzero time (from the individuals_time array)
+        gives a lower bound on the age of sites where the individual carries a
+        derived allele.
+
+        :return: A numpy array of the lower bound for each sites time.
+        :rtype: numpy.ndarray(dtype=float64)
+        """
+        samples_individual = self.samples_individual[:]
```
hyanwong, line 1328: I think sample data files have to have individuals, but I wonder if we should put an `assert np.all(samples_individual >= 0)` in there just to be safe?

tsinfer/formats.py
```diff
@@ -1315,6 +1315,32 @@ def subset(self, individuals=None, sites=None, **kwargs):
...
+        samples_time = self.individuals_time[:][samples_individual]
+        if np.any(samples_time < 0):
+            raise ValueError("Individuals cannot have negative times")
+        historic_samples = samples_time != 0
+        historic_samples_time = samples_time[samples_time != 0]
```
hyanwong, line 1333: `historic_samples_time = samples_time[historic_samples]

hyanwong:
For a name, how about `min_site_times(include_bespoke)`? "bound" implies both lower and upper, whereas `min` is shorter and more specific.

awohns:
Sounds good to me, I guess the `individuals_only` bit would not use the values in `sites_time` if it were `False`? i.e. the default would return the `np.maximum(sites_time[:], samples.min_sites_time(individuals_only=True))`, where `samples.min_sites_time(individuals_only=True))` is the behaviour now.

Also, `min_sites_time`, rather than `min_site_times` to be consistent with `sites_time[:]` right?

hyanwong (commented):
> Sounds good to me, I guess the `individuals_only` bit would not use the values in `sites_time` if it were `False`? 

Yes, exactly. I don't know if we want the default to be `True` or `False` though.

> Also, `min_sites_time`, rather than `min_site_times` to be consistent with `sites_time[:]` right?

Good point about consistency. But I didn't want to imply that it was the minimum of `sites_time`, but rather the minimum *for each site* of the times. So possibly either `min_site_time` or `min_site_times`?

tsinfer/formats.py
```diff
@@ -1315,6 +1315,32 @@ def subset(self, individuals=None, sites=None, **kwargs):
...
+        sites_bound = np.zeros(self.num_sites)
```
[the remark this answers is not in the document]
awohns: hmm, I don't immediately see when this would be useful... Also, it would require some way of noting a "special" site time that was user added.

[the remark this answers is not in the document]
hyanwong: But we do have "special" site times that are added during add sites - that's what I meant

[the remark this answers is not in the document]
awohns: Just discussed on jitsi and we're actually on same page

jeromekelleher:
Are we happy to merge here @hyanwong?

PR #338, opened 2020-09-17
Add two parameters to from_tree_sequence()

awohns:
@hyanwong and I have found that we often want to use `sampledata.individuals_time[:]` but not `sampledata.sites_time[:]` for testing purposes when inferring and dating tree sequences.

This PR removes the `use_times` parameter in `tsinfer.formats.SampleData.from_tree_sequence()` and replaces it with two parameters: `use_site_times` keeps site times from the tree sequence and `use_individual_times` keeps the times associated with individuals from the tree sequence.


jeromekelleher (approved):
LGTM - I don't we need to worry about the compatibility paths.

PR #339, opened 2020-09-18
add modify_site_times

awohns:
New function to modify `SampleData.site_times`. This is useful when reinferring tree sequences with dates from an external source. 

Not sure if I have the :attr: part right in the docstring...

### DIFF
--- a/tsinfer/formats.py
+++ b/tsinfer/formats.py
@@ -1317,10 +1317,10 @@ class SampleData(DataContainer):
 
     def min_site_times(self, individuals_only=False):
         """
-        Returns a numpy array of the lower bound of the time of sites in the SampleData
-        file. Each individual with a nonzero time (from the individuals_time array)
-        gives a lower bound on the age of sites where the individual carries a
-        derived allele.
+        Returns a numpy array of the lower bound of the time of sites in the
+        :class:`SampleData` file. Each individual with a nonzero time (from the
+        individuals_time array) gives a lower bound on the age of sites where the
+        individual carries a derived allele.
 
         :return: A numpy array of the lower bound for each sites time.
         :rtype: numpy.ndarray(dtype=float64)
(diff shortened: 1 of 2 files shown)


tsinfer/formats.py
```diff
@@ -1344,6 +1344,37 @@ def min_site_times(self, individuals_only=False):
             sites_bound = np.maximum(self.sites_time[:], sites_bound)
         return sites_bound
 
+    def modify_site_times(self, bespoke_times=None, individuals_only=False, **kwargs):
```
jeromekelleher, line 1347: I'm finding this method pretty unnecessary, to be honest. Isn't all we're doing, really,
```python
sites_time = # some value we've got somewhere; usually,
sites_time = self.min_site_times()
with sample_data.copy() as copy:
    copy.sites_time = sites_time

# Do something with copy?

**Review by jeromekelleher — COMMENTED** [review]
I'm not really seeing much point in this method actually, @awohns. Just setting the attribute on the copy really isn't that big a hassle and a pretty simple pattern to follow. Also, we have to think about the precedent we're setting - do we write a new method for every update we might want to make to other attributes?

Sorry about this, I probably sent you in the wrong direction the other day as I hadn't thought this through properly.

**Comment by awohns** [comment]
I see your point @jeromekelleher, I'd been wondering about the same thing. Happy to leave it out!

File: tsinfer/formats.py (hunk 1 above)
Comment (awohns) [reply tsinfer/formats.py:1347]: Yep, that's basically it. Happy to leave it out

**Comment by jeromekelleher** [comment]
OK, thanks, let's just leave it out thne.

**Comment by hyanwong** [comment]
I thought of this method ("contain sites_time?") as a tiny more complicated, but not much. Something like

```
sites_time = # some value we've got somewhere; usually,
sites_time = np.maximum_at(self.min_site_times(), sites_time)
with sample_data.copy() as copy:
    copy.sites_time = sites_time

# Do something with copy?
```

But I guess the same still holds - simple enough to type these lines of code in every time (or have it as a tsdate "helper" function, if we expect it to be a standard part of the dating pathway, I suppose).

hyanwong:
Actually, I wonder if it's a bit more complicated, because of having to deal with `tskit.UNKNOWN_TIME`? I'm not sure what we would do here: presumably return `UNKNOWN_TIME` at a site if all the values are `UNKNOWN_TIME`, otherwise report the maximum of the non `UNKNOWN_TIME` values?

jeromekelleher:
It's still a function for *computing* the new ``sites_time`` though, not one for *setting* it. The setting bit should be isolated for a number of reasons.

hyanwong:
> It's still a function for _computing_ the new `sites_time` though, not one for _setting_ it. The setting bit should be isolated for a number of reasons.

Yes, that's a fair point. I guess one issue here is the (slightly clunky IMO) need to make a copy of the sample_data file before changing columns like `sites_time`. But that may be unavoidable given how it is setup, use of zarr, etc.

jeromekelleher:
> Yes, that's a fair point. I guess one issue here is the (slightly clunky IMO) need to make a copy of the sample_data file before changing columns like sites_time. But that may be unavoidable given how it is setup, use of zarr, etc.

That's the reality of how things are set up at the moment, changing this would be major surgery. Things will be much more flexible when we move to xarray/sgkit in the next iteration.

hyanwong:
Can we close this?

awohns:
Sure, it's in tsdate under [`tsdate.add_sampledata_times()`](https://tsdate.readthedocs.io/en/latest/python-api.html#tsdate.add_sampledata_times)

PR #348, opened 2020-10-26
Do not simplify out populations and individuals

hyanwong:
I suddenly realised that we might be changing the IDs of populations and individuals when simplifying after inference. Although I'm not sure we guarantee that the IDs remain the same between the sample_data file and the final tree sequence, I think it's good practice to avoid this.

I guess I should also add a unit test to check that the population and individual IDs in the final ts match that in the SD file?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,8 @@
 [0.2.0] - XXXX-XX-XX
 ********************
 
+**New features**:
+
 **Breaking changes**:
 
 - The ancestors tree sequence now contains the real alleles and not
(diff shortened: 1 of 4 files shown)


jeromekelleher:
Good spot, and this is the right fix. It'd be nice to get a simple unit test in to verify we're actually doing it. We could add an empty population to the sample data file, and verify it comes through into the ts. Not sure how to verify individuals, maybe there's a similar trick?

hyanwong:
Unit tests added, although it turns out to be hard to remove samples from a SampleData file without removing the individuals that go with them - I've had to resort to low-level trickery here.

I think this can be merged.

CHANGELOG.rst
```diff
@@ -12,6 +14,13 @@
   This functionality has been replaced by the ``exclude_positions`` argument
   to the ``infer`` and ``generate_ancestors`` functions.
 
+**Bugfixes**:
+
+- Individuals and populations in the sampledata file are kept in the returned tree
+  sequence, even if they are not referenced by any sample. The individual and population
+  ids should therefore stay the same between the sample data file and the inferred tree
```
jeromekelleher, line 21: "Should" seems weak - what about "ids therefore guaranteed to the same between"

jeromekelleher (approved):
LGTM, minor nit

CHANGELOG.rst
hyanwong: Sure. Changed.

PR #349, opened 2020-10-28
Fix docs in from_tree_sequence()

awohns:
Minor change in docs for from_tree_sequence(). Remove part of sentence which says that "all samples are set at time 0" if `use_sites_time=False`. This is misleading as `individuals_time` will not be set at time 0.


jeromekelleher (approved):
Good catch, thanks!

benjeffery:
@mergifyio rebase

Issue #296, opened 2020-07-16
Change supported versions to match tskit


jeromekelleher:
Setup.py currently lists 3.6 as the newest version, but we want to make this the oldest.

hyanwong:
This was fixed in #350 and can therefore be closed

PR #350, opened 2020-11-21
Upgrade to requiring tskit 0.3

hyanwong:
And simplify some of the equality code


hyanwong:
Update so that it fixes https://github.com/tskit-dev/tsinfer/issues/296 too

tsinfer/formats.py
```diff
@@ -64,6 +64,14 @@
 DEFAULT_MAX_FILE_SIZE = 2 ** 30 if sys.platform == "win32" else 2 ** 40
 
 
+def np_obj_equal(np_obj_array1, np_obj_array2):
+    """
+    A replacement for np.array_equal to test equality of numpy arrays that
+    contain objects, as used e.g. for metadata, location, alleles, etc.
+    """
+    return all(itertools.starmap(np.array_equal, zip(np_obj_array1, np_obj_array2)))
```
jeromekelleher, line 72: What happens if we call this on two unequal sized arrays? I suspect we need a check on the shape of the two arrays first.

This isn't necessary in the code where we're using it, but on its own the function should behave sensibly.

jeromekelleher: Function probably needs a few tests, I expect.

jeromekelleher (commented):
Looks good, but the (unrelated?) change to comparison code needs tests.

Also, what about the C API, do we need to update this?

tsinfer/formats.py
hyanwong: It was unrelated, but such a minor improvement to readability that I stuck it in as a separate commit. We could append an underscore if we don't want to guarantee it working in general. I just saw a lot of code repetition and thought we should make it cleaner.

Re unrelated changes, this related back to my infuriation with GitHub. Normally I like to be able to correct things like spelling mistakes etc as I go along, without switching to a new branch. Often I have several things like that on the go, and I don't know how best to sort them all out. I guess as separate commits within a single PR like this is OK.

jeromekelleher: It's fine to add extra commits but you need to consider things from the maintainer's perspective as well as your own. In this case, the function could easily be reused *within* tsinfer and developers would expect it to do what it says it does. So, it needs tests.

Please keep changes within a PR at least related on a single theme.

hyanwong: > Please keep changes within a PR at least related on a single theme.

Yeah, I'd like to get a workflow where that's easy. I haven't found one yet though. Suggestions welcome.

hyanwong:
Updated to be only the 0.3 update change. Will stick the obj_array improvement in a separate PR

tsinfer/formats.py
benjeffery: For my workflow it usually goes like this:

Say you're working on `topic` branch and you see something you'd like to fix that `topic` requires, but is unrelated.
`git stash`
`git checkout main`
`git checkout -b hotfix`
Make the change
`git commit`
`git push`
File PR on github
`git checkout topic`
`git rebase main`
`git stash apply`

hyanwong: Yeah, but then I lose my place in the file, and have to find it again. Doing all that for a single typo means I completely lose my train of thought. So I end up simply not othering to correct stuff :(

hyanwong: Also, I'm usually switching like this several times in a session. Because I don't like working on only a single thing at once!

benjeffery: > Yeah, but then I lose my place in the file

How? You shouldn't need to close your editor.

hyanwong: If I keep the editor open, then save the file, all the previous changes I made in the same file, which I was in the middle of sorting for the first PR, will also get included in the spelling correction PR (or whatever the second correction is)

benjeffery: Ah! So your editor should have a setting for this. I have it set to auto-load if the file changes on disk - so when I `git stash` the file instantly changes in my editor.

hyanwong:
Could we merge this @jeromekelleher - it's trivial now, I think?

jeromekelleher (approved):
LGTM!

jeromekelleher:
@Mergifyio rebase

Issue #272, opened 2020-06-01
Output tree sequence should include metadata schemas


hyanwong:
With the nice new [metadata capabilities](https://github.com/tskit-dev/tskit/pull/543) in tskit, we might assume that people will want to have metadata associated with nodes in tree sequences created from real data. In this case we'll need a way to associate a schema with tables in a SampleData file, from where it will be pulled in to the generated tree sequence. @awohns is encountering this as he wants to associate metadata with ancient samples during inference. It's not at all urgent, however, as he can just use the "old way" of JSON encoded metadata for the moment. Pinging @benjeffery as he knows most about the metadata stuff, although he might not want to dive back into it now that it's nicely signed off in tskit!

benjeffery:
How I see this working is that `tskit.MetadataSchema` would be used to encode metadata when it is stored in the SampleData file, the encoded bytes and schema would just be copied across to the tree sequence.

hyanwong:
This sounds right, and actually, not that much work, when we get round to it eventually?

jeromekelleher:
That's the plan - we need tskit 0.3.0 released before starting on it here though.

jeromekelleher:
Following the basic principles laid out in https://github.com/tskit-dev/tskit/issues/1051 we should define minimal metadata schemas for the entities that we support metadata for. I will update with a proposal PR ASAP.

hyanwong:
Do we need a JSON schema that supports `b''` as a metadata item, rather than requiring `null` or `{}`?

benjeffery:
No such schema currently exists - it could be made to if JSONCodec interpreted b'' as None

PR #353, opened 2020-11-22
Switch to null for empty metadata

### LINKED ISSUE #352: Switch to default metadata being `null` rather than `{}`?
In tskit, if the metadata schema is JSON, then blank entries are filled with `null` rather than `{}`. In tsinfer (where the JSON format is assumed, and schemas have not yet been implemented), we set blank entries to be `{}` instead. It might help transition to the schema style if we switch to `null` to match tskit. I don't think this will cause any backwards incompatibilities, as `{}` will continue to be valid in old files. It seems like a trivial change to me - can you see any problems with this @jeromekelleher ?

hyanwong:
Fixes #352

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,11 +6,21 @@
 
 - The ancestors tree sequence now contains the real alleles and not
   0/1 values as before.
+
 - Times for undated sites now use frequencies (0..1), not as counts (1..num_samples),
   and are now stored as -inf, then calculated on the fly in the variants() iterator.
+
 - The SampleData file no longer accepts the ``inference`` argument to add_site.
   This functionality has been replaced by the ``exclude_positions`` argument
   to the ``infer`` and ``generate_ancestors`` functions.
+
+**New features**:
+
+- The default metadata value if no metadata is specified for sites, samples, populations,
+  and individuals is now "null" (which translates to Python ``None``) rather than the
+  empty dict ``{}``. This matches the standard JSON schema in tskit. Old files should
+  continue to be valid.
+
 
 ********************
 [0.1.5] - 2019-09-25
--- a/tests/test_formats.py
+++ b/tests/test_formats.py
@@ -154,7 +154,7 @@ class TestSampleData(unittest.TestCase, DataContainerMixin):
     def verify_data_round_trip(self, ts, input_file):
         self.assertGreater(ts.num_sites, 1)
         for pop in ts.populations():
-            input_file.add_population(metadata=json.loads(pop.metadata or "{}"))
+            input_file.add_population(metadata=json.loads(pop.metadata or "null"))
         # For testing, depend on the sample nodes being sorted by individual
         for i, group in itertools.groupby(
             ts.samples(), lambda n: ts.node(n).individual
@@ -166,16 +166,16 @@ class TestSampleData(unittest.TestCase, DataContainerMixin):
                         ploidy=1,
                         population=node.population,
                         time=node.time,
-                        samples_metadata=[json.loads(node.metadata or "{}")],
+                        samples_metadata=[json.loads(node.metadata or "null")],
                     )
             else:
                 input_file.add_individual(
                     ploidy=len(nodes),
                     population=nodes[0].population,
-                    metadata=json.loads(ts.individual(i).metadata or "{}"),
+                    metadata=json.loads(ts.individual(i).metadata or "null"),
                     location=ts.individual(i).location,
                     time=nodes[0].time,
-                    samples_metadata=[json.loads(n.metadata or "{}") for n in nodes],
+                    samples_metadata=[json.loads(n.metadata or "null") for n in nodes],
                 )
         for v in ts.variants():
             t = np.nan  # default is that a site has no meaningful time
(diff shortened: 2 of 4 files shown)


hyanwong:
Weirdly this is complaining about a line that I'm sure is covered, e.g. in

```
nosetests tests/test_inference.py:TestHistoricalSamples
```

- If I raise an error in that part of the loop (`if ind.metadata is None:`) then it triggers in this test. So I have no idea why CI is failing here. If we agree with #352  I think we should just merge this PR, even with CI complaint. it's clearly the correct logic.

CHANGELOG.rst
```diff
@@ -6,12 +6,22 @@
...
   0/1 values as before.
+
 - Times for undated sites now use frequencies (0..1), not as counts (1..num_samples),
   and are now stored as -inf, then calculated on the fly in the variants() iterator.
+
 - The SampleData file no longer accepts the ``inference`` argument to add_site.
   This functionality has been replaced by the ``exclude_positions`` argument
   to the ``infer`` and ``generate_ancestors`` functions.
 
+**New features**:
```
jeromekelleher, line 17: That's a breaking change, surely.

tests/test_formats.py
```diff
@@ -478,10 +479,11 @@ def test_from_tree_sequence_omitting_metadata(self):
         self.assertTrue(np.all(sd2.sites_position[:] == sd1.sites_position[:]))
         self.assertTrue(np.all(sd2.sites_genotypes[:] == sd1.sites_genotypes[:]))
         self.assertTrue(np.all(sd2.sites_time[:] == sd1.sites_time[:]))
-        self.assertTrue(np.all(sd2.populations_metadata[:] == {}))
-        self.assertTrue(np.all(sd2.individuals_metadata[:] == {}))
-        self.assertTrue(np.all(sd2.samples_metadata[:] == {}))
-        self.assertTrue(np.all(sd2.sites_metadata[:] == {}))
+        print(sd2.populations_metadata[:])
```
jeromekelleher, line 482: stray print

jeromekelleher (commented):
I'm not sure what we're actually solving with this, and that this isn't a change we're going to end up undoing when we do the metadata schemas properly. 

I think I'd rather keep things as they are, and instead define a schema which permits the empty dict as metadata. That will break less code, IMO

CHANGELOG.rst
hyanwong: I don't think it breaks anything, to be honest. We can still read-in sample data files that have `{}` in the metadata file. The reason that I would like to do this is that I want to round-trip data (specifically metadata) from a tree sequence through `.from_tree_sequence()`, via inference, and back to another tree sequence, and this eases that transition, I think.

jeromekelleher: It can break downstream code that depends on the ts having a dictionary in the metadata, not None.

hyanwong: Fair enough - I can add it to breaking changes. But the JSON codec for tree sets allows this (and it's the default), so I think any such code may be problematic anyway (and will stop working when we switch to defaulting to the JSON codec for inferred tree sequences, which I assume we want to do?

jeromekelleher: I just think there's no point in doing this and making the change before we've added the schemas - we're adding extra breakage that we'll have to keep track of and document.

So, add the schemas first. We'll have to do this within a week anyway, so there's no point in putting it off.

hyanwong:
> I think I'd rather keep things as they are, and instead define a schema which permits the empty dict as metadata. That will break less code, IMO

An empty dict *is* permitted as metadata in the tskit JSON schema. It's just that's not the *default*. So we can't round-trip a JSON tree sequence via tsinfer, which is a pain, IMO - you end up changing all the metadata from `null` to `{}`.

You could argue with @benjeffery that the default "empty" metadata field for JSON in tskit should be `{}`, but I assume there's a good reason for choosing `'null'`?

jeromekelleher:
I don't want to get into the details of what the schema should and shouldn't be for tsinfer here. Let's implement the schema first before making backwards incompatible changes.

CHANGELOG.rst
hyanwong: Well, if we are doing it this week, then sure!

benjeffery:
> You could argue with @benjeffery that the default "empty" metadata field for JSON in tskit should be `{}`, but I assume there's a good reason for choosing `'null'`?

There is no default metadata for JSON - the default metadata arg to `add_row` has always been `None`.

hyanwong:
> > You could argue with @benjeffery that the default "empty" metadata field for JSON in tskit should be `{}`, but I assume there's a good reason for choosing `'null'`?
> 
> There is no default metadata for JSON - the default metadata arg to `add_row` has always been `None`.

I guess the problem is that when we get a metadata value of `None` passed to tsinfer we actually store `{}`, which is not what tskit does.

jeromekelleher:
Let's just drop this for now @hyanwong - I'll pick up the metadata in tsinfer question once the parsimony stuff is fixed.

hyanwong:
Sure. It turned out to only be a thing for some of the unit tests in https://github.com/tskit-dev/tsinfer/pull/358, which I've just changed with a comment to fix

hyanwong:
Closing in favour of a proper approach to metadata (https://github.com/tskit-dev/tsinfer/issues/272)

PR #356, opened 2020-11-23
Switch to tskit.UNKNOWN_TIME, not the bespoke tsinfer.TIME_UNSPECIFIED

### LINKED ISSUE #336: Switch to tskit.UNKNOWN_TIME?
We are currently using `tsinfer.TIME_UNSPECIFIED = -np.inf`. Should we switch to the newly instituted `tskit.UNKNOWN_TIME` which is a specific sort of NaN value, instead?

hyanwong:
Fixes #336

### DIFF
--- a/tests/test_formats.py
+++ b/tests/test_formats.py
@@ -497,9 +497,7 @@ class TestSampleData(unittest.TestCase, DataContainerMixin):
         sd1 = formats.SampleData.from_tree_sequence(ts, use_individuals_time=False)
         self.assertTrue(np.all(sd1.individuals_time[:] == 0))
         sd2 = formats.SampleData.from_tree_sequence(ts, use_sites_time=False)
-        self.assertTrue(
-            np.array_equal(sd2.sites_time[:], np.full(ts.num_sites, -np.inf))
-        )
+        self.assertTrue(np.all(tskit.is_unknown_time(sd2.sites_time[:])))
         self.assertTrue(np.array_equal(sd2.individuals_time[:], sample_times))
 
     def test_chunk_size(self):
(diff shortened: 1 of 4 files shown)


CHANGELOG.rst
```diff
@@ -4,6 +4,9 @@
 
 **Breaking changes**:
 
+- The ``TIME_UNSPECIFIED`` value, which is used to indicate that frequency should be
```
jeromekelleher, line 7: Did we actually release a version of tsinfer with this constant in the public API? If so, we need to be a bit more careful. If not, then there's no point in putting it in the changelog.

jeromekelleher (commented):
The change is fine, but we need to be clear on whether the original value was released or not (I suspect not?)

CHANGELOG.rst
hyanwong: Yeah, good point. It wasn't in 0.1.4, which it the last tagged release, so I guess I can just delete that from the changelog?

jeromekelleher: Yeah, that makes things a lot simpler. We can delete this from the changelog.

But, we should probably bump the file version number so that files we have lying around at the moment don't get misinterpreted.

hyanwong: Yep. I was going to bump the number once we had added flags into the individuals, and (possibly) switched to `null` in the metadata fields.

hyanwong: Removed from changelog. Version bump later today, I guess?

hyanwong:
Codecov is only complaining here because of the reduction in lines, I think. So This can be merged

jeromekelleher (approved):
LGTM

jeromekelleher:
Merging manually as mergify seems stuck

Issue #351, opened 2020-11-21
Update to use pytests


hyanwong:
After discussion with @benjeffery - it would be good to update tsinfer to the same as tskit. Also, Perhaps we should switch to Github Actions for testing, as we did for tskit

PR #358, opened 2020-11-23
Add Individual flag

hyanwong:
This allows us to match up with tskit individuals. ~~But the (IMO) more extensive tests currently depend on the default metadata  being `null` (#353 ), so I guess I'll need to figure out how to rework that.~~ (sorry - was being dense, this should work now)

The major change in testing is that I've put the functions to generate complex tree sequences with metadata etc into `tests/tsutil.py`, so that they can be used in both `test_formats.py` and (for round-tripping) in `test_inference.py`. Hope that's OK.

I've also started using the `from_tree_sequence()` function to check that all the necessary data (flags, metadata, etc etc) from a ts is being round-tripped via tsinfer. This seems sensible to me, and we weren't checking it before.

### DIFF
--- a/tests/tsutil.py
+++ b/tests/tsutil.py
@@ -1,0 +1,150 @@
+#
+# Copyright (C) 2020 University of Oxford
+#
+# This file is part of tsinfer.
+#
+# tsinfer is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# tsinfer is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with tsinfer.  If not, see <http://www.gnu.org/licenses/>.
+#
+"""
+Extra utility functions used in several test files
+"""
+import json
+
+import numpy as np
+import msprime
+import tskit
+
+
+def json_metadata_is_subset(metadata1, metadata2):
+    metadata1_dict = json.loads(metadata1)
+    if metadata1_dict is None:
+        metadata1_dict = {}
+    metadata2_dict = json.loads(metadata2)
+    if metadata2_dict is None:
+        metadata2_dict = {}
+    return metadata1_dict.items() <= metadata2_dict.items()
+
+
+def get_example_ts(sample_size, sequence_length, mutation_rate=10, random_seed=100):
+    return msprime.simulate(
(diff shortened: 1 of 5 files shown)


hyanwong:
Round-trip tests failing, pending a solution to the metadata != `null`  issue. May be worth waiting until we actually sort metadata properly, if we can do that this week (sounds a bit tight, though!)

hyanwong:
Once we [switch to pytest](https://github.com/tskit-dev/tsinfer/issues/351), @benjeffery [points out](https://github.com/tskit-dev/tskit/pull/931#issuecomment-720597295) a nicer way to include a suite of fixed example tree sequences for testing using `conftest.py` with scope="session". I guess we should use that for the functions `get_example_ts`, `get_example_individuals_ts_with_metadata`, and `get_example_historical_sampled_ts` which I have put in tsutil in the current PR.

tsinfer/formats.py
```diff
@@ -1540,6 +1571,7 @@ def add_individual(
         :param list samples_metadata: A list of JSON encodable dict-like objects,
             of length ``ploidy``, giving metadata to be associated with each
             sample, or None to give all samples an empty metadata field.
+        :param list flags: The bitwise flags for this individual.
```
jeromekelleher, line 1574: type is int

tests/tsutil.py
```diff
@@ -0,0 +1,153 @@
+#
+# Copyright (C) 2018 University of Oxford
```
jeromekelleher, line 2: 2020?

tests/tsutil.py
```diff
@@ -0,0 +1,153 @@
...
+    nodes_individual = tables.nodes.individual  # Assign individuals to sample nodes
+    sample_individuals = np.repeat(
+        np.arange(n, dtype=tables.nodes.individual.dtype), ploidy
+    )
+    if skip_last:
+        # Should work even if some sample nodes are not assigned to an individual
+        sample_individuals[sample_individuals == n - 1] = tskit.NULL
+    nodes_individual[ts.samples()] = sample_individuals
+    tables.nodes.individual = nodes_individual
+    print(tables.populations)
```
jeromekelleher, line 122: print

jeromekelleher (commented):
LGTM, some minor comments.

tests/tsutil.py
hyanwong: Good catch

tests/tsutil.py
hyanwong: Damn - I need something to take these stray debugging things out. I always seem to miss one or two. Sorry!

jeromekelleher: happens us all

hyanwong:
Done, thanks @jeromekelleher !

jeromekelleher (approved):

jeromekelleher:
conflict here @hyanwong

PR #359, opened 2020-11-23
Use a general function for checking np object array equality

hyanwong:
Simple test to be added in a moment. But I wanted to ask here: lots of these tests for equality go:

```
return (A and B and C and D)
```

where A, B, C, D are quite complicated. When these fail, it becomes really difficult to work out which of A, B, C, or D caused the failure. So I was wondering if we should switch to something like:

```
if not A:
    return False
if not B:
    return False
if not C:
    return False
if not D:
    return False
return True
```

What do you think @benjeffery ?

### DIFF
--- a/tests/test_formats.py
+++ b/tests/test_formats.py
@@ -45,6 +45,32 @@ import tsutil
 IS_WINDOWS = sys.platform == "win32"
 
 
+class ConvenienceFunctions(unittest.TestCase):
+    """
+    Tests for a couple of convenience functions at the top level of formats.py
+    """
+
+    def test_np_obj_equal_bad(self):
+        obj_array = np.array([None])
+        self.assertRaises(AttributeError, formats.np_obj_equal, obj_array, None)
+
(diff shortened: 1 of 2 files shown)


jeromekelleher:
The function here is a useful addition, so I'd say lets merge that first and think about other things later.

Just needs a handful of example tests, nothing elaborate.

jeromekelleher:
Can you add a couple of unit tests here please @hyanwong? Really nothing elaborate, just checking the different size handling and an object array that should compare True and False.

hyanwong:
Yep, sorry - just getting onto this! By the way, could you see if https://github.com/tskit-dev/tsinfer/pull/363 makes sense, then I can move forward with that too.

hyanwong:
> Can you add a couple of unit tests here please @hyanwong? Really nothing elaborate, just checking the different size handling and an object array that should compare True and False.

I guess the most recent push should do it?

jeromekelleher (approved):

PR #362, opened 2020-11-25
Use autodoc_mock_imports

hyanwong:
I think this is how you are meant to do this now (and strangely, is the only way I could get doc compilation for tsdate working). Might it be an idea to update other tskit-dev repos, @benjeffery ?

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -17,16 +17,8 @@ import os
 import sys
 
 import pkg_resources
-from unittest.mock import MagicMock
 
-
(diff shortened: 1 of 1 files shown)


jeromekelleher:
The plan going forward is to not use this as we migrate away from rtd, as we can compile the c modules in GitHub actions. This is a good thing to do here now, though.

jeromekelleher (approved):

jeromekelleher:
@mergifyIO rebase

PR #363, opened 2020-11-25
Exclude sites if nan even after checking for unknown times

hyanwong:
This is for discussion, but should help with a pipeline for tsdate, so it would be nice to get it in for 0.2.0. I've talked it through with @awohns , who may have some comments too.

The PR follows on from the convention that we now mark sites where freq is a proxy for time with `site_time=tskit.UNKNOWN_TIME` (which is a special form of NaN). This PR deals with the case where the `site_time` is *not* `tskit.UNKNOWN_TIME`, but some *other* form of NaN. For these sites I take it that the very meaning of a site time is nonsensical (e.g. a site where there is no variation at all). If we see such sites, it seems reasonable to me to mark them as not-for-inference, which is all that this PR does.

This gives an easy way to mark sites, during the dating process, that we know are nonsensical. It goes like this. After the first inference, we examine the tree sequence, and find the sites with huge numbers of mutations. We assume the pattern at these sites is nonsense and should be ignored. We take a copy of the tree sequence with the mutations at those sites removed, and date it using tsdate. We then extract the site times from the dated tree sequence (using the `sites_time_from_ts` tsdate function). Quite reasonably, this function marks these sites with a time of `np.nan`, because there is no variation there (it's not like an UNKNOWN_TIME, it's literally a meaningless time). We stick these times back into the *original* tree sequence, and because the nonsense sites are NaNs, they are not used for inference, but are simply stuck in using parsimony, which is what we want.

It sounds a bit convoluted, but I think the logic is solid. Basically, if there's a NaN in the time then we skip inference unless it's the very special `tskit.UNKNOWN_TIME` NaN, in which case we use time=freq.

### DIFF
--- a/tsinfer/inference.py
+++ b/tsinfer/inference.py
@@ -690,6 +690,17 @@ class AncestorsGenerator(object):
         Add all sites that are suitable for inference into the
         ancestor builder (and subsequent inference), unless they
         are held in the specified list of excluded site positions.
+        Suitable sites have only 2 listed alleles, with at least two
+        samples carrying the derived allele and at least one sample
+        carrying the ancestral allele.
+
+        Suitable sites will be added at the time given by site.time, unless
+        site.time is  ``np.nan`` or ``tskit.UNKNOWN_TIME``. In the first case,
+        the site will simply excluded as if it were in the list of
+        ``excluded_positions``. In the second case, then the time associated with
+        the site will be the frequency of the derived allele (i.e. the number
+        of samples with the derived allele divided by the total number of samples
+        with non-missing alleles).
         """
         if exclude_positions is None:
             exclude_positions = set()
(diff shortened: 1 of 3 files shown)


awohns:
Looks good to me Yan.

At the most basic level, here's what I think the PR does: if you want to use `sites_time` to order mutations in tsinfer, but for whatever reason you _can't_ assign a time to some sites, they shouldn't be used for inference.

tsinfer/inference.py
```diff
@@ -714,6 +715,9 @@ def add_sites(self, exclude_positions=None):
                     # Time = freq of *all* derived alleles. Note that if n_alleles > 2 this
                     # may not be sensible: https://github.com/tskit-dev/tsinfer/issues/228
                     time = counts.derived / counts.known
+                if np.isnan(time):
```
jeromekelleher, line 718: Should be ``elif`` here - need more tests!

jeromekelleher (changes requested):
Yep, seems like a reasonable thing to do. Looks to me like the current implementation will also make ``is_unknown_time`` sites non-inference, though.

tsinfer/inference.py
hyanwong: I'm not sure it should be, should it. If it's `tskit.UNKNOWN_TIME` then it get replaced with `time=freq` on the line above, so it shouldn't be excluded from inference on this line. The only reason I didn't use an elif is that this catches the case when `counts.known` is 0, which seems worth doing (even though it's actually excluded by the logic above, which I guess could therefore be simplified). So I think the tests *do* cover this?

hyanwong:
> Yep, seems like a reasonable thing to do. 

Great, thanks for looking @jeromekelleher 

> Looks to me like the current implementation will also make `is_unknown_time` sites non-inference, though.

I don't think so (see inline comment above). Happy to be convinced otherwise though.

tsinfer/inference.py
jeromekelleher: But ``use_site`` will be False in all cases where the time is a NaN by the time we get to line 720, won't it?

jeromekelleher: Ah, right, I see, you've set the time within the condition block. Sorry, my bad.

jeromekelleher (approved):
Yep, looks good. We should document the semantics around time somewhere though I guess - the docstring for ``add_site``?

hyanwong:
Added details to the doctoring and removed an unneeded assert (because we were already checking that `counts.known > 1` at the top of the loop).

PR #364, opened 2020-11-26
Stop inference if sample_data mixes user times and freq-times

### LINKED ISSUE #330: Treatment of SampleData files with only some user-specified site times
Is it sensible to allow users to set a bespoke time for a site (`SampleData.add_site(time=XXX)`) if we are using the allele frequency as a proxy for time? Or, if there are any bespoke times specified, should we assume that the other sites (where time is unspecified) should not be used for inference? In other words, should we only use time=freq if *all*  sites_time values are `tsinfer.TIME_UNSPECIFIED`, and otherwise just mark sites with unspecified times as not-for-inference?

hyanwong:
Fixes #330

### DIFF
--- a/tests/test_inference.py
+++ b/tests/test_inference.py
@@ -1331,6 +1331,22 @@ class TestBuildAncestors(unittest.TestCase):
             with self.assertRaises(error):
                 generator.ancestor_builder.make_ancestor([0], h)
 
+    def test_mixed_freq_and_user_times(self):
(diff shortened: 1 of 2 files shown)


hyanwong:
I thought it best to check this when actually building the ancestors. After all there's no reason why we can't build a sample data file with a mix (and maybe change the times afterwards). It's only when actually building ancestors from the file that we want to catch the problem.

jeromekelleher (approved):
LGTM

PR #369, opened 2020-11-27
Revamp progress bars, allowing use outside of CLI

### LINKED ISSUE #259: Add ability to show progress monitor when running from within python
It's a trivial thing, but I usually run inferences (e.g. testing 1000G) using a python script, rather than the command-line `tsinfer` version, even on long inferences. It would be lovely to be able to pass `progress_monitor=True` to e.g. `tsinfer.infer()`, instead of requiring the progress_monitor function to take `None` or a `ProgressMonitor` object. Perhaps in `infer()` we could do something like:

```
if progress_monitor == True:
    progress_monitor = tsinfer.cli.ProgressMonitor(
        generate_ancestors=True,
        match_ancestors=True,
        augment_ancestors=False,
        match_samples=True)
```

And similarly, with the appropriate parameters in `generate_ancestors` and friends. What do you think @jeromekelleher ? I'm happy to code this up, although I wonder if the ProgressMonitor class would therefore be best moved to its own source file (`progress.py`?)

hyanwong:
Fixes #259

### DIFF
--- a/tsinfer/inference.py
+++ b/tsinfer/inference.py
@@ -39,6 +39,7 @@ import tsinfer.algorithm as algorithm
 import tsinfer.threads as threads
 import tsinfer.provenance as provenance
 import tsinfer.constants as constants
+import tsinfer.progress as progress
 
 logger = logging.getLogger(__name__)
 
@@ -91,34 +92,15 @@ def allele_counts(genotypes):
     )
 
 
-class DummyProgress(object):
+def _get_progress_monitor(progress_monitor, **kwargs):
     """
-    Class that mimics the subset of the tqdm API that we use in this module.
+    Check if this really is a ProgressMonitor, if not, return something usable as one
     """
-
-    def update(self):
-        pass
-
-    def close(self):
-        pass
-
-
-class DummyProgressMonitor(object):
-    """
-    Simple class to mimic the interface of the real progress monitor.
-    """
-
-    def get(self, key, total):
-        return DummyProgress()
-
-    def set_detail(self, info):
-        pass
-
-
-def _get_progress_monitor(progress_monitor):
-    if progress_monitor is None:
-        progress_monitor = DummyProgressMonitor()
-    return progress_monitor
+    if isinstance(progress_monitor, progress.ProgressMonitor):
+        return progress_monitor
+    if progress_monitor:
+        return progress.ProgressMonitor(**kwargs)
+    return progress.DummyProgressMonitor()
 
 
 def _encode_metadata(obj):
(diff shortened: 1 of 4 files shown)


jeromekelleher:
I'm not sure this is such a good idea tbh, and I'm sure we don't want to tie ourselves down to displaying progress monitors as part of the Python API. Longer term the plan is to use Dask for our parallelism, and we'll hook into its methods for reporting progress. The plan is for the next major version of tsinfer to use Dask. Apologies, I haven't actually explained this before (I'll open a roadmap issue now.)

So yes, I'd rather not make this change and instead keep the progress as a CLI only feature.

hyanwong:
Sure, if you have long term plans - happy to drop this. It's pretty helpful for the moment to be able to display progress bars in Jupiter notebooks, though. Any suggestions for how to do this?

jeromekelleher:
See #373. I guess we can keep this once we don't document the API, and are happy that the notebooks will break after 0.3.0?

hyanwong:
Yes, I guess we could merge this and keep it undocumented, if you're OK with that? It seems pretty handy to spin the progress bars off to a separate file anyway, so we can replace or delete them easily, once we move to Dask? I can't see much harm in merging this, TBH.

hyanwong:
Oh, and I guess that even in 0.3.0 we might want to allow `infer(progress_monitor=True)` in a notebook, which would use whatever reporting mechanism we decide upon for progress, rather than tqdm. So perhaps notebooks wouldn't break?

jeromekelleher:
> Oh, and I guess that even in 0.3.0 we might want to allow infer(progress_monitor=True) in a notebook, which would use whatever reporting mechanism we decide upon for progress, rather than tqdm. So perhaps notebooks wouldn't break?

They are different mechanisms in [dask](https://docs.dask.org/en/latest/diagnostics-distributed.html#progress-bar) for doing this, and I don't want to get tied down to our current way of doing things. Basically we should be able to take care of progress with no code at all on our part, if we use Dask properly.

So yes, if you want to merge this then let's remove the documentation.

hyanwong:
> They are different mechanisms in [dask](https://docs.dask.org/en/latest/diagnostics-distributed.html#progress-bar) for doing this, and I don't want to get tied down to our current way of doing things.

Yes, that would be great.

> So yes, if you want to merge this then let's remove the documentation.

It's not documented anyway. So can probably be merged?

tsinfer/inference.py
```diff
@@ -227,10 +209,17 @@ def infer(
         main inference process using parsimony. The list does not need to be
         in to be in any particular order, and can include site positions that
         are not present in the sample data file.
+    :param bool progress_monitor: Should we display a progress monitor?
```
jeromekelleher, line 212: We're documenting it here, aren't we?

hyanwong: Oh. Sorry. Didn't spot that. My bad.

hyanwong: Removed

jeromekelleher (approved):
LGTM

PR #370, opened 2020-11-27
Explicit error message if CLI inference attempted on a ts

### LINKED ISSUE #342: Sensible warning should be given if trying to run infer on a tree seq file
In #341 , @stsmall tried to run inference on a tree sequence file, not a sample data file.

```
tsinfer infer simulation-source.trees
```

And got `MDB_INVALID: File is not an LMDB file` - perhaps we should make this a more obvious error message?

hyanwong:
Fixes #342


jeromekelleher (approved):
LGTM, thanks.

PR #392, opened 2020-12-02
Metadata schemas.

jeromekelleher:
It looks like the most graceful path to enabling tsinfer users access to metadata schemas is to allow then to store the schemas for the different entities in the SampleData file. If a schema is present, it is set in the end-result tree sequence. If not, code continues to work as before. This lets us keep compatibility with old code, which given a path to being able to actually use the new metadata goodness. Particularly in light of the sharp compatiblity break we're looking at for 0.3, I think it's good to keep things working as before by default for the 0.2 series.

This is the initial sketch of how it might be done - there's a bunch of details to be worked out and so on. I think the responsibility for creating user-friendly tools for *describing* a schema should be pushed up to tskit, though, so I'm pretty happy with asking the user for tskit MetadataSchema instance. The ``permissive_json()`` function is just a temporary thing to help me write test cases.

### DIFF
--- a/tsinfer/formats.py
+++ b/tsinfer/formats.py
@@ -62,6 +62,18 @@ DEFAULT_COMPRESSOR = numcodecs.Zstd()
 DEFAULT_MAX_FILE_SIZE = 2 ** 30 if sys.platform == "win32" else 2 ** 40
 
 
+def permissive_json_schema():
+    # A version of this should be in tskit, but we keep it here for now
+    # for convenience
+    return {
+        "codec": "json",
+        "type": "object",
+        "properties": {},
+        "required": [],
+        "additionalProperties": True,
+    }
+
+
 def np_obj_equal(np_obj_array1, np_obj_array2):
     """
     A replacement for np.array_equal to test equality of numpy arrays that
@@ -804,7 +816,6 @@ class Sample:
     # TODO document properly.
     id = attr.ib()
     individual = attr.ib()
-    metadata = attr.ib()
 
 
 @attr.s
@@ -927,6 +938,8 @@ class SampleData(DataContainer):
 
         super().__init__(**kwargs)
         self.data.attrs["sequence_length"] = float(sequence_length)
+        self.data.attrs["metadata"] = {}
+        self.data.attrs["metadata_schema"] = permissive_json_schema()
         chunks = (self._chunk_size,)
         populations_group = self.data.create_group("populations")
         metadata = populations_group.create_dataset(
@@ -937,11 +950,13 @@ class SampleData(DataContainer):
             dtype=object,
             object_codec=self._metadata_codec,
         )
+        populations_group.attrs["metadata_schema"] = None
         self._populations_writer = BufferedItemWriter(
             {"metadata": metadata}, num_threads=self._num_flush_threads
         )
 
         individuals_group = self.data.create_group("individuals")
+        individuals_group.attrs["metadata_schema"] = None
         metadata = individuals_group.create_dataset(
             "metadata",
             shape=(0,),
@@ -997,20 +1012,13 @@ class SampleData(DataContainer):
             compressor=self._compressor,
             dtype=np.int32,
         )
-        metadata = samples_group.create_dataset(
-            "metadata",
-            shape=(0,),
-            chunks=chunks,
-            compressor=self._compressor,
-            dtype=object,
-            object_codec=self._metadata_codec,
-        )
         self._samples_writer = BufferedItemWriter(
-            {"individual": individual, "metadata": metadata},
+            {"individual": individual},
             num_threads=self._num_flush_threads,
         )
 
         sites_group = self.data.create_group("sites")
+        sites_group.attrs["metadata_schema"] = None
         sites_group.create_dataset(
             "position",
             shape=(0,),
@@ -1051,7 +1059,7 @@ class SampleData(DataContainer):
 
         self._last_position = 0
         self._sites_writer = None
-        # We are initially in the ADDING_POPULATIONS.
+        # We are initially in the ADDING_POPULATIONS state.
         self._build_state = self.ADDING_POPULATIONS
 
     def summary(self):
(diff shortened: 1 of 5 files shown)


jeromekelleher:
I think this is ready for review @benjeffery @hyanwong. I ended up doing a fair bit of messing around to support both the old-form and new-form metadata, so hopefully we won't break too much code. I'll make a follow-up to bump the file-format version and see what else needs doing compat-wise.

tsinfer/formats.py
```diff
@@ -1378,6 +1435,10 @@ def from_tree_sequence(
         with ``u`` sample nodes but no individuals will be translated into a SampleData
         file with ``u`` haploid individuals and ``u`` samples.
 
+        If metadata is present in the tree sequence it must either have a schema
```
hyanwong, line 1438: Should we say that metadata is transferred into the sample data file? It always is now, right (apart from node metadata)? Would it be worth saying that it's only top level, individuals, populations, and sites metadata that are taken from the TS? Presumably all others (e.g. node metadata) are ignored, and so could be any old schema.

So something like:

"Metadata associated with individuals, populations, sites, and at the top level of the tree sequence, is also stored in the appropriate places in the returned SampleData instance. Any such metadata must either have a schema defined[insert link to metadata schema discussion in tskit docs] or be JSON encodable text. Attempts to use a tree sequence where any such metadata is not JSON and has no schema will fail.

tsinfer/inference.py
```diff
@@ -44,6 +45,32 @@
...
+sample_data_time_metadata_definition = {
+    "description": "Time of an individual from the SampleData file.",
+    "type": "number",
+    # Defaults aren't currently used, see
+    # https://github.com/tskit-dev/tskit/issues/1073
+    "default": 0,
+}
+
+inference_type_metadata_definition = {
+    "description": "The type of inference used at this site.",
```
hyanwong, line 57: Should we give a reference to the types we allow here?

hyanwong (approved):
Looks fine to me, although I don't know if I have delved all the subtleties. What happens to the auto-populationed metadata such as `inference_type` if we take a SampleData file, infer it, send it back through `from_tree_sequence()`, the re-infer it? I'm guessing we simply overwrite the key that was there before, but I wonder if we should issue a warning?

Also, it seems a bit inconsistent that (AFAIK) the "empty" metadata field for plain JSON metadata in tsinfer is `{}` but in tskit is (as far as I gather from Ben), `None`. I have a slight preference for `None`, but a stronger preference for consistency?

benjeffery:
> Also, it seems a bit inconsistent that (AFAIK) the "empty" metadata field for plain JSON metadata in tsinfer is `{}` but in tskit is (as far as I gather from @benjeffery ), `None`? I have a slight preference for `None`, but a stronger preference for consistency between tsinfer & tskit.

See https://github.com/tskit-dev/tskit/issues/1084 `None` will be default when schema is absent, `{}`  when schema is present.

hyanwong:
> See [tskit-dev/tskit#1084](https://github.com/tskit-dev/tskit/issues/1084) `None` will be default when schema is absent, `{}` when schema is present.

Ah, great. So I guess the times when we inject `{}` in this PR it's because a schema (JSON-default?) is present ?

benjeffery:
One quick question before I do full review - I wonder if `tskit.canonical_json` should be used where you currently use `json.dumps`. I know it won't change results - but I think it is good to promote its usage.

jeromekelleher:
> One quick question before I do full review - I wonder if tskit.canonical_json should be used where you currently use json.dumps. I know it won't change results - but I think it is good to promote its usage.

:shrug: Sure, I can do it if you think it's a good idea, but it's not part of tskit documented public API so not sure who I'm promoting it to?

tsinfer/inference.py
```diff
@@ -667,14 +695,15 @@ def insert_missing_sites(
         # We can't perform parsimony inference if all sites are missing, and there's no
         # point if all non-missing sites are the ancestral state, so skip these cases
         if np.all(np.logical_or(G == tskit.MISSING_DATA, G == anc_value)):
+            metadata = _update_site_metadata(
+                site.metadata, inference_type=constants.INFERENCE_NONE
+            )
+            if schema is None:
+                metadata = _encode_raw_metadata(metadata)
```
benjeffery, line 702: If `schema is None` here then that is basically the same as the schema being the permissive json schema as the metadata is being json encoded without checking the schema. Would it be better just to default to the permissive schema? Or have I missed something about how this would break old code?

benjeffery (approved):
LGTM couple of comments - Once we do another round of improvements in tskit I think this can be cleaned up a bit.

tsinfer/inference.py
jeromekelleher: To support old code we don't want any schema - we don't want the metadata to be decoded automatically at all. That way analysis code will only break if you explicitly opt-in to using a Schema. We'll want to change this going forward, but this seems like a reasonable path forward.

tsinfer/inference.py
```diff
@@ -1553,20 +1593,41 @@ def get_samples_tree_sequence(self, map_additional_sites=True):
...
-        num_ancestral_individuals = len(tables.individuals)
 
-        # Currently there's no information about populations etc stored in the
-        # ancestors ts.
+        schema = self.sample_data.metadata_schema
+        tables.metadata_schema = tskit.MetadataSchema(schema)
+        tables.metadata = self.sample_data.metadata
+
+        schema = self.sample_data.populations_metadata_schema
+        if schema is not None:
```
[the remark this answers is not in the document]
jeromekelleher: We need the "if" here because we want to write the data as raw bytes rather than pass it in as a dictionary. It's ugly, but it was the best way I could think of to provide a path forward without breaking too much code.

tsinfer/inference.py
jeromekelleher: I've added an enum with a description of what the values mean. We should think about how to display documentation like this for a tree sequence.

jeromekelleher: https://github.com/tskit-dev/tskit/issues/1092

tsinfer/formats.py
jeromekelleher: Done.

PR #393, opened 2020-12-03
Add tsi_strerror function.

### LINKED ISSUE #318: Add tsi_strerror for error codes
These do get propagated back to the user, so we need messages.

jeromekelleher:
Closes #318

### DIFF
--- a/lib/err.c
+++ b/lib/err.c
@@ -1,0 +1,97 @@
+/*
+** Copyright (C) 2020 University of Oxford
+**
+** This file is part of tsinfer.
+**
+** tsinfer is free software: you can redistribute it and/or modify
+** it under the terms of the GNU General Public License as published by
+** the Free Software Foundation, either version 3 of the License, or
+** (at your option) any later version.
+**
+** tsinfer is distributed in the hope that it will be useful,
+** but WITHOUT ANY WARRANTY; without even the implied warranty of
+** MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+** GNU General Public License for more details.
+**
+** You should have received a copy of the GNU General Public License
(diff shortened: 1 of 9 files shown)


lib/err.c
```diff
@@ -0,0 +1,97 @@
...
+{
+    const char *ret = "Unknown error";
+
+    switch (err) {
+        case 0:
+            ret = "Normal exit condition. This is not an error!";
+            break;
+
+        case TSI_ERR_GENERIC:
+            ret = "Generic error - please file a bug report.";
```
benjeffery, line 33: Maybe "Generic tsinfer error"? Or is that context always clear in the stacktrace?

lib/tree_sequence_builder.c
```diff
@@ -655,7 +655,7 @@ tree_sequence_builder_make_pc_node(
     }
     min_parent_time -= PC_ANCESTOR_INCREMENT;
     if (min_parent_time <= mapped_child_time) {
-        ret = TSI_ERR_ASSERTION_FAILURE;
+        ret = TSI_ERR_PC_ANCESTOR_TIME;
```
benjeffery, line 658: Getting a coverage warning here - I assume it is a hard to reach error state?

benjeffery (approved):
LGTM! Just two little nits.

lib/tree_sequence_builder.c
jeromekelleher: Yeah, exactly. It's a hard error to provoke, and tsinfer test coverage is in general lower than it should be. This one only showed up in the coverage diff because I changed the name to something more sensible.

lib/err.c
jeromekelleher: Done.

Issue #303, opened 2020-07-17
Names and formats for mismatch_rate and recombination_rate parameters


hyanwong:
(Continuing on the discussion from #273 ), currently if we provide either the `mismatch_rate` or the `recombination_rate` parameter to match_samples or match_ancestors, it is expected to be an array of length the number of inference sites. These give the switching vs emission probabilities in the LS HMM. One thing to bear in mind is that with https://github.com/tskit-dev/tsinfer/pull/266 it may not be obvious how long to make this array. But it might also be that we want at least the recombination_rate to be specified as a true *rate* per unit sequence, akin to the recombination_rate parameter of msprime, rather than a set of values which will vary depending on the distance between adjacent sites (which is what we do currently, and which I think of as more of a probability than a rate).

For recombination, It would certainly make sense to me to have the `recombination_rate` parameter set to either a constant per-base rate for the whole genome, or possibly an msprime.RecombinationMap. That way we don't need to worry about the inference sites thing. For experts, we could provide an alternative parameter, say `recombination_distances` or `sitewise_recombination_probabilities` or something, that explicitly sets the switch probabilities for each site.

I don't have any particularly well formed ideas for the mismatch rate. I tend to set each value in the mismatch array to a constant for the whole genome, based on the median recombination probability, rather than setting each value as a proportion of the recombination probability at the previous site. I'm not sure what a sensible name or expected format for this parameter should be, so would welcome some discussion.

hyanwong:
This should be for 0.2.0, and should use the intervals.py file that I am reworking in https://github.com/tskit-dev/msprime/pull/1185 (see also https://github.com/tskit-dev/msprime/issues/1255#issuecomment-725535906)

hyanwong:
OK, so here's what I imagine doing. We include the new intervals.py file from msprime into tsinfer, to give us a `tsinfer.RateMap` class. For `tsinfer.infer` we have a new `recombination_rate` parameter which we just pass down to `match_samples` and `match_ancestors`, but we don't have a `mismatch_rate` parameter - if you want to set that to something other than the default of 1, you need to run MA and MS explicitly.

The match_ancestors `recombination_rate` parameter, could be documented something like this (with a similar wording for match_samples):

```
:param recombination_rate: Either a RateMap object, or a floating point value
    giving a single rate :math:`rho` across the entire sequence.
    Integrating this rate between adjacent sites provides a list of relative
    likelihoods of recombination, :math:`r`, which determine the trade off between 
[...]

Issue #260, opened 2020-03-31
New LS code doesn't allow mutation_rates of zero


hyanwong:
Even if a recombination_rate is >0, tsinfer bombs out with a -12 error (see https://github.com/tskit-dev/tsinfer/issues/257) when any of the elements of mutation_rate are 0. We should either relax this requirement, or test for zeros in the mutation_rate array before running matching, and fail in the Python code with a more useful error message (e.g. "Zero values not allowed in mutation_rate array")

hyanwong:
This is probably a specific case of the more general issue opened in  https://github.com/tskit-dev/tsinfer/issues/233

jeromekelleher:
This seems like less of an issue for now, since we've decided to have a ``relative_mismatch`` parameter which is a single value. We can reason about the semantics of having this as zero elsewhere, I guess.

Moving out of the 0.2.0 milestone anyway.

hyanwong:
Very good point. No need for this in 0.2.0. It would be nice to solve the zero recombination-rate case (https://github.com/tskit-dev/tsinfer/issues/233), though.

hyanwong:
I'm actually hitting this problem in the current tests, in the limit of a single inference site. In this case, there is no recombination probability, so the recombination array is set to [0], which therefore sets the mutation rate to 0 too, and inference fails.

I'm not sure what to do in this edge case (or indeed the edge case when we have more than one site but all recombination rates are 0). Should I just hack around it and set the mutation probabilities to (say) 0.1 @jeromekelleher ? Yuck, though!

hyanwong:
I guess we could say that if you specify a recombination_rate value, but there is only one inference site, then we should error out instead, the justification being that the recombination rate is meaningless in this case (and inference *will* work if no recombination rate is set). But users might reasonably want to set a recombination rate per base pair to the average genomic value, and not care that there's only one inference site, so I'm not sure erring out is the right thing to do.

jeromekelleher:
This depends on how we're computing the relative mutation rate though, doesn't it? I would have thought we'd compute this from the map *before* we compute the array of recombination proba values. So, the corner case becomes handled at that level, i.e., if there's only one site then we compute the mismatch value relative to the recombination rate over the whole map.

It's a remote corner case, so there's no point in worrying about how meaninginful the recombination rates are outside of the one site we are using for inference, it's really just about making the test suite work.

hyanwong:
[...]

PR #395, opened 2020-12-04
Mismatch recombination params

### LINKED ISSUE #236: Weight LS recombination rate by distance along the genome
Currently there's no concept of distance along the genome for recombination rates. We should weight the recombination rates by the distance back to the previous site (but, we need to be careful about what this means for the mutation rate).

See #235 for an example where we're not doing what we should.

hyanwong:
Mostly implements the calling convention and documentation in https://github.com/tskit-dev/tsinfer/issues/303, and also fixes #236. Needs more tests, and an implementation of RateMap to be included in tsinfer, but passes the existing tests. A few points for discussion:

* Instead of naming the parameter `relative_mismatch`, I have named it `mismatch_ratio` (a higher ratio gives more mismatch). I prefer the name because it puts mismatch first, and it *is* a ratio.

*  ~~I don't allow `mismatch_ratio` to be set in `tsinfer.infer()`, only in `match_ancestors()` and `match_samples`. Should I add it as a param to `.infer()`, and in this case use the same given ratio in both matching phases?~~

* I have changed "recombination_rate" and "mismatch_rate" in the C code to plain old "recombination" and "mismatch", as I think these aren't rates but probabilities. Calling them rates was starting to confuse me, especially as recombination_rate really is a (per base pair) value with a different meaning. At some point it would be logical to require an array of size (num_sites-1) for the recombination array passed into the C process, but I haven't bothered changing this yet, just left a comment.

* ~~There's a horrible hack for cases with only one inference site where a recombination_rate is specified. In this case, we have no recombination probability to use, as a recombination probability only applies between 2 sites or more. This results in recombination and mutation arrays of a single zero, which then fails due to #260 . This case is identified and the mismatch probability arbitrarily set to 0.1. We might be able to remove this hack if #260 is fixed.~~

* ~~The logic for checking that the user doesn't specify both `recombination_rate` and `recombination`, or `mismatch_ratio` and `mismatch`, or doesn't specify either mismatch value without a recombination specification, is a bit involved. Suggestions for improvements welcome.~~

* There are some arbitrary parameterisations here. For example, when neither recombination or mismatch are specified, we set recombination probs to 0.01, and mismatch to 1e-20. I tried setting recombination to 1, but this kills the matching process, presumably because it is working in probability space. ~~We should possibly restrict the parameters so that values >1 in either recombination probs or mismatch probs are not allowed. More properly, we shouldn't measure the cumulative mass between 2 sites if the recombination rates are high. In the limit, this probability should max out at 1, so we really need a different function to convert rates along the genome into probabilities of recombination at each site. This might be worth doing before 0.2. It shouldn't affect most datasets however, because when recombination rates are low it makes very little difference.~~ (see #398 )

### DIFF
--- a/tsinfer/inference.py
+++ b/tsinfer/inference.py
@@ -169,29 +169,46 @@ def check_sample_indexes(sample_data, indexes):
 def infer(
     sample_data,
     *,
-    num_threads=0,
+    recombination_rate=None,
+    mismatch_ratio=None,
     path_compression=True,
     simplify=True,
-    recombination_rate=None,
-    mismatch_rate=None,
-    precision=None,
     exclude_positions=None,
+    num_threads=0,
+    # Deliberately undocumented parameters below
+    precision=None,
     engine=constants.C_ENGINE,
     progress_monitor=None,
 ):
     """
-    infer(sample_data, *, num_threads=0, path_compression=True, simplify=True,\
-            exclude_positions=None)
+    infer(sample_data, *, recombination_rate=None, mismatch_ratio=None,\
+            path_compression=True, simplify=True, exclude_positions=None, num_threads=0)
 
     Runs the full :ref:`inference pipeline <sec_inference>` on the specified
     :class:`SampleData` instance and returns the inferred
-    :class:`tskit.TreeSequence`.
+    :class:`tskit.TreeSequence`.   See
+    :ref:`matching ancestors & samples<sec_inference_match_ancestors_and_samples>`
+    in the documentation for details of ``recombination_rate``, ``mismatch_ratio``
+    and ``path_compression``.
+
+    .. note::
+        For finer grained control over inference, for example to set different mismatch
+        ratios when matching ancestors versus samples, run
+        :func:`tsinfer.generate_ancestors`, :func:`tsinfer.match_ancestors` and
+        :func:`tsinfer.match_samples` separately.
 
     :param SampleData sample_data: The input :class:`SampleData` instance
         representing the observed data that we wish to make inferences from.
-    :param int num_threads: The number of worker threads to use in parallelised
-        sections of the algorithm. If <= 0, do not spawn any threads and
-        use simpler sequential algorithms (default).
+    :param recombination_rate: Either a RateMap object, or a floating
+        point value giving a single rate, :math:`\\rho`, across the entire sequence,
+        used to calculate the probability of recombination between adjacent sites.
+        If ``None``, all matching conflicts are resolved by recombination and
+        all inference sites will have a single mutation (equivalent to mismatch_ratio
+        near zero)
+    :type recombination_rate: Union[float, RateMap]
+    :param float mismatch_ratio: The probability of a mismatch relative to the median
+        probability of recombination between adjacent sites: can only be used if a
+        recombination rate has been set (default: 1)
     :param bool path_compression: Whether to merge edges that share identical
         paths (essentially taking advantage of shared recombination breakpoints).
     :param bool simplify: Whether to remove extra tree nodes and edges that are not
@@ -207,6 +224,9 @@ def infer(
         main inference process using parsimony. The list does not need to be
         in to be in any particular order, and can include site positions that
         are not present in the sample data file.
+    :param int num_threads: The number of worker threads to use in parallelised
+        sections of the algorithm. If <= 0, do not spawn any threads and
+        use simpler sequential algorithms (default).
     :return: The :class:`tskit.TreeSequence` object inferred from the
         input sample data.
     :rtype: tskit.TreeSequence
@@ -230,7 +250,7 @@ def infer(
         engine=engine,
         num_threads=num_threads,
         recombination_rate=recombination_rate,
-        mismatch_rate=mismatch_rate,
+        mismatch_ratio=mismatch_ratio,
         precision=precision,
         path_compression=path_compression,
         progress_monitor=progress_monitor,
@@ -241,7 +261,7 @@ def infer(
         engine=engine,
         num_threads=num_threads,
         recombination_rate=recombination_rate,
-        mismatch_rate=mismatch_rate,
+        mismatch_ratio=mismatch_ratio,
         precision=precision,
         path_compression=path_compression,
         simplify=simplify,
@@ -253,16 +273,17 @@ def infer(
 def generate_ancestors(
     sample_data,
     *,
-    num_threads=0,
     path=None,
     exclude_positions=None,
+    num_threads=0,
+    # Deliberately undocumented parameters below
     engine=constants.C_ENGINE,
     progress_monitor=None,
     **kwargs,
 ):
     """
-    generate_ancestors(sample_data, *, num_threads=0, path=None, \
-            exclude_positions=None, **kwargs)
+    generate_ancestors(sample_data, *, path=None, exclude_positions=None,\
+        num_threads=0, **kwargs)
 
     Runs the ancestor generation :ref:`algorithm <sec_inference_generate_ancestors>`
     on the specified :class:`SampleData` instance and returns the resulting
@@ -279,14 +300,14 @@ def generate_ancestors(
 
     :param SampleData sample_data: The :class:`SampleData` instance that we are
         genering putative ancestors from.
-    :param int num_threads: The number of worker threads to use. If < 1, use a
-        simpler synchronous algorithm.
     :param str path: The path of the file to store the sample data. If None,
         the information is stored in memory and not persistent.
     :param array_like exclude_positions: A list of site positions to exclude
         for full inference. Sites with these positions will not be used to generate
         ancestors, and not used during the copying process. The list does not
         need be in any particular order.
+    :param int num_threads: The number of worker threads to use. If < 1, use a
+        simpler synchronous algorithm.
     :return: The inferred ancestors stored in an :class:`AncestorData` instance.
     :rtype: AncestorData
     """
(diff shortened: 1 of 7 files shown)


tsinfer/inference.py
```diff
@@ -230,7 +244,7 @@ def infer(
         engine=engine,
         num_threads=num_threads,
         recombination_rate=recombination_rate,
-        mismatch_rate=mismatch_rate,
+        # mismatch_ratio=mismatch_ratio,
```
jeromekelleher, line 247: Looks like ``mismatch_ratio`` is commented out here?

tsinfer/inference.py
```diff
@@ -253,16 +267,17 @@ def infer(
 def generate_ancestors(
     sample_data,
     *,
-    num_threads=0,
     path=None,
     exclude_positions=None,
+    # Performance-related parameters below
```
jeromekelleher, line 272: I don't think this comment really helps - there's only one performance related parameter (num_threads) which is pretty self-explanatory.

tsinfer/inference.py
```diff
@@ -317,47 +332,81 @@ def match_ancestors(
...
-    mismatch_rate=None,
-    precision=None,
-    extended_checks=False,
-    engine=constants.C_ENGINE,
-    progress_monitor=None,
+    mismatch_ratio=None,
+    recombination=None,  # Deliberately undocumented: allows passing prob array directly
+    mismatch=None,  # Deliberately undocumented: allows passing prob array directly
+    path_compression=True,
+    # performance-related parameters below
```
jeromekelleher, line 340: I think it'd be make more sense to group the undocumented parameters together. Since they're all kwargs, the order isn't important for forward compatibility.  As above,  I don't think the classification of everything below as performance helps - progress_monitor and progress_monitor and engine and definitely not knobs to tweak to influence performance.

tsinfer/inference.py
```diff
@@ -317,47 +332,81 @@ def match_ancestors(
...
     complete ancestry of the putative ancestors.
 
+    The ``mismatch_ratio`` parameter determines the balance of recombination to
+    multiple mutations at a site. Setting a high ``mismatch_ratio`` will result in
+    tree sequences with more recurrent mutations and fewer recombinations (and edges).
+    Setting a low value will result in tree sequences with more recombination events
+    and edges, and fewer mutations. In the limit, as the mismatch_ratio tends to zero,
+    only one mutation will be inferred per variable site. This is the default
+    behaviour if no ``recombination_rate`` is given. Alternatively, if
+    ``recombination_rate`` is set, ``mismatch_ratio`` defaults to a setting that
```
jeromekelleher, line 363: We should say what this value is - "defaults to 1, which has been shown ..."

tsinfer/inference.py
```diff
@@ -317,47 +332,81 @@ def match_ancestors(
...
         representing the input data.
     :param AncestorData ancestor_data: The :class:`AncestorData` instance
-        representing the set of ancestral haplotypes that we are finding
-        a history for.
-    :param int num_threads: The number of match worker threads to use. If
-        this is <= 0 then a simpler sequential algorithm is used (default).
+        representing the set of ancestral haplotypes for which we are finding
+        a history.
+    :param recombination_rate: Either a RateMap object, or a floating
+        point value giving a single rate, :math:`\\rho`, across the entire sequence.
```
jeromekelleher, line 376: I think we should set aside a section of the documentation to describe and discuss the recombination and mismatch parameters in detail rather than trying to do it here in the context of the API docs. E.g. "See :ref:`sec_model_recombination_mutation`` for details on this parameter and how it interacts with the mismatch process".

We don't need to write this section now, can just make it a TODO for the moment (probably better, in order to get this PR merged ASAP).

tsinfer/inference.py
```diff
@@ -317,47 +332,81 @@ def match_ancestors(
...
+        Integrating this rate between adjacent inference sites results in a list of
+        estimates of relative probabilities of recombination :math:`r`, which
+        in turn determine, at each site, the trade off between recombination and
+        multiple mutation during the matching process. If None, prioritise
+        recombination and do not infer multiple mutations at a site.
+    :type recombination_rate: Union[float, RateMap]
+    :param float mismatch_ratio: The mismatch ratio for all sites. If a
+        ``recombination_rate`` has been provided, the underlying probability
```
jeromekelleher, line 384: Ditto here, let's refer this up to a documentation section.

tsinfer/inference.py
```diff
@@ -370,18 +419,21 @@ def augment_ancestors(
...
-    mismatch_rate=None,
-    precision=None,
-    extended_checks=False,
-    engine=constants.C_ENGINE,
-    progress_monitor=None,
+    mismatch_ratio=None,
+    recombination=None,  # Deliberately undocumented: allows passing prob array directly
+    mismatch=None,  # Deliberately undocumented: allows passing prob array directly
+    path_compression=True,
+    # Performance related parameters below
```
jeromekelleher, line 427: Ditto comments above - let's order the params by documented/undocumented, with a "# Following parameters are deliberately undocumented and may change arbitrarily in future versions".

tsinfer/inference.py
```diff
@@ -890,23 +980,77 @@ def __init__(
...
+        if recombination_rate is None and recombination is None:
+            if mismatch_ratio is None and mismatch is None:
+                # Special case: revert to tsinfer 0.1 behaviour with no mismatch allowed
+                mismatch = np.full(self.num_sites, 1e-20)  # A miniscule value
+                # arbitrary, large relative to mismatch, but < 1
+                if self.num_sites == 0:
+                    recombination = []
+                else:
+                    recombination = np.full(self.num_sites - 1, 1e-2)
+                # Shouldn't need insane precision, although 3 is not enough in trials
```
jeromekelleher, line 999: Let's keep precision out of this change - there's enough moving parts here. Keep a constant default, we can use something more sophisticated later.

tsinfer/inference.py
```diff
@@ -890,23 +980,77 @@ def __init__(
             [inference_site_position, [sample_data.sequence_length]]
         )
         self.position_map[0] = 0
+        self.recombination = np.zeros(self.num_sites)  # TODO: reduce len by 1
+        self.mismatch = np.zeros(self.num_sites)
+
+        # Set the canonical 'recombination' and 'mismatch' arrays to appropriate values
+        # NB 'recombination' must be of length num_sites-1, mismatch of length num_sites
```
jeromekelleher, line 987: I think we should simplify this a bit - you can *either* use the low-level ``recombination`` and ``mismatch`` parameters OR the high-level ``recombination_rate`` and ``mismatch_ratio`` but not mix them. Checking this first and deciding which path we're on will simplify things.

tsinfer/inference.py
```diff
@@ -890,23 +980,77 @@ def __init__(
...
+            # Default if recombination but no mismatch params have been specified
+            mismatch_ratio = 1.0
+
+        assert not (recombination_rate is None and recombination is None)
+        assert not (mismatch_ratio is None and mismatch is None)
+
+        # Set recombination from recombination_rate
+        if recombination is None:
+            try:
+                recombination = np.diff(
```
jeromekelleher, line 1014: Is this to support both pre and post 1.0 msprime RateMaps? If so, we should pull this out into a function so we can test it rather than being deeply nested in here.

tsinfer/inference.py
```diff
@@ -890,23 +980,77 @@ def __init__(
...
+
+        # Set mismatch from mismatch_ratio
+        if mismatch is None:
+            if len(recombination) > 0:
+                mismatch = np.full_like(
+                    self.mismatch, mismatch_ratio * np.median(recombination)
+                )
+            else:
+                # Hack if < 2 sites: https://github.com/tskit-dev/tsinfer/issues/260
+                mismatch = np.full_like(self.mismatch, 0.1)
```
jeromekelleher, line 1034: Why such a large value? This is basically the same case as pre 0.2 ``mismatch = np.full(self.num_sites, 1e-20)`` above, right?

tsinfer/inference.py
```diff
@@ -890,23 +980,77 @@ def __init__(
...
+        elif mismatch_ratio is not None:
+            raise ValueError(
+                "Cannot specify both a mismatch ratio and mismatch probabilities"
+            )
+        if len(mismatch) != self.num_sites:
+            raise ValueError("Bad length for mismatch array")
 
         if precision is None:
+            nonzero_probs = np.concatenate(
```
jeromekelleher, line 1043: Again, let's separate out precision from this update. It's not necessary and we want to change the minimum number of things we can when making fundamental changes like this.

jeromekelleher (commented):
Looks good, thanks @hyanwong. My main input would be 

1) Let's simplify the documentation and reduce duplication by creating a section in one of the rst files to discuss recombination/mismatch and put in pointers to that. We can settle on the text for that section in a follow up.
2) Keep a constant default of 12 or 14 for precision - no point in adding extra complexity here.
3) Simplify the logic for dealing with the high-level and low-level paremeters for the HMM by requiring the use of one or the other.

hyanwong:
> Looks good, thanks @hyanwong.

Great, thanks for the review @jeromekelleher . Happy with your suggestions below. With 1. below, shall I leave the documentation in the PR as-is, with a follow-up PR to remove the duplicate documentation and put into an `.rst` file? Or do you want me to remove the text it and replace with some boilerplate somehow?

Also, I've been wondering about the points I discuss in #398 - I feel that we should think about this, but perhaps it would be preferable to do that in a follow-up PR? Otherwise I could add another commit to this one, once we've decided on the best strategy?

> My main input would be
> 
> 1. Let's simplify the documentation and reduce duplication by creating a section in one of the rst files to discuss recombination/mismatch and put in pointers to that. We can settle on the text for that section in a follow up.
> 2. Keep a constant default of 12 or 14 for precision - no point in adding extra complexity here.
> 3. Simplify the logic for dealing with the high-level and low-level paremeters for the HMM by requiring the use of one or the other.

tsinfer/inference.py
hyanwong: That's because I don't currently allow it to be specified in the call to `infer()`; it's restricted to `.match_samples()` and `match_ancestors()` - see my point 2 in the intro. Do you think we should allow it in `infer()`, and then set the same mismatch_ratio in both phases of the algorithm?

tsinfer/inference.py
hyanwong: Fair enough - I was just trying to sort out a standardised order for listing parameters. Given that these are keyword-only, it's only to make things logical. And for the undocumented ones it doesn't really matter, I guess (see below).

jeromekelleher:
Great, thanks @hyanwong. I suggest keeping the documentation you have cut-pasted into the new section of the docs with a ``.. todo:: this documentation is incomplete``. I'd rather not commit the n copies of the current documentation as it's much easier to keep track of when when things are added first.

We can follow up on the details you're discussin  in #398 in a follow-up. Let's get the minimal version of this diff making the changes to the parameter names merged as soon as we can and make tweaks to it in follow-ups.

tsinfer/inference.py
hyanwong: OK. Happy to put whatever order you think is logical. As you say, since they are kwargs it's purely semantic.

tsinfer/inference.py
hyanwong: No, it's so that we can support either a floating point value *or* a RateMap (or indeed, anything that provides a `.get_cumulative_mass()` method). If no such method exists, we treat recombination_rate as a float.

tsinfer/inference.py
hyanwong: Oh, good point. We could set it to anything, I guess. I assume matching doesn't run anyway if there is only 1 inference site? Or maybe it does run as a special case. This would be run with the recombination probability set to 0, so if it were insanely small we might hit rounding errors if it is run as a special case. But I guess it worked before with recombination=0, mismatch=1e-20?

jeromekelleher: I don't know, but if we're going to pull values out of thin air we might as well minimise the number of different values in case readers think there's some particular significance to them.

tsinfer/inference.py
jeromekelleher: OK, let's make a ``parse_recombination_rate`` function then to abstract this out.

tsinfer/inference.py
hyanwong: Good point. I've added (but not pushed yet) a unit test for this, and 1e-20 seems fine, so I'll refactor with that in mind.

jeromekelleher:
Can you ping me again when this is ready for review please @hyanwong? I'm getting a lot of updates so going to silence the thread.

tsinfer/inference.py
hyanwong: I have flip-flopped on this and now put mismatch_ratio back into `infer()`. But if you want to specify the full probability arrays you can't use `infer()` - you have to do the 2 matching stages separately.

tsinfer/inference.py
hyanwong: Done, although I called it `recombination_rate_to_probs`

hyanwong:
I just cherry-picked your test fix @jeromekelleher, to allow this to pass the tests. Hope that's OK.

tsinfer/inference.py
```diff
@@ -890,23 +978,51 @@ def __init__(
...
+                recombination = recombination_rate_to_probs(
+                    recombination_rate, inference_site_position
+                )
+                if mismatch_ratio is None:
+                    mismatch_ratio = 1.0
+                mismatch = np.full(
+                    self.num_sites, mismatch_ratio * np.median(recombination)
+                )
+
+        if len(recombination) != max(0, self.num_sites - 1):
```
jeromekelleher, line 1017: Can use ``num_intervals`` here

jeromekelleher (commented):
LGTM, one minor tweak suggested.

Let's get the other PRs merged and then rebase.

tsinfer/inference.py
hyanwong: Great spot. Left over from when I didn't define that up top. Thanks.

jeromekelleher (approved):

PR #397, opened 2020-12-05
Put some basic tree seqs & sample files into fixtures

hyanwong:
This saves some code and is useful for future testing

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -26,15 +26,21 @@ it clear in test code.
 
 For example to use the `ts` fixture (a tree sequence with data in all fields) in a test:
 
-def test_something(ts):
-    assert ts.some_method() == expected
+class TestClass:
+    def test_something(self, ts_fixture):
+        assert ts_fixture.some_method() == expected
 
 Fixtures can be parameterised etc. see https://docs.pytest.org/en/stable/fixture.html
 
-Note that fixtures have a "scope" for example `ts` below is only created once per
-test session and re-used for subsequent tests.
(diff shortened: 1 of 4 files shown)


hyanwong:
I guess @benjeffery would be best placed to review this, as it's a pytest thing.

tests/conftest.py
```diff
@@ -59,3 +65,48 @@ def pytest_collection_modifyitems(config, items):
...
+    """
+    A simple 1-tree sequence with at least 2 inference sites
+    (i.e. mutations above a non-sample node)
+    """
+    ts = msprime.simulate(10, mutation_rate=1, random_seed=1)
+    assert num_nonsample_muts(ts) > 1
+    return ts
+
+
+@fixture(scope="session")
```
benjeffery, line 85: Is `SampleData` immutable? Otherwise you can't set `scope` to session.

benjeffery (approved):
LGTM - just one question about scope.

tests/conftest.py
hyanwong: Thanks Ben. You can't modify a SampleData file - you have to take a copy. There's a flag ".READ_MODE" on it that is set once it is made, and all write methods check this before doing anything.  ".READ_MODE" can't be unset AFAIK. So I think the answer is that practically, it is immutable, although perhaps that's not known by Python.

benjeffery:
@jeromekelleher not sure what out merging etiquette is for `tsinfer` yet, but this LGTM!

jeromekelleher (approved):
LGTM, nice to get rid of some of this code.

PR #401, opened 2020-12-07
Fix faulty test.

### LINKED ISSUE #399: Changing precision alters which sites are flagged for full inference?
Very weirdly, if I change the default precision value to 13 here:

https://github.com/tskit-dev/tsinfer/blob/7c23e61425254674b1caafa4c7419851d647141b/tsinfer/inference.py#L896

Then this test fails:

```
pytest tests/test_inference.py::TestMetadataRoundTrip::test_site_metadata
...
>               assert value == tsinfer.INFERENCE_FULL
E               AssertionError: assert 'parsimony' == 'full'
E                 - full
E                 + parsimony

tests/test_inference.py:819: AssertionError
============================================================== short test summary info ===============================================================
FAILED tests/test_inference.py::TestMetadataRoundTrip::test_site_metadata - AssertionError: assert 'parsimony' == 'full'
```

I have no idea why that should be

jeromekelleher:
Closes #399.

I think the test was backwards here  causing the problem noted in #399.

### DIFF
--- a/tests/test_inference.py
+++ b/tests/test_inference.py
@@ -791,13 +791,13 @@ class TestMetadataRoundTrip:
             assert all_metadata[variant.site.id] == variant.site.metadata
 
         output_ts = tsinfer.infer(sample_data)
-        samples = set(output_ts.samples())
-        for site in output_ts.sites():
+        for variant in output_ts.variants():
+            site = variant.site
             decoded_metadata = json.loads(site.metadata)
             assert "inference_type" in decoded_metadata
             value = decoded_metadata.pop("inference_type")
             # Only singletons should be parsimony sites in this simple case
-            if len(site.mutations) == 1 and site.mutations[0].node in samples:
+            if np.sum(variant.genotypes > 0) == 1:
                 assert value == tsinfer.INFERENCE_PARSIMONY
             else:
                 assert value == tsinfer.INFERENCE_FULL


hyanwong:
Thanks for fixing @jeromekelleher . It was causing me problems.

hyanwong (approved):
The fix looks good to me, but I struggle to see why is was failing before. Surely a site with only a single mutation which is above a sample should be a singleton, and if there is >1 mutation or there is 1 but it is not above a sample, it can't be a singleton. So I don't quite get the difference, which worries me a little.

hyanwong:
Ah, unary nodes! Here's the part of the tree causing the issue:

Screenshot 2020-12-08 at 08 54 22

This is for a parsimony site. So where there are unary nodes, `map_mutations()` puts them as high as possible (i.e. above the unary node. This contradicts the [documentation](https://tskit.readthedocs.io/en/latest/python-api.html#tskit.Tree.map_mutations) for map_mutations, which says:

"For each mutation, node is the tree node at the bottom of the branch on which the transition occurred,"

I don't know what is the most reasonable behaviour here, but I'm pleased that this came to light, as we sometime judge whether a mutation is a "sample mutation" if it's node is a sample node, but it's made me realised that with unary nodes, that need not be the case.

jeromekelleher:
@mergifyIO rebase

benjeffery (approved):

Issue #403, opened 2020-12-08
Are the HMM emission probabilities parameterised correctly?


hyanwong:
Currently, the emission probabilities seem to be given by this line: https://github.com/tskit-dev/tsinfer/blob/7f69d59b37217146b700364fa2de0da4fefc56c8/tsinfer/algorithm.py#L736

That means when `mu` is 1 (which I would take to mean that the emitted allele is completely unrelated to the haplotype we are copying from) then for a 2-allele model, p_e is `1- (2-1) * 1 = 0` Can this be correct? Shouldn't the minimum emission probability be `1/num_alleles`, @jeromekelleher ?

hyanwong:
Another thought: it would be better (at least in the match-samples phase) to use the actual allele frequency to determine the "uninformative" emission probabilities. If the derived allele is at freq `p`, then (assuming the underlying state is uninformative), the emissions should be `p` for the derived allele, and `1-p` for the ancestral allele.

This is a bit more difficult during the match_ancestors phase, because early in time, the derived allele will be at freq 0. But even if we used the current freq as a proxy for this, I suspect it would be better than using equal probabilities of 0.5/0.5.

jeromekelleher:
The emission proba is based on the existing treatments: see, .e.g, [Rosen and Paten](https://almob.biomedcentral.com/articles/10.1186/s13015-019-0144-9), [Lunter](https://academic.oup.com/bioinformatics/article/35/5/798/5079326) and [Stepens and Donelly](https://arxiv.org/abs/1006.1514).

We have dropped the "coalescenty" terms from the emission proba in, e.g., S&D, as these are constants. But, we are, I think, consistent with the other derivations. How do you see what we're doing as different from these treatments?

hyanwong:
Thanks for the links @jeromekelleher. I guess this is a question of the interpretation of the "mutation" rate: we aren't parameterising it as a mutation but potentially as an "error", which might change the angle we come at it.

Regardless, I can't see how it makes sense to allow negative emission probabilities in their formulations (e.g. for mu=1, A>2). I feel I should ask Gil about this anyway.

hyanwong:
[...]

PR #404, opened 2020-12-08
Correct recombination probabilities

### LINKED ISSUE #398: Convert cM to probability of recombination
As @JereKoskela pointed out, the distance in cM between sites is only a good measure of probability of recombination for small distances. To allow for larger distances, it might be sensible to convert the genetic distance to actual probability of recombination (e.g. see [the wikipedia article](https://en.wikipedia.org/wiki/Centimorgan#Relation_to_the_probability_of_recombination) ):

```
(1-exp(-2 * dist_in_morgans))/2
```

This would mean that users wouldn't be able to specify a pathological recombination probability, e.g. by setting recombination_rate to 1 in a 1Mb genome, which was causing problems in testing ([the last bullet point here](https://github.com/tskit-dev/tsinfer/pull/395#issue-532845030) ). It wouldn't stop users specifying probabilities greater than 1 if they set the `recombination` array directly, but that's not documented, so _caveat emptor_. However, the maximum probability that could occur along the genome would be 0.5 rather than 1, and therefore this line in the algorithm:

https://github.com/tskit-dev/tsinfer/blob/7f69d59b37217146b700364fa2de0da4fefc56c8/tsinfer/algorithm.py#L725

would max out at `(1 - 0.5 + 0.5/n)`, rather than the more expected, symmetrical value of `1/n` 

Perhaps this is OK, though, as the value is always going to be used in proportion to the mismatch rate? Otherwise we might want to e.g. multiply it by 2 - although we'd need to think through the rationale carefully.

That brings up the point that it would still, however, be possible to set a mismatch ratio which created mismatch probabilities greater than 1 because, at the moment, we simply multiply the recombination probability by the mismatch ratio. It might be useful to scale the mismatch ratio somehow too, so that even if someone set a ratio of (say 1000), if the recombination probability were 0.5, we didn't set a mismatch probability at that site of 500.0. That raises the question: if the recombination probability is maxed out, at near 0.5, what does it mean to have a mismatch probability that is 1000 times higher? Is that even a meaningful statement?

Incorporating the formula above into a (new) RecombinationMap object would be ideal, but for the moment, I think we should just hard code it into the tsinfer part of the algorithm, when converting rate to LS probabilities.

hyanwong:
This is a template for the sort of thing I suggest we might use to fix #398 . It still suffers from pathological probabilities with extreme values of e.g. mismatch_ratio, though. Some of this might need to be discussed in conjunction with #403

### DIFF
--- a/tsinfer/algorithm.py
+++ b/tsinfer/algorithm.py
@@ -741,10 +741,15 @@ class AncestorMatcher:
                 max_L_node = u
 
         if max_L == 0:
-            assert self.mismatch[site] == 0
-            raise ValueError(
-                "Trying to match non-existent allele with zero mutation rate"
-            )
+            assert self.mismatch[site] in (0, 1)
+            if self.mismatch[site] == 0:
+                raise ValueError(
+                    "Trying to match non-existent allele with zero mismatch rate"
+                )
+            if self.mismatch[site] == 1:
+                raise ValueError(
+                    "Match impossible: mismatch prob=1 & no haplotype with other allele"
+                )
 
         for u in self.likelihood_nodes:
             x = self.likelihood[u] / max_L
--- a/tsinfer/inference.py
+++ b/tsinfer/inference.py
@@ -343,8 +343,8 @@ def match_ancestors(
     path_compression=True,
     num_threads=0,
     # Deliberately undocumented parameters below
-    recombination=None,  # Allows direct setting of prob array of length (num_sites-1)
-    mismatch=None,  # Allows direct setting of prob array of length (num_sites)
+    recombination=None,  # See :class:`Matcher`
+    mismatch=None,  # See :class:`Matcher`
     precision=None,
     engine=constants.C_ENGINE,
     progress_monitor=None,
@@ -415,8 +415,8 @@ def augment_ancestors(
     path_compression=True,
     num_threads=0,
     # Deliberately undocumented parameters below
-    recombination=None,  # Allows direct setting of prob array of length (num_sites-1)
-    mismatch=None,  # Allows direct setting of prob array of length (num_sites)
+    recombination=None,  # See :class:`Matcher`
+    mismatch=None,  # See :class:`Matcher`
     precision=None,
     extended_checks=False,
     engine=constants.C_ENGINE,
@@ -498,8 +498,8 @@ def match_samples(
     force_sample_times=False,
     num_threads=0,
     # Deliberately undocumented parameters below
-    recombination=None,  # Allows direct setting of prob array of length (num_sites-1)
-    mismatch=None,  # Allows direct setting of prob array of length (num_sites)
+    recombination=None,  # See :class:`Matcher`
+    mismatch=None,  # See :class:`Matcher`
     precision=None,
     stabilise_node_ordering=False,
     extended_checks=False,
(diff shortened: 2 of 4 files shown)


hyanwong:
I have now updated this to bound both the recombination probabilities and mismatch probabilities at 0.5, as discussed in https://github.com/tskit-dev/tsinfer/issues/403#issuecomment-741878676. For our own reference, I have written a reasonably extensive discussion of the parameterisations & meaning of the arrays in the docstrings of the `Matcher` class, which are not exposed publicly. I guess that's OK @jeromekelleher ?

This should fix #398

tsinfer/inference.py
```diff
@@ -939,15 +939,34 @@ def run(self):
...
+    Note that the ``recombination`` and ``mismatch`` parameters can be used in
+    ``match_ancestors`` and ``match_samples`` and are passed directly to this
+    function, but are deliberately not publicly documented in those methods.
+    They are expected to be numpy arrays of length ``num_inference_sites - 1`` and
+    ``num_inference_sites`` respectively, containing values between 0 and 1, and
+    allow recombination and mismatch probabilities to be set directly. The
+    ``recombination`` probabilities measure the probability of a recombination event
+    between adjacent inference sites, used to calculate the HMM transition probabilities
+    in the L&S-like matching algorithm. When matching a haplotype against the ancestor
+    in the immediately previous generation, ``recombination`` probabilities should reach
```
jeromekelleher, line 958: "generation" is a bit misleading here - we don't have any concept of time and I think this will confuse people. A recombination proba of > 0.5 is always a weird thing, for example when dealing with LD calculations. This is a pretty standard issue when thinking about these extreme values for probability values in mutation and recombination processes. I think we can just say something like "Note that values > 0.5 for the recombination and mutation parameters will likely lead to pathological behaviour - for example, a mismatch probability of 1 means that a mismatch is *required* at every site.".

But nailing these parameters down as *probabilities* is very valuable and a great clarification!

tsinfer/inference.py
```diff
@@ -1076,6 +1099,42 @@ def __init__(
...
+        out at 0.5 as genetic_distance -> infinity
+        """
+        return (1 - np.exp(-genetic_distances * 2)) / 2
+
+    @staticmethod
+    def mismatch_ratio_to_probs(ratio, genetic_distances, num_alleles=2):
+        """
+        Convert a mismatch ratio, relative to a genetic distance, to a probability
+        of mismatch. A mismatch probability of 1 means that the emitted allele has a
+        100% probabilitie of being different from the allele implied by the hidden
```
jeromekelleher, line 1127: typo probabilitie

tsinfer/inference.py
```diff
@@ -1076,6 +1099,42 @@ def __init__(
...
+        Return the mean number of recombinations between adjacent positions (i.e.
+        the genetic distance in Morgans) given either a fixed rate or a RateMap
+        """
+        try:
+            return np.diff(rho.get_cumulative_mass(positions))
+        except AttributeError:
+            return np.diff(positions) * rho
+
+    @staticmethod
+    def recombination_dist_to_prob(genetic_distances):
```
jeromekelleher, line 1114: These functions are great. I think we just need a handful of test cases that illustrate the behaviour on small simple examples where we pre-calculate the expected result. This will make it easier to see what they do, also.

jeromekelleher (commented):
Looks great! Needs some tests and minor clarifications, but I think it's basically there.

tsinfer/inference.py
hyanwong: Yep. I suspect they are actually *likelihoods* (because they are conditional on a hypothesis, that this is the actual hidden state), but since everyone calls them probabilities, I guess we might as well stick with that.

hyanwong:
> Looks great! Needs some tests and minor clarifications, but I think it's basically there.

Great. Do we want an assert (perhaps only in the python version of the algorithm) that we have only 2 alleles, so that we don't forget we need to change the emission probs for > 2?

jeromekelleher:
I'll follow up with something to take care of that (#415)

tsinfer/inference.py
hyanwong: I've just pushed a few more commits with a load of tests for these functions. Look in particular for `test_inference.py`: `test_mismatch_ratio_to_prob_known`. Is that the sort of thing you mean @jeromekelleher ?

I put these as new commits so you can see the changes easily, but once you've OKed it I'll squash them all together, I guess?

tsinfer/algorithm.py
```diff
@@ -741,10 +741,15 @@ def update_site(self, site, haplotype_state):
...
 
         if max_L == 0:
-            assert self.mismatch[site] == 0
-            raise ValueError(
-                "Trying to match non-existent allele with zero mutation rate"
-            )
+            assert self.mismatch[site] in (0, 1)
+            if self.mismatch[site] == 0:
+                raise ValueError(
+                    "Trying to match non-existent allele with zero mismatch rate"
```
hyanwong, line 747: This error was not being covered by the code previously either!

jeromekelleher (approved):
LGTM!

Issue #402, opened 2020-12-08
Cannot match if mismatch probabilities are 1 (assertion fails in PY_ENGINE, path not found in C_ENGINE)


hyanwong:
In my attempts to crash the PY_ENGINE implementation of the matching algorithm, I find I can trigger an assertion error by setting the mismatch probabilities to 1. Perhaps we don't care, as this is only the python side and the C_ENGINE just returns `_tsinfer.LibraryError: Match impossible under input parameters.`, but I flag it up in case it's emblematic of a deeper issue. 

```
ts = msprime.simulate(10, mutation_rate=1, random_seed=1)
sd = tsinfer.SampleData.from_tree_sequence(ts, use_sites_time=False)
anc = tsinfer.generate_ancestors(sd)
num_loci = anc.num_sites
r = np.full(num_loci - 1, 0.1)
prob = 1
m = np.full(num_loci, prob)
tsinfer.match_ancestors(sd, anc, recombination=r, mismatch=m, engine=tsinfer.PY_ENGINE)
```

gives

```
  File "tsinfer/tsinfer/algorithm.py", line 910, in find_path
    self.update_site(site, h[site])
  File "tsinfer/tsinfer/algorithm.py", line 744, in update_site
    assert self.mismatch[site] == 0
```

Also if we set `prob` to much greater than 1 (e.g. 100), the PY_ENGINE never completes but seems hang at 100%CPU, even on this very small example. Interrupting it appears to suggest that it may be caught in an infinite `while` loop:

```
    left, right, parent = matcher.find_path(haplotype, start, end, match)
  File "tsinfer/tsinfer/algorithm.py", line 960, in find_path
    return self.run_traceback(start, end, match)
  File "tsinfer/tsinfer/algorithm.py", line 1011, in run_traceback
    while self.allelic_state[v] == -1:
```

If we implement https://github.com/tskit-dev/tsinfer/issues/398 then it should be impossible to set these pathological values using the documented interface, although it would still be possible to pass them in using the undocumented `mismatch` and `recombination` parameters, as above.

jeromekelleher:
Good stuff, thanks @hyanwong. Sorting out this behaviour will help us figure out what to do for #233

hyanwong:
I'm adding some tests to test_inference and flagging them up for skipping, so we have them there to test better behaviour.

hyanwong:
[...]

PR #411, opened 2020-12-09
Use Donnelly & Leslie parameterization of LS emission probs

hyanwong:
Here's my suggestion for how we should be currently parameterising mismatch, following [Donnelly & Leslie](https://arxiv.org/abs/1006.1514) eq 4.2 (see #403) I haven't touched the C code yet. I think this should go some way to fixing https://github.com/tskit-dev/tsinfer/issues/402 too, as the mismatch probability can never be 1 in this parameterisation (although I guess it can get close, if there are lots of haplotypes)

### DIFF
--- a/tsinfer/algorithm.py
+++ b/tsinfer/algorithm.py
@@ -722,6 +722,7 @@ class AncestorMatcher:
                 assert v != -1
 
             p_last = self.likelihood[u]
+            # Stephens & Donnelly Eq 4.1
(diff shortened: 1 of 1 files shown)


hyanwong:
Closed in favour of #404 as we are not using the Donnelly & Leslie formulation (see https://github.com/tskit-dev/tsinfer/issues/403)

PR #413, opened 2020-12-09
Do not include missing allele in tree seq builder num_alleles

### LINKED ISSUE #406: Add "exclude_missing" to num_alleles() method?
It strikes me that when we use `num_alleles` in the algorithm, we should probably not be counting `-1` as an allele, e.g. here:https://github.com/tskit-dev/tsinfer/blob/7f69d59b37217146b700364fa2de0da4fefc56c8/tsinfer/algorithm.py#L736

I wonder if we should be populating the tree sequence builder `num_alleles` array with the number without missing data? It seems odd that the HMM probabilities might depend on whether we allowed an extra allele type for missing data or not.

If I'm right (not 100% sure yet) then perhaps the num_alleles method could have an extra param (e.g. "exclude_missing", default false, so that previous behaviour is maintained) https://github.com/tskit-dev/tsinfer/blob/7f69d59b37217146b700364fa2de0da4fefc56c8/tsinfer/formats.py#L1898

And we would set `exclude_missing=True` here:
https://github.com/tskit-dev/tsinfer/blob/7f69d59b37217146b700364fa2de0da4fefc56c8/tsinfer/inference.py#L931

hyanwong:
Fixes #406 and addresses https://github.com/tskit-dev/tsinfer/issues/403#issuecomment-741846467 about counting missing data as an allele.

I'm slightly tempted to default to "count_missing=False", but this would not be backward compatible. OTOH, I'm not sure the missing data stuff has been previously released anyway. What do you think @jeromekelleher ?

### DIFF
--- a/tests/test_formats.py
+++ b/tests/test_formats.py
@@ -1142,6 +1142,28 @@ class TestSampleData(DataContainerMixin):
                         0, [0, 0], alleles=[str(x) for x in range(num_alleles)]
                     )
 
+    def test_num_alleles_with_missing(self):
+        u = tskit.MISSING_DATA
+        sites_by_samples = np.array(
+            [
+                [u, u, u, 1],
+                [u, u, u, 1],
+                [u, u, u, 1],
+                [u, 0, 0, 1],
+                [u, 0, 1, 1],
+                [u, 0, 1, 0],
+            ],
+            dtype=np.int8,
+        )
+        with tsinfer.SampleData() as sd:
+            for col in range(sites_by_samples.shape[1]):
+                genos = sites_by_samples[:, col]
+                alleles = [None if x == u else str(x) for x in np.unique(genos)]
+                if alleles[0] is None:
+                    alleles = alleles[1:] + [None]  # Put None at the end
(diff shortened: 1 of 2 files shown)


hyanwong:
FWIW, the tests pass if we default to `count_missing=False' too.

jeromekelleher:
Looks good, but why would we want to count the missing data as an allele? I think it's just a bug and the parameter isn't needed. There's nothing to be compatible with, this is all unreleased.

hyanwong:
Good point. Parameter removed. Now if `None` is last in the allele list we don't count it as an allele. I guess that precludes people from adding "None" as true allele type, which is probably fine anyway.

jeromekelleher:
It's what we do [here](https://github.com/tskit-dev/tsinfer/blob/main/tsinfer/inference.py#L785) already anyway. Which is why I think our existing analyses should be fine?

hyanwong:
Yep - just merge it I think.

jeromekelleher (approved):

jeromekelleher:
@mergifyio rebase

PR #414, opened 2020-12-09
Check that mismatch and recombination are bounded

hyanwong:
This is good practice anyway - it should mean that we rarely have pathological settings. It has picked up some problematic settings in the tests which should be fixed by #404, which is why it's not currently passing.


jeromekelleher:
Yep, good practise to bound these as probablilties.

hyanwong:
I'll tag this onto the end of #404

Issue #420, opened 2020-12-11
find_path fails on long stretches of zero-recombination, even with mismatch


hyanwong:
@awohns was finding it impossible to infer some simulated data using the HapMap chr20 recombination map, which has a long region of zero-recombination-rate over the centromere. By my reckoning, we should always be able to match, even in zero-rate regions, if the mismatch rate is > 0, as we should always be able to match to the oldest (root) ancestor, with all conflicting sites mismatching.

Here's some code that replicated the problem:

```python
ts = msprime.simulate(
    10,
    length=1e5, # Currently fails with length=1e5, but passes with length=1e4
    Ne=10000,
    mutation_rate=1e-8,
    recombination_rate=1e-8,
    random_seed=1,
) 
sd = tsinfer.SampleData.from_tree_sequence(ts, use_sites_time=False)
anc = tsinfer.generate_ancestors(sd)
m = np.full(anc.num_sites, 1e-4)
# Ban recombination
r = np.full(anc.num_sites - 1, 0)
anc_ts = tsinfer.match_ancestors(sd, anc, recombination=r, mismatch=m)
ts = tsinfer.match_samples(sd, anc_ts, recombination=r, mismatch=m)
```

hyanwong:
A few more observations. It also fails using PY_ENGINE, which is good, as it will help us debug. It fails even if path_compression is false in both matching stages. It fails because `max_L` is 0 here https://github.com/tskit-dev/tsinfer/blob/d74f0e1f5f97418ca381f76c21e1300e9e8c4b56/tsinfer/algorithm.py#L743

This is 0 because the transmission probability `p_t` is 0, which in turn is `p_last` is 0 for (in this case) node 22

hyanwong:
Further, in this case, the 2 nodes flagged as `self.likelihood_nodes` are nodes 1 and 22, both of which have a likelihood of 0. It looks like we might not be calculating the likelihood of matching against the oldest node (node 0?)

PR #421, opened 2020-12-11
Zero recombination rate

hyanwong:
Helps to debug #420

### DIFF
--- a/lib/err.c
+++ b/lib/err.c
@@ -63,7 +63,17 @@ tsi_strerror(int err)
             ret = "Bad path information: right > num_sites";
             break;
         case TSI_ERR_MATCH_IMPOSSIBLE:
-            ret = "Match impossible under input parameters.";
+            ret = "Unexpected failure to find matching haplotype; please open "
+                  "an issue on GitHub";
+            break;
+        case TSI_ERR_MATCH_IMPOSSIBLE_EXTREME_MUTATION_PROBA:
+            ret = "Cannot find match: the specified mismatch probability is "
+                  "0 or 1 and no matches are possible with these parameters";
+            break;
+        case TSI_ERR_MATCH_IMPOSSIBLE_ZERO_RECOMB_PRECISION:
+            ret = "Cannot find match: the specified recombination probability is"
+                  "zero and no matches could be found. Increasing the 'precision' "
+                  "may help, but recombination values of 0 are not recommended.";
             break;
         case TSI_ERR_BAD_HAPLOTYPE_ALLELE:
             ret = "Input haplotype contains bad allele information.";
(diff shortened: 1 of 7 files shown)


jeromekelleher:
Good stuff, I'll look at this and tack on a commit to resolve.

jeromekelleher:
The problem was to do with the precision parameter @hyanwong, nothing to get worried about I think. I've added a specific exception to cover this case, so we can catch it more effectively.

hyanwong:
Thanks for looking into this. Interesting that precision was causing an issue even with the default setting of 13, when mismatch was pretty large (1e-4), and there were only 11 sites. You'll have to explain to me some time!

lib/err.c
```diff
@@ -63,7 +63,17 @@ tsi_strerror(int err)
             ret = "Bad path information: right > num_sites";
             break;
         case TSI_ERR_MATCH_IMPOSSIBLE:
-            ret = "Match impossible under input parameters.";
+            ret = "Unexpected failure to find matching haplotype; please open "
+                  "an issue on GitHub";
+            break;
+        case TSI_ERR_MATCH_IMPOSSIBLE_EXTREME_MUTATION_PROBA:
+            ret = "Cannot find match: the specified mutation probability is "
```
hyanwong, line 70: Since this is in tsinfer, not tskit, should we be calling these mismatch probabilities, rather than mutation probabilities?

tests/test_inference.py
```diff
@@ -3228,6 +3229,63 @@ def test_bad_mismatch(self, small_sd_anc_fixture):
...
+            length=1e4,
+            Ne=10000,
+            mutation_rate=1e-8,
+            recombination_rate=1e-8,
+            random_seed=50,
+        )
+        sd = tsinfer.SampleData.from_tree_sequence(ts, use_sites_time=False)
+        anc = tsinfer.generate_ancestors(sd)
+        # Need to be sure that mu is large here or the value associated with the
+        # root hapoltype can become less than precision, and we therefore
```
hyanwong, line 3248: typo. "haplotype"

jeromekelleher:
> Thanks for looking into this. Interesting that precision was causing an issue even with the default (high IMO) setting of 13, when mismatch was pretty large (1e-4), and there were only 11 sites.

It does seem pathological all right - we must be in a regime where the renormalising isn't effective. Since it is pathological though, I'm not going to worry too much about it. Let's see if the error crops up with real data, the error message is pretty clear now.

hyanwong:
I should say that most of the recombination maps we are likely to use will have zero recombination regions, which is why I think we will hit this with real data.

jeromekelleher:
> I should say that most of the recombination maps we are likely to use will have zero recombination regions, which is why I think we will hit this with real data.

We should hit it pretty quickly then in that case - we can follow it up then if we do.

PR #422, opened 2020-12-11
Output which haplotype fails

### LINKED ISSUE #419: Output the haplotype being matched when find_path fails
When debugging some problems with haplotypes that include the centromere. @awohns and I needed to find out which ancestral haplotypes were failing, which was hard, because the only error message was

```
File "tsinfer/tsinfer/inference.py", line 1149, in _find_path
    left, right, parent = matcher.find_path(haplotype, start, end, match)
_tsinfer.LibraryError: Match impossible under input parameters.
```

I hacked around it by wrapping the `matcher.find_path` call in a try-except block, catching the `_tsinfer.LibraryError` and adding the ``child_id`` variable to the message then re-raising. I think this might be a useful thing to do all the time, to allow people to track down what's causing the problems we see in https://github.com/tskit-dev/tsinfer/issues/233.

hyanwong:
Fixes #419, but not tests yet, until the approach is approved

### DIFF
--- a/tsinfer/inference.py
+++ b/tsinfer/inference.py
@@ -1235,7 +1235,14 @@ class AncestorMatcher(Matcher):
         end = ancestor.end
         assert ancestor.haplotype.shape[0] == (end - start)
         haplotype[start:end] = ancestor.haplotype
-        self._find_path(ancestor.id, haplotype, start, end, thread_index)
+        try:
+            self._find_path(ancestor.id, haplotype, start, end, thread_index)
+        except _tsinfer.LibraryError as e:
+            if "Match impossible" in str(e):
+                raise RuntimeError(
+                    f"Could not find valid copying path for ancestor {ancestor.id}"
+                ) from e
+            raise
 
     def __start_epoch(self, epoch_index):
         start, end = self.epoch_slices[epoch_index]
@@ -1484,8 +1491,15 @@ class SampleMatcher(Matcher):
             )
         )
 
-    def __process_sample(self, sample_id, haplotype, thread_index=0):
-        self._find_path(sample_id, haplotype, 0, self.num_sites, thread_index)
+    def __process_sample(self, orig_sample_id, sample_id, haplotype, thread_index=0):
+        try:
+            self._find_path(sample_id, haplotype, 0, self.num_sites, thread_index)
+        except _tsinfer.LibraryError as e:
+            if "Match impossible" in str(e):
+                raise RuntimeError(
+                    "Could not find valid copying path "
+                    f"for sample {orig_sample_id} (internal id {sample_id})"
+                ) from e
 
     def __match_samples_single_threaded(self, indexes):
         sample_haplotypes = self.sample_data.haplotypes(
@@ -1493,7 +1507,7 @@ class SampleMatcher(Matcher):
         )
         for j, a in sample_haplotypes:
             assert len(a) == self.num_sites
-            self.__process_sample(self.sample_id_map[j], a)
+            self.__process_sample(j, self.sample_id_map[j], a)
 
     def __match_samples_multi_threaded(self, indexes):
         # Note that this function is not almost identical to the match_ancestors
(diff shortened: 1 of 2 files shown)


tsinfer/inference.py
```diff
@@ -1174,7 +1174,14 @@ def __ancestor_find_path(self, ancestor, thread_index=0):
         end = ancestor.end
         assert ancestor.haplotype.shape[0] == (end - start)
         haplotype[start:end] = ancestor.haplotype
-        self._find_path(ancestor.id, haplotype, start, end, thread_index)
+        try:
+            self._find_path(ancestor.id, haplotype, start, end, thread_index)
```
jeromekelleher, line 1178: Rather than wrapping ``_find_path`` twice, perhaps we should put the try-except in the place that can fail in that function?

tsinfer/inference.py
```diff
@@ -1174,7 +1174,14 @@ def __ancestor_find_path(self, ancestor, thread_index=0):
...
+        except _tsinfer.LibraryError as e:
+            if "Match impossible" in str(e):
```
jeromekelleher, line 1180: Matching on the strerror is brittle. What we should do is add specific exception in the C API and raise this exception when it happens. We might even be able to do this at the Python-C API directly, and avoid some boilerplate.

Actually, that's what we should do.

jeromekelleher (commented):
Looks good, but we should intercept the error at a lower level. Leave it with me.

tsinfer/inference.py
hyanwong: OK - I'll leave it with you. Thanks

hyanwong:
Only just seen your commented here @jeromekelleher - sorry. I've just pushed a version with tests, so you could just nick the tests and the wording?

tsinfer/inference.py
hyanwong: Yep, but then we find it difficult to distinguish between ancestors and sample haplotypes, which is why I put it in twice.

It's actually a bit of a pain to find the ID of the sample haplotype (see my latest push) but worth it, I think.

jeromekelleher:
It's probably best to tackle this one from the ground up so we get informative errors about what type of failure we got. So, I'll pick it up as part of #421 (which is another subtly different type of error).

I'll pull in your tests there.

jeromekelleher:
After adding the MismatchImpossible error in #421 I'm wondering if there's any real need for this - these errors only occur when we have extreme values for the recombination and mismatch parameters. Does it actually do you any extra good to know which sample/ancestor ID the error happened in, above being told whether it was the extreme mutation rates or 0 recombination rate?

hyanwong:
Well, it was very helpful in working out what the problem was with Wilder's real dataset. I wouldn't have realised it was the centromere until I saw that the haplotype that was being matched basically spanned the centromere. And I couldn't work out what the haplotype was until I hacked the python code.

In general I find the C errors not very helpful, because they often don't say which data was being processed by the algorithm when something went wrong. So being able to output that data feels right to me, and it seemed like a simple fix in this case.

hyanwong:
Is there any reason why we wouldn't want to tell the user which bits of the dataset were causing the problem? Is it simply that looking for the C error on the python side is a bit messy?

jeromekelleher:
Sure, I'll update once #421 is merged.

hyanwong:
> Sure, I'll update once #421 is merged.

Great, thanks for this. I do think it will be helpful down the line.

jeromekelleher:
It's quite fiddly getting good debug output here, and I'm wondering if it's really needed. Have you tried using the logging facilities @hyanwong? You'll get the same information, I think, if you turn it up to DEBUG mode. This what the logging is for, after all. I'd rather improve the logging code rather than clutter things up with lots of super specialised error messages - it just makes it really hard to see where the actual code is.

hyanwong:
Hm, debug mode is a good idea. I didn't try that. Thanks.

p.s. @awohns is still getting "cannot find a path" errors with standard recombination maps (that have regions of 0 rate), even with pretty high precision levels, I think. Is that right @awohns ?

jeromekelleher:
> p.s. @awohns is still getting "cannot find a path" errors with standard recombination maps (that have regions of 0 rate), even with pretty high precision levels, I think. Is that right @awohns ?

Let's open an issue to track, if so.

PR #423, opened 2020-12-14
Re-enable 3.9 now that 3.9.1 is released

benjeffery:
Due to #376 we had disabled testing and wheel building on python 3.9. 3.9.1 should fix this.

PR #424, opened 2020-12-14
Macos py39 debug

jeromekelleher:
The debug output in #423 seems to say that we're segfaulting when adding mutations, on the ``self`` pointer being NULL. This *should* be impossible, as we've checked for that before running any code.

The only thing I can think of is some weird threading thing, so trying to disable that first to see what happens.


jeromekelleher:
Hmm, nope, that's not it.

hyanwong:
Is there any clue in the fact that this seems not to happen when `-O0` is set as the optimisation level?

PR #428, opened 2020-12-17
Get Python 3.9 working on CI

jeromekelleher:
... hopefully

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -28,7 +28,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python: [ 3.6, 3.8 ]
+        python: [ 3.6, 3.9 ]
         os:  [ macos-latest, ubuntu-latest, windows-latest ]
     defaults:
       run:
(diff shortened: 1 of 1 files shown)


jeromekelleher:
Woot, works!

benjeffery:
WOW! How did you realise it was a compiler mis-match?!

benjeffery (approved):
LGTM! We should copy this across to the other repos too.

jeromekelleher:
> WOW! How did you realise it was a compiler mis-match?!

Clang 12 was being a bit painful so I thought I'd factor that out first, and mixed compiler versions are a definite no-no on Windows. I'm not sure it even is the compiler, I think it's some other part of the low-level toolchain like the linker or something.

jeromekelleher:
Merging manually as we're updating the github workflows.

jeromekelleher:
> EDIT: Just seen the branch name. ￼

Snappy, eh?

jeromekelleher:
What about the wheel builds, though?

benjeffery:
Yep, they will need this too. Things like this is also why I'd like to run the test suite on the built wheels if it isn't too much work.

jeromekelleher:
> Yep, they will need this too. Things like this is also why I'd like to run the test suite on the built wheels if it isn't too much work.

Hmm, I'm a bit nervous about mixing compilers for the shipped wheels. We should be building these without using conda, but using  "native" compiler. I'll take a look.

jeromekelleher:
No, there's no conda in the wheel building so we're all good. No need to change things I think.

benjeffery:
> No, there's no conda in the wheel building so we're all good. No need to change things I think.

Phew! Sorry, I assumed I'd had to resort to that.

Issue #430, opened 2020-12-18
How to create a RateMap suitable for use as the recombination_rate param in match_samples/ancestors


hyanwong:
At the moment we document that you can pass either a floating point number *or* an `msprime.RateMap` as the `recombination_rate` parameter in `match_samples` and `match_ancestors`: more specifically, we look for an object that has the `get_cumulative_mass()` method which was only recently merged as a RateMap method in the msprime main branch.

However, since this isn't yet part of a released msprime version, there isn't a way for a user to create such a RateMap without installing the msprime dev version.

We previously thought we would simply duplicate the msprime `intervals.py` file into tsinfer for the time being, but @jeromekelleher 's not that keen. Simplest at the moment is to remove documentation referring to RateMap in tsinfer, until msprime 1.0 alpha is released, then we can tell users to grab that if they want to use the RateMap functionality.

awohns:
Related to this, I think we should fail when the `get_cumulative_mass()` returns 0 recombination probabilities. Right now the error message says they are "not recommended", but does matching every work with recombination probabilities equal to 0?
If we want to support recombination maps, then should either change 0 probabilities to some very small probability epsilon, or raise an error immediately with information about how to fix the given recombination map

hyanwong:
Yes, it does sometimes work with values of 0. @jeromekelleher thinks it is purely a precision question (see #421 ). I'm not so sure. I guess this discussion belongs in #420 ?

awohns:
Sounds good re: moving that discussion in #420.
A more pertinent question to this issue though is what to do when the ratemap and the site positions are incompatible. For example, if sites exist beyond end of the ratemap we could either error out or, more sensibly I think, assign a recombination probability of zero to those sites.
We should probably start catching these kinds of cases for when the ratemap is documented.

hyanwong:
The new read_hapmap function takes a `sequence_length` param, which assigns a zero rate at all points past the end of the last site in the text file. I think that should do it.

PR #431, opened 2020-12-18
Temporarily remove references to RateMap

hyanwong:
Needed before release 0.2 (see #430). This undocuments the passing of RateMap as a recombination_rate parameter. We still allow it to be used, so that @awohns can use tsinfer 0.2 as the basis of his analyses. The idea is to revert this PR when msprime 1.0 alpha is released.


jeromekelleher (approved):
Great, thanks.

jeromekelleher:
@benjeffery This PR isn't merging even though it's all hunky dorey. I think this is because we've got a rule about needing tests to pass on platforms we're not testing on any more?

jeromekelleher:
@mergifyio rebase

PR #467, opened 2021-02-12
Truncate old ancestors

### LINKED ISSUE #276: Long ancient ancestors / slowness in HGDP analysis

awohns:
Adds a function to the `AncestorData` class which provides for the truncation of old ancestors based on the length observed in younger ancestors.
Minor decision point was how to handle the lengths of haplotypes: we decided to extend the haplotype to left and right of the leftmost and rightmost focal sites by half of the maximum haplotype observed in the user-specified interval. 
It's not a perfect solution, but are we happy to close #276 when this is merged?

### DIFF
--- a/tests/test_formats.py
+++ b/tests/test_formats.py
@@ -1768,7 +1768,9 @@ class TestAncestorData(DataContainerMixin):
 
     class_to_test = formats.AncestorData
 
-    def get_example_data(self, sample_size, sequence_length, num_ancestors):
+    def get_example_data(
+        self, sample_size, sequence_length, num_ancestors, use_sites_time=True
+    ):
         ts = msprime.simulate(
             sample_size,
             recombination_rate=1,
@@ -1776,7 +1778,9 @@ class TestAncestorData(DataContainerMixin):
             length=sequence_length,
             random_seed=100,
         )
-        sample_data = formats.SampleData.from_tree_sequence(ts)
+        sample_data = formats.SampleData.from_tree_sequence(
+            ts, use_sites_time=use_sites_time
+        )
 
         num_sites = sample_data.num_sites
         ancestors = []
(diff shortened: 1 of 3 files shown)


tsinfer/formats.py
```diff
@@ -2480,6 +2480,127 @@ def insert_proxy_samples(
...
+        **kwargs,
+    ):
+        """
+        Truncates the length of ancestors above a given time and returns a new
+        :class:`.AncestorData` instance.
+
+        The maximum length of an ancestor existing between ``lower_time_bound'' and
+        ``upper_time_bound'', multiplied by ``length_multplier'', is used to truncate
+        the length of all ancestors older than ``upper_time_bound'' which contain
+        at least one ``focal_site''.
```
jeromekelleher, line 2497: Small nit here: writing ``focal_site`` like this implies that it's an input parameter or other variable. But really, it's just a "focal site", right? Same point in next para

tsinfer/formats.py
```diff
@@ -2480,6 +2480,127 @@ def insert_proxy_samples(
...
+
+        Specifically, the haplotypes of ancestors older than ``upper_time_bound'' may
+        extend no further than *half of this maximum length* to the either side of the
+        leftmost and rightmost focal sites on each ancestor. Note that ancestors above
+        ``upper_time_bound'' may still be longer than the maximum length if the
+        ancestor contains greater than 2 ``focal_sites''.
+
+        This function should be used when :func:`tsinfer.generate_ancestors` generates
+        old ancestors which are very long, as this can adversely affect performance.
```
jeromekelleher, line 2506: "performance" could be interepreted as "statistical performance". Better to be concrete "as this can significantly slow down matching time".

tsinfer/formats.py
```diff
@@ -2480,6 +2480,127 @@ def insert_proxy_samples(
...
+        Older ancestors should generally be shorter than younger ancestors, so
+        truncating the lengths of older ancestors has negligible effect on inference
+        accuracy.
+
+        :param float lower_time_bound: Defines the lower bound of the closed interval
```
jeromekelleher, line 2511: Usually in tskit we make intervals open-closed, so it might be confusing to have a fully closed interval here. It'll make no difference in practise. Might be worth writing down the operation in psuedocode

```
Given a set of haplotypes H such that lower_time_bound <= h.time < upper_time_bound, we let max_len = max(max(h.length) for h in H). Then, we truncate all haplotypes where h.time >= upper.
```

tsinfer/formats.py
```diff
@@ -2480,6 +2480,127 @@ def insert_proxy_samples(
...
+        end = self.ancestors_end[:]
+        time = self.ancestors_time[:]
+        focal_sites = self.ancestors_focal_sites[:]
+        haplotypes = self.ancestors_haplotype[:]
+        if upper_time_bound > np.max(time) or lower_time_bound > np.max(time):
+            raise ValueError("Time bounds cannot be greater than older ancestor")
+
+        truncated = self.copy(**kwargs)
+        anc_in_bound = np.logical_and(
+            self.ancestors_time[:] >= lower_time_bound,
```
jeromekelleher, line 2551: You've already got a copy of ``time`` in a local variable, should use it.

tsinfer/formats.py
```diff
@@ -2480,6 +2480,127 @@ def insert_proxy_samples(
...
+            self.ancestors_time[:] <= upper_time_bound,
+        )
+        if np.sum(anc_in_bound) == 0:
+            raise ValueError("No ancestors in time bound")
+        max_length = length_multiplier * np.max(self.ancestors_length[:][anc_in_bound])
+        if max_length > self.sequence_length:
```
jeromekelleher, line 2557: I'm not sure this is worth raising an error for. It's not going to do any harm if ``max_length`` does happen to be > L, right? It'll just end up being annoying where this error gets tripped in corner cases that don't really matter anyway.

tsinfer/formats.py
```diff
@@ -2480,6 +2480,127 @@ def insert_proxy_samples(
...
+        :rtype: AncestorData
+        """
+        self._check_finalised()
+        if upper_time_bound <= 0 or lower_time_bound <= 0:
+            raise ValueError("Time bounds cannot be negative")
+        if length_multiplier <= 0:
+            raise ValueError("Length multiplier cannot be negative")
+        if upper_time_bound < lower_time_bound:
+            raise ValueError("Upper bound must be >= lower bound")
+        sites = self.sites_position[:]
```
jeromekelleher, line 2540: ``position`` would be easier to read below rather than ``sites``

tsinfer/formats.py
```diff
@@ -2480,6 +2480,127 @@ def insert_proxy_samples(
...
+            )
+
+        for anc in self.ancestors():
+            if anc.time > upper_time_bound and len(anc.focal_sites) > 0:
+                if sites[anc.end - 1] - sites[anc.start] > max_length:
+                    left_focal_pos = sites[np.min(anc.focal_sites)]
+                    right_focal_pos = sites[np.max(anc.focal_sites)]
+                    if right_focal_pos - left_focal_pos > max_length:
+                        logger.debug(
+                            "Ancestor contains focal sites > max_length "
```
jeromekelleher, line 2571: Looks like you wanted this to be an f-string? Not sure how helpful this debug message is going to be TBH, may just end up being confusing.

tsinfer/formats.py
```diff
@@ -2480,6 +2480,127 @@ def insert_proxy_samples(
...
+                        )
+                    insert_pos_start = np.maximum(
+                        anc.start,
+                        np.searchsorted(sites, left_focal_pos - max_length / 2),
+                    )
+                    insert_pos_end = np.minimum(
+                        anc.end,
+                        np.searchsorted(sites, right_focal_pos + max_length / 2),
+                    )
+                    logger.debug(f"Truncating ancestor {anc.id} at time {anc.time}")
```
jeromekelleher, line 2582: Be good to tell us how much we've chopped off in this message.

tests/test_formats.py
```diff
@@ -2201,6 +2205,57 @@ def test_insert_proxy_sample_oldest(self):
...
+
+    def test_length_multiplier(self):
+        sample_data, _ = self.get_example_data(10, 10, 40, use_sites_time=False)
+        ancestors = tsinfer.generate_ancestors(sample_data)
+        with pytest.raises(ValueError, match="Maximum length of haplotype"):
+            ancestors.truncate_ancestors(0.4, 0.6, 50)
+        trunc_anc = ancestors.truncate_ancestors(0.9, 1, 1)
+        assert np.array_equal(ancestors.ancestors_length, trunc_anc.ancestors_length)
+
+    def test_ancestors_truncated_length(self):
```
jeromekelleher, line 2230: Be good to test a few more things

- Do we leave the younger haplotypes alone?
- What happens when we specify extreme intervals, like min=0, max=1? Or such that there's exactly one haplotype in set we're truncating?
- Do we still round-trip the data exactly when we have this turned on for various parameters? We should probably add another TestRoundTrip subclass somewhere, so that we inherit all the weird examples from them.

jeromekelleher (commented):
Looks great, thanks @awohns! I have only nits to offer on the code. We do need more tests though, there's a large space of weird things that can happen and need to get more exposure of this code on them.

awohns:
I've written some more tests as per your recommendations @jeromekelleher and rewrote a bit of the documentation. Let me know what you think!

jeromekelleher (commented):
LGTM! Let's see how it performs on real data before merging, though.

awohns:
I think this is ready to merge, results on real data look good!

jeromekelleher (approved):

PR #473, opened 2021-02-28
check finalised

### LINKED ISSUE #472: Check for finalised SampleData files in merge
`merge` doesn't check that the sampledata files being merged are finalised. A silent error in creating a downstream SampleData file caused trouble after merging that file with others and trying to infer a tree sequence.
I can add a check that both SampleData files being merged are finalised in `merge`. 
Spoke to @hyanwong and he thinks it's a good idea. What do you think @jeromekelleher. It's a one line fix.

awohns:
Small change to check that `SampleData` files are finalised before merging them (with a test)

Closes #472

### DIFF
--- a/tests/test_formats.py
+++ b/tests/test_formats.py
@@ -1481,6 +1481,15 @@ class TestSampleDataMerge:
     Tests for the sample data merge operation.
     """
 
+    def test_finalised(self):
+        ts1 = tsutil.get_example_ts(2, 2, 1)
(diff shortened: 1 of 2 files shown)


awohns:
I guess the sandbox violation is the same as issue #1485 in msprime

jeromekelleher (approved):
LGTM, thanks. Yes, likely we're hitting the same issue as msprime with meson (although I thought we'd sorted it here)

PR #484, opened 2021-03-19
Account for change in tskit individuals.add_row

hyanwong:
As of tskit 0.3.5 there is now an extra param in `individuals.add_row`, so we need to switch to named parameters to get tests to pass.


jeromekelleher (approved):
LGTM

PR #485, opened 2021-03-19
Set root time to 1 if all times < 1

hyanwong:
It's confusing that, with the "time" of ancestors in 0..1, the root ancestor is placed at just under "time" 2, and the ultimate ancestor at ~ time 3. Here's a quick hack to make the root ancestor at time 1 if possible (normally we shouldn't have any ancestors at time 1, because that would imply the variant was fixed), and the ultimate ancestor at a fraction more than 1. This should help when plotting ancestor times, which can look a bit wacky otherwise.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,4 +1,23 @@
 ********************
+[0.2.2] - 2021-0X-XX
+********************
+
+**Bugfixes**:
+
+- Mutations at non-inference sites are now guaranteed to be fully parsimonious.
+  Previous versions required a mutation above the root when the input ancestral state
+  disagreed with the ancestral state produced by the parsimony algorithm. Now fixed by
+  using the new map_mutations code from tskit 0.3.7 (:pr:`557`, :user:`hyanwong`)
+
+**New Features**:
+
+**Breaking changes**:
+
+- Oldest nodes in a standard inferred tree sequence are no longer set to frequencies ~2
+  and ~3 (i.e. 2 or 3 times as old as all the other nodes), but are spaced above the
+  others by the mean time between unique ancestor ages (:pr:`485`, :user:`hyanwong`)
+
+********************
 [0.2.1] - 2021-05-26
 ********************
 
--- a/tsinfer/inference.py
+++ b/tsinfer/inference.py
@@ -946,8 +946,10 @@ class AncestorsGenerator:
             logger.info(f"Starting build for {self.num_ancestors} ancestors")
             progress = self.progress_monitor.get("ga_generate", self.num_ancestors)
             a = np.zeros(self.num_sites, dtype=np.int8)
-            root_time = max(self.timepoint_to_epoch.keys()) + 1
-            ultimate_ancestor_time = root_time + 1
+            root_time = max(self.timepoint_to_epoch.keys())
+            av_timestep = root_time / len(self.timepoint_to_epoch)
+            root_time += av_timestep  # Add a root a bit older than the oldest ancestor
+            ultimate_ancestor_time = root_time + av_timestep
             # Add the ultimate ancestor. This is an awkward hack really; we don't
             # ever insert this ancestor. The only reason to add it here is that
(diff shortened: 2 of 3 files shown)


tsinfer/inference.py
```diff
@@ -947,8 +947,12 @@ def run(self):
             logger.info(f"Starting build for {self.num_ancestors} ancestors")
             progress = self.progress_monitor.get("ga_generate", self.num_ancestors)
             a = np.zeros(self.num_sites, dtype=np.int8)
-            root_time = max(self.timepoint_to_epoch.keys()) + 1
-            ultimate_ancestor_time = root_time + 1
+            root_time = max(self.timepoint_to_epoch.keys())
+            if root_time < 1:
+                root_time = 1  # The normal case, when time==freq
+            else:
+                root_time += 1  # Could happen with user-specified times
```
jeromekelleher, line 954: I don't understand this comment - what could happen?

tsinfer/inference.py
```diff
@@ -947,8 +947,12 @@ def run(self):
...
+            ultimate_ancestor_time = root_time + 1 / self.num_ancestors
```
jeromekelleher, line 955: Why the new 1/N fraction?

jeromekelleher (commented):
Seems OK to me. Needs a bit more clarity in the comments and a couple of simple tests to check the behaviour.

tsinfer/inference.py
hyanwong: We could have root_time >=1 (because the root time could be any number)

tsinfer/inference.py
hyanwong: What do you suggest using instead? I could use root_time+1, as before, which would mean in the usual case most nodes would be at times `[0..1)`, apart from the root (at time 1), and the ultimate ancestor (at time 2). This creates weird plots with all node times 0..1 apart from one at time 2.

If we use `np.nextafter`, the augment_ancestors functionality fails, as it adds 1 to all the times. So `1 / self.num_ancestors` seemed like a reasonable epsilon.

jeromekelleher (commented):
LGTM, just need to udpate the CHANGELOG with this detail (or we'll forget it)

hyanwong:
> LGTM, just need to udpate the CHANGELOG with this detail (or we'll forget it)

Great point. Updated (and I added the change for map_mutations too)

CHANGELOG.rst
```diff
@@ -1,3 +1,21 @@
+********************
+[0.2.2] - 2021-0X-XX
+********************
+
+**Bugfixes**:
+
+- Placement of mutations at non-inference sites was not always truely parsimonious,
```
jeromekelleher, line 7: truly

CHANGELOG.rst
```diff
@@ -1,3 +1,21 @@
...
+  since the ancestral state could not be pre-specified. This is fixed by using the
+  new map_mutations code from tskit 0.3.7 (:pr:`557`, :user:`hyanwong`)
+
+**New Features**:
+
+- Oldest nodes in a standard inferred tree sequence are no longer set to frequencies ~2
```
jeromekelleher, line 13: This isn't a feature, it's a breaking change - people might be depending on the old behaviour.

CHANGELOG.rst
jeromekelleher: Maybe "fully parsimonious" would be a better description. Also this won't make sense to an outsider: "Mutations at non-inference sites are now guaranteed to be fully parsimonious. Previous versions required a mutation above the root when the input ancestral state disagreed with the ancestral state produced by the parsimony algorithm."

jeromekelleher (commented):
Needs a bit of a rephrase - the CHANGELOG is for users who don't closely follow development to read and be informed of the important changes.

CHANGELOG.rst
hyanwong: Yep, sounds good. I was probably dozy when I wrote it.

jeromekelleher (approved):

Issue #147, opened 2019-04-15
Extra parallelization possibilities for match_ancestors?


hyanwong:
At the moment we are finding that the slow stage of inference is `match_ancestors`, since it is only parallelized within an epoch (for ancestors at exactly the same age). This is because we need to make sure that ancestors can copy off all older ancestors if necessary.

As mentioned to @awohns, I think we could, however, parallelize ancestor matching as long as the ancestors we are processing in parallel do not cover the same region of genome. In other words, if we keep a list, `L`, of  the `start` and `end` values of the ancestors we are currently matching, I think we can process the next-ancestor-in-time as long as its start and end values don't intersect with the segments in `L`.

Is my thinking correct? If so, I don't know how much speedup we would get, but I suspect that a long chromosome (when ancestors are short) might be substantially parallelizable,

jeromekelleher:
This should work, but will require a bit of replumbing. While we're at it, it might be better to grasp the nettle and build a full interval tree of the ancestors and figure out what could possibly overlap with what. It should be possible to build up a work-list based on that. I'm sure there'll be tricky technical details along the way though, so it's not an easy win.

hyanwong:
@jeromekelleher and I just chatted about some possibilities for this issue, now that we can have [smaller old ancestors](https://github.com/tskit-dev/tsinfer/pull/467).

The problem can be thought of as a dependancy graph, where ancestors are dependent on older ancestors that (at least partially) overlap, and those ancestors are dependent on older ancestors, etc.

Maybe this is easier than we think. The dependency graph is, I think, trivial to extract:

```
ts = msprime.sim_mutations(
    msprime.sim_ancestry(
        1000, population_size=1e4,
        recombination_rate=1e-8,
        sequence_length=1e5, random_seed=1),
    rate=1e-8, random_seed=1)

sd = tsinfer.SampleData.from_tree_sequence(ts, use_sites_time=False)
ancestors = tsinfer.generate_ancestors(sd)

start = ancestors.ancestors_start[:]
end = ancestors.ancestors_end[:]
atime = ancestors.ancestors_time[:]

for a in ancestors.ancestors():
    older_dependencies = np.where(np.logical_and.reduce((start < a.end, end > a.start, atime > a.time)))[0]
    print(f"{a.id} depends on {older_dependencies}")
```

[...]

PR #486, opened 2021-03-19
Use dependency levels rather than epochs

hyanwong:
OMG @jeromekelleher . I think this might actually work to address #147 . It's only failing on 3 tests, all of them when doing `self.tree_sequence_builder.add_path()`, with the following failure

```
_tsinfer.LibraryError: Bad path information: time
```

I suspect that's because `add_path()` is checking that stuff is added in time order, but it probably doesn't need to? Perhaps you could take a quick look to confirm. Then I can try perf testing this!!!

### DIFF
--- a/tsinfer/formats.py
+++ b/tsinfer/formats.py
@@ -2546,16 +2546,20 @@ class AncestorData(DataContainer):
     # Read mode
     ####################################
 
-    def ancestors(self):
+    def ancestors(self, indexes=None):
         """
-        Returns an iterator over all the ancestors.
+        Returns an iterator over all the ancestors. If indexes is provided, it should
+        be a sorted list of indexes giving a subset of ancestors to return.
+        For efficiency, the indexes should be a numpy integer array.
         """
-        # TODO document properly.
         start = self.ancestors_start[:]
         end = self.ancestors_end[:]
         time = self.ancestors_time[:]
         focal_sites = self.ancestors_focal_sites[:]
-        for j, h in enumerate(chunk_iterator(self.ancestors_haplotype)):
+        haplotypes = chunk_iterator(self.ancestors_haplotype, indexes)
+        if indexes is None:
+            indexes = range(len(time))
+        for j, h in zip(indexes, haplotypes):
             yield Ancestor(
                 id=j,
                 start=start[j],
--- a/tsinfer/inference.py
+++ b/tsinfer/inference.py
@@ -853,11 +853,10 @@ class AncestorsGenerator:
             start, end = self.ancestor_builder.make_ancestor(focal_sites, a)
             duration = time.perf_counter() - before
             logger.debug(
-                "Made ancestor in {:.2f}s at timepoint {} (epoch {}) "
+                "Made ancestor in {:.2f}s at timepoint {} "
                 "from {} to {} (len={}) with {} focal sites ({})".format(
                     duration,
                     t,
-                    self.timepoint_to_epoch[t],
                     start,
                     end,
                     end - start,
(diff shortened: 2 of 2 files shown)


hyanwong:
NB: at the moment I'm saving the ancestor ids for each level in a normal python dict: `{level1:[ancestor_id1, ancestor_id2,..], level1:[ancestor_id3, ancestor_id4,ancestor_id5..], ...}`. But if we have millions of ancestors this might be a bit of a burden, so we could possibly store them e.g. in a zarr array instead, or even re-calculate them on the fly from the intermediate numpy array which I've called `dep_level`.

jeromekelleher (commented):
I'd need some time to digest and process this @hyanwong. In the mean time, how about trying stuff out with just the Python engine? This should be easier to understand and see if there's any assumptions about time in there.

hyanwong:
The reason that it's raising an error (e.g. on test_inference.py::TestMetadataRoundTrip::test_from_historical_tree_sequence) is a logic bug (damn), which means that it's possible to get overlapping ancestors that have the same time value, but are at different dependency levels. That means one can become the ancestor of another, but this is disallowed if they have the same time value.

Unless there is a way to restrict LS matching to a subset of the reference panel, which I understand is quite tricky, I think we'll have to solve this by "boosting" overlapping individuals within the same timeslice to the highest dependency level. In the extreme case, with all ancestors at a give time overlapping, this would revert to something like the old scheme. But I don't think we'll be in that situation. Probably worth re-testing with a newer algorithm, though.

hyanwong:
I've added the simple approach mentioned above (where we take all ancestors in the same time slice and set their dependency level to be the maximum of any ancestor in the time slice) and now all tests pass - woohoo. I worry slightly that this will adversely affect the ability to parallelise - I'm just testing it now.

As a follow up, here's a few suggestions for improving the current solution. The problem is not *too* bad because I believe this issue only applies to ancestors that have exactly the same time value (which means that the "dependency level" approach should still be effective in solving the problem of multiple similar ancestor times caused by missing data).

Perhaps the most sophisticated approach is to identify ancestors that are logically independent across positions - i.e. use some sort of interval library to identify groups of ancestors whose joint intervals do not overlap. Within each group, we can then set the dependency level to the maximum in the group. However, I think this will be computationally intensive.

Another approach (which could be combined with the above), would be to allow the *construction* of ancestors to happen at their initially calculated dependency level, but not actually *add* the ancestors to the growing tree sequence until all the ancestors in a given level have been built. For example, each ancestor could have a "dependency level (build)" and a "dependency level (add)". We build the ancestors during their "build" level, but don't add them until their "add" level has been reached. This would, however, involve buffering the ancestors somewhere, so is an additional level of complication.

hyanwong:
Here's the result of the new code on a simulation of 10,000 samples for 10Mb. Looks like it is still working OK. I'm running it on a larger example now - although actually working out the dependencies takes enough time that it seems to be a bit slower in total. I suspect this can be avoided with some more efficient code. It would also be worth trying out on the HGDP (missing data) examples, as I suspect this is where the approach will show major benefits (versus the unbinned ancestors).

Screenshot 2021-03-23 at 12 21 17

hyanwong:
Here's the result from a simulation an order of magnitude larger, with 300,000 sites.

Screenshot 2021-03-23 at 12 55 45

And the result for Wilder's unified dataset, chr20.

Screenshot 2021-03-23 at 14 08 17


The main problem is that finding the dependency levels is slow. I think that can be sped up considerably using an interval library such as https://github.com/biocore-ntnu/ncls, although there are some details to be worked out (e.g. ideally we would gradually build up the interval list as we go down the ancestors, whereas the docs for that library seem to imply that the list is constructed in a single pass.

jeromekelleher:
Thanks @hyanwong. Let's park this here for now. I'll need to get into the weeds on this, and have other things that need finishing up first.

hyanwong:
> Thanks @hyanwong. Let's park this here for now. I'll need to get into the weeds on this, and have other things that need finishing up first.

Yes, I agree. I feel like I've done as much as necessary on this, have shown it works, and identified some issues. That's enough for the moment.

benjeffery:
@hyanwong I've rebased to main, shall I push here or a new PR?
I think there are some quick perf wins here (e.g. not using the `.ancestors` iterator when you only need the times) so I'm just working on those.

hyanwong:
> @hyanwong I've rebased to main, shall I push here or a new PR?

As you like. If a new PR then you're a bit more in control?

> I think there are some quick perf wins here (e.g. not using the `.ancestors` iterator when you only need the times) so I'm just working on those.

Good point, thanks. I didn't try to optimise this in any way, so not surprised there are several wins.

hyanwong:
Improved by @benjeffery 's reworking, so no longer relevant.

PR #333, opened 2020-08-22
New msprime calling convention

hyanwong:

PR #492, opened 2021-04-22
Allow non-ancestral alleles to be in any order in verify

### LINKED ISSUE #490: Alleles not equal on tsinfer verify, v 0.2.0
Hi,
I am working with the tsinfer version 0.2.0. When running verify, I get an error about alleles not being equal. I checked that all my sites have only 2 alleles and that the ancestral allele matches either the REF or ALT. I also removed all singletons, not sure if this was advised or not. I used this script (https://github.com/stsmall/Kiribina_Folonzo/blob/debug/vcf2tsinfer.py) which was a modification of the one in the tutorial. tsinfer ran to completion without error.
thanks for the help!
@stsmall

> $ tsinfer list chrX.trees
path      = chrX.trees
size      = 519.0 MiB
edges     = 9215138
trees     = 910682
sites     = 2160904
mutations = 2124826

> $ tsinfer verify chrX.samples chrX.trees
2021-04-22 10:36:02,999 [47534] CRITICAL root: Traceback (most recent call last):
  File "tsinfer", line 8, in <module>
    sys.exit(main())
  File "tsinfer/__main__.py", line 5, in main
    cli.tsinfer_main()
  File "tsinfer/cli.py", line 499, in tsinfer_main
    args.runner(args)
  File "tsinfer/cli.py", line 238, in run_verify
    tsinfer.verify(samples, ts, progress_monitor=args.progress)
  File "tsinfer/inference.py", line 181, in verify
    raise ValueError(f"alleles not equal: {var1.alleles} != {var2.alleles}")
ValueError: alleles not equal: ('C', 'T') != ('C',)

hyanwong:
Also ignored if not in geno array.

Fixes #490, but @jeromekelleher might not like my use of a "sentinel" value (if that's the right word) of `-2` to check for cases where we might have a value that's not in the mapping. But we can't use `tskit.NULL==-1` for this, so other approaches welcome.

### DIFF
--- a/tests/test_inference.py
+++ b/tests/test_inference.py
@@ -2863,6 +2863,21 @@ class TestVerify:
         with pytest.raises(ValueError):
             tsinfer.verify(samples, ts)
 
+    def test_bad_ancestral_allele(self):
+        n = 2
+        ts = msprime.simulate(n, mutation_rate=5, random_seed=1)
+        assert ts.num_sites > 1
+        with tsinfer.SampleData(sequence_length=ts.sequence_length) as samples:
+            for var in ts.variants():
+                samples.add_site(
+                    position=var.site.position,
+                    alleles=["1", "0"],
+                    genotypes=var.genotypes,
+                )
+
+        with pytest.raises(ValueError, match="Ancestral"):
+            tsinfer.verify(samples, ts)
+
     def test_bad_alleles(self):
         n = 2
         ts = msprime.simulate(n, mutation_rate=5, random_seed=1)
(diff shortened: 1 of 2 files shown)


jeromekelleher:
I think I understand the basic approach, but I'm not following the details of how the mapping is done. A simpler (but slower) approach would be to generate the list of strings for each, and compare those. So, something like

```
if alleles1 != alleles2:
   # revert to slow method
   g1 = [alleles1[allele_index for allele_index in genotypes1]
   g2 = [alleles2[allele_index for allele_index in genotypes2]
   assert g1 == g2
else:
   assert np.all(genotypes1 == genotypes2)
```
It's slow, but it's easy to follow and it's not really critical path code, so I think I'd favour simplicity?

hyanwong:
> It's slow, but it's easy to follow and it's not really critical path code, so I think I'd favour simplicity?

I wasn't sure if this would be used for really large inference examples. But if, as you say, it's not a critical path, then I agree. Simpler is better. Shall I switch for your version?

We also need to merge https://github.com/tskit-dev/tsinfer/pull/333 or equivalent to get the other tests to pass with msprime 1.0.0

jeromekelleher (approved):
LGTM

benjeffery:
@mergifyio rebase

PR #516, opened 2021-05-19
Update tutorial.rst

minesrebollo:
The interval in the text didn't match the one in the diagram and ather small text change.


hyanwong:
That's great. Thanks so much for spotting and correcting this @minesrebollo . @benjeffery - could you merge this, as Jerome is away?

At some point soon I'll re-jig the examples to use msprime v1.0.0 calls, and a discrete genome, which will change things again, I suspect. But best to make these corrections now.

PR #533, opened 2021-06-07
Remove .egg dir on make clean

hyanwong:
It seems as if when I now compile tsinfer it puts stuff in the .egg dir, and then I have to clear it out to remake. I *think* adding it to `make clean` is therefore correct: is that right @benjeffery ?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,3 +15,4 @@ ctags:
 clean:
 	rm -f *.so *.o tags
 	rm -fR build
+	rm -fR .eggs


benjeffery:
Why are you needing to delete `.eggs` to remake? That directory is a cache of downloads for setuptools. `tsinfer`  should build the same if it is present or not.

hyanwong:
I dunno, but was getting an error when trying to re-`make` stuff using different compiler setups. Not sure I can replicate it now, though, annoyingly.

benjeffery:
Ok, thanks for checking, re-open if you hit it again.

PR #537, opened 2021-06-16
Fixes https://github.com/tskit-dev/tsinfer/discussions/523#discussion…

hyanwong:
…comment-880019

I'm pretty sure this is a bug in the example VCF reading code and should be corrected ASAP. @awohns - you presumably didn't use this code for reading in any important VCFs did you: it looks to me as if we might be accidentally flipping allelic states when the ancestral allele is not the ref. Erk. Could you check that (a) the fix is the correct one and (b) it's not caused any cock-ups for our analyses?

Sorry about that.

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -397,7 +397,7 @@ variants from chromosome 24 of ten Norwegian and French house sparrows,
                 for row in variant.genotypes
                 for old_index in row[0:2]
             ]
-            samples.add_site(pos, genotypes=genotypes, alleles=alleles)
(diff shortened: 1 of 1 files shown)


awohns:
Good catch @hyanwong! I didn't use this code at all so worries on my end

jeromekelleher (approved):

hyanwong:
@benjeffery - not sure why this is failing. Could we perhaps merge https://github.com/tskit-dev/tsinfer/pull/556 then try this one again?

benjeffery:
@mergifyio rebase

PR #558, opened 2021-07-04
Place provenance recording in top level functions

hyanwong:
Like the other functions in Tskit, I think we should set provenance in the top level functions such as tsinfer.infer, or tsinfer.match_ancestors. This pulls out the provenance stuff from within lower level functions and puts it in the top functions. Hopefully this will make it easier to include more parameters etc in the provenance.

### DIFF
--- a/tests/test_provenance.py
+++ b/tests/test_provenance.py
@@ -68,6 +68,77 @@ class TestProvenanceValid:
         self.validate_file(ancestor_data)
 
 
+class TestIncludeProvenance:
+    """
+    Test that we can include or exclude provenances
+    """
+
+    def test_no_provenance_infer(self, small_sd_fixture):
+        ts = tsinfer.infer(small_sd_fixture, record_provenance=False)
+        assert ts.num_provenances == small_sd_fixture.num_provenances
+
+    def test_no_provenance_generate_ancestors(self, small_sd_fixture):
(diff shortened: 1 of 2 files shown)


jeromekelleher:
Do we have an application for this? I feel like a lot more time has been spent on provenance that has actually been gained. Unless there's an immediate problem we're trying to solve I'd rather just leave it alone.

hyanwong:
It would be pretty useful to me (and presumably others) to be able to know from the tree sequence the mismatch rates and recombination rates used for inference. That's the main thing in my book, and I'm assuming it would be pretty easy to add those to the provenance.

jeromekelleher:
And they are not being recorded in the lower levels?

hyanwong:
No, there's no way to get them back out of the tree sequence, AFAIK.

hyanwong:
Agreed. I would like to keep it active for after 0.3.0 though.

hyanwong:
I've revised my opinion on this @jeromekelleher - I don't think this is controversial, is it? We aren't recording anything more in the provenances here, simply doing what we do in tskit. It confused me for a while when looking at the CLI injected provenances that all the match_ancestor etc stuff was in there too, so a way to take it out (assuming we want to do that) when calling via the CLI would be helpful, I think.

jeromekelleher (approved):
I see, yes, this is helpful.

PR #570, opened 2021-08-04
Add .ancestor(n) method

### LINKED ISSUE #569: Provide ancestor() accessor method
I'm back to messing with tsinfer, and realised that we can't do e.g.

```
target_ancestor = ancestor_data.ancestor(2)
```

Which would be quite useful in a number of circumstances. I guess there's no reason why not to provide it. It's not something that sgkit will render irrelevant.

hyanwong:
And allow __eq__ for ancestors (which contain numpy arrays)

Fixes #569

### DIFF
--- a/tests/test_formats.py
+++ b/tests/test_formats.py
@@ -2056,6 +2056,39 @@ class TestAncestorData(DataContainerMixin):
                 start=0, end=num_sites, time=2, focal_sites=[], haplotype=haplotype
             )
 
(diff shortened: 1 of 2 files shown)


hyanwong:
It would be nice to merge this. Nothing controversial here, I think.

hyanwong:
> LGTM
I guess we can flag this for automerging then

PR #574, opened 2021-08-23
Add note about validity of node times & tsdate

hyanwong:


jeromekelleher (approved):
LGTM. We should probably put something similar in the GitHub README too so it's unmissable.

hyanwong:
> LGTM. We should probably put something similar in the GitHub README too so it's unmissable.

Shall I add this to this PR? Also I see there's a README.rst which I assume is obsolete - I could remove that at the same time?

jeromekelleher:
Please do. I'd leave README.rst alone though, I'm sure this is involved in the Python packaging (it's the front page on PyPI AFAIK)

hyanwong:
> Please do. I'd leave README.rst alone though, I'm sure this is involved in the Python packaging (it's the front page on PyPI AFAIK)

Ah, I didn't realise that. Probably worth putting a comment in the rst file to that effect?

PR #599, opened 2021-11-02
Change time defaults for from_tree_sequence

hyanwong:
As in discussions with @awohns and @jeromekelleher. Note that we now error out, saying "Incompatible timescales" if we are using times as freqs, but have historical samples. The workaround if a user really does want to do this (which is unlikely, I think: it would imply that the historical samples were placed at times equivalent to the frequencies of alleles) is to set the site times manually.

### DIFF
--- a/tsinfer/formats.py
+++ b/tsinfer/formats.py
@@ -1419,8 +1419,8 @@ class SampleData(DataContainer):
     def from_tree_sequence(
         cls,
         ts,
-        use_sites_time=True,
-        use_individuals_time=True,
+        use_sites_time=None,
+        use_individuals_time=None,
         **kwargs,
     ):
         """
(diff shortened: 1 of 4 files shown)


tsinfer/formats.py
```diff
@@ -1419,8 +1419,8 @@ def min_site_times(self, individuals_only=False):
     def from_tree_sequence(
         cls,
         ts,
-        use_sites_time=True,
-        use_individuals_time=True,
+        use_sites_time=False,
```
jeromekelleher, line 1422: Let's change these to None while we're breaking things. We may want to have different behaviour sometime in the future and hard-coding the defaults in the signature makes things very inflexible.

jeromekelleher (approved):
LGTM, but lets not embed the defaults in the signature

tsinfer/formats.py
hyanwong: Good call.

jeromekelleher (approved):

PR #612, opened 2021-11-22
Calculate max size of memory requested

### LINKED ISSUE #489: Allocate larger blocks for ancestor builder
It may happen that the number of sites used to finalise ancestor building is so large that [tsk_blkalloc_init](https://github.com/tskit-dev/tsinfer/blob/main/lib/ancestor_builder.c#L145) allocates too little memory in [ancestor_builder_alloc](https://github.com/tskit-dev/tsinfer/blob/main/lib/ancestor_builder.c#L124). cf discussion #488 ; the bug manifested when some [370k sites](https://github.com/tskit-dev/tsinfer/discussions/488#discussioncomment-588565) were used to finalise ancestor building, which resulted in [focal_sites](https://github.com/tskit-dev/tsinfer/blob/main/lib/ancestor_builder.c#L534) being assigned a NULL value.

### LINKED ISSUE #600: MemoryError in tsinfer.infer()
I am trying to infer tree sequence from the sample data file for each chromosome separately for a plant species using tsinfer.infer()  for 286 individuals. This works fine for the shortest chromosome but fails for all the other chromosomes with 'MemoryError' at 'tsinfer/inference.py", line 949, in run
    self.descriptors = self.ancestor_builder.ancestor_descriptors()'
I doubt it is the RAM causing this issue because until this step the maximum RAM it consumes is less than a GB and I run it on an HPC with more than 700 GB RAM unshared with others. I observe that the number of threads evoked is more than 150. I already tried setting the num_threads to 0. Is the number of threads evoked causing this issue? Also, how can I better control the number of threads evoked within the tool?

hyanwong:
Fixes #489, fixes #600. Not sure how we test this easily without running a huge inference.

### DIFF
--- a/lib/ancestor_builder.c
+++ b/lib/ancestor_builder.c
@@ -125,6 +125,7 @@ ancestor_builder_alloc(
     ancestor_builder_t *self, size_t num_samples, size_t max_sites, int flags)
 {
     int ret = 0;
+    unsigned long max_size = 1024 * 1024;
 
     memset(self, 0, sizeof(ancestor_builder_t));
     if (num_samples <= 1) {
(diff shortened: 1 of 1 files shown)


jeromekelleher:
Thinking this through: suppose we have 10^6 samples and 10^8 sites. The samples bit would still be less than a megabyte, so that's all good. The ``sites`` bit would be 380 MiB which is quite a lot, but I guess not outrageous if you're working with 100 million sites (which is a *lot*).

OK, LGTM, thanks @hyanwong!

jeromekelleher (approved):

PR #642, opened 2022-03-05
Use struct for node metadata

### LINKED ISSUE #416: Add tsinfer specific schema for node metadata.
We can add a Struct schema to the NodeTable which will let us efficiently store information about the inference process. Currently, this should contain "ancestor_data_id" (default = -1) and "sample_data_id" (default=-1). I made a start on doing this, but it turned out to be quite messy, mainly due to the lack of default value handling in tskit metadata.

Once these two issues are resolved we have another look at this:

https://github.com/tskit-dev/tskit/issues/1073
https://github.com/tskit-dev/tskit/issues/1084

It's not clear whether the struct schema of JSON would end up being most efficient here, but we should start with struct anyway just to  see how it works out.

This is purely an interface for tsinfer to encode information about the nodes - tsinfer users will not have the option of setting or updating the schema.

Follow on from #392.

hyanwong:
Fixes #416

### DIFF
--- a/tsinfer/formats.py
+++ b/tsinfer/formats.py
@@ -74,6 +74,42 @@ def permissive_json_schema():
     }
 
 
+def node_metadata_schema():
+    # This is fixed by tsinfer: users cannot add to the node metadata
+    return tskit.MetadataSchema(
+        {
+            "codec": "struct",
+            "type": ["object", "null"],
+            "properties": {
+                "tsinfer": {
+                    "description": "Information about node identity "
+                    "from the tsinfer inference process",
+                    "type": "object",
+                    "properties": {
+                        "ancestor_data_id": {
+                            "description": "The corresponding ancestor ID "
+                            "in the ancestors file created by the inference process, "
+                            "or -1 if not applicable",
+                            "type": "number",
+                            "binaryFormat": "i",
+                            "default": -1,
+                        },
+                        "sample_data_id": {
+                            "description": "The corresponding sample ID "
+                            "in the sample data file used for inference, "
+                            "or -1 if not applicable",
+                            "type": "number",
+                            "binaryFormat": "i",
+                            "default": -1,
+                        },
+                    },
+                },
+            },
+            "additionalProperties": False,
+        }
+    )
+
+
 def np_obj_equal(np_obj_array1, np_obj_array2):
     """
     A replacement for np.array_equal to test equality of numpy arrays that
--- a/tests/test_inference.py
+++ b/tests/test_inference.py
@@ -1094,9 +1094,14 @@ class TestMetadataRoundTrip:
             assert i1.flags == i2.flags
             assert tsutil.json_metadata_is_subset(i1.metadata, i2.metadata)
         # Unless inference is perfect, internal nodes may differ, but sample nodes
-        # should be identical
+        # should be identical. Node metadata is not transferred, however, and a tsinfer-
+        # specific node metadata schema is used (where empty is None rather than b"")
+        assert (
+            ts_inferred.table_metadata_schemas.node
+            == tsinfer.formats.node_metadata_schema()
+        )
         for n1, n2 in zip(ts.samples(), ts_inferred.samples()):
-            assert ts.node(n1) == ts_inferred.node(n2)
+            assert ts.node(n1).replace(metadata=None) == ts_inferred.node(n2)
         # Sites can have metadata added by the inference process, but inferred site
         # metadata should always include all the metadata in the original ts
         for s1, s2 in zip(ts.sites(), ts_inferred.sites()):
@@ -1586,12 +1591,13 @@ class TestAncestorsTreeSequence:
             ancestors_time = ancestor_data.ancestors_time[:]
             num_ancestor_nodes = 0
             for n in ancestors_ts.nodes():
-                md = json.loads(n.metadata) if n.metadata else {}
+                md = n.metadata if n.metadata else {}
                 if tsinfer.is_pc_ancestor(n.flags):
-                    assert not ("ancestor_data_id" in md)
+                    if "tsinfer" in md:
+                        assert "ancestor_data_id" not in md["tsinfer"]
                 else:
-                    assert "ancestor_data_id" in md
-                    assert ancestors_time[md["ancestor_data_id"]] == n.time
+                    assert "tsinfer" in md and "ancestor_data_id" in md["tsinfer"]
+                    assert ancestors_time[md["tsinfer"]["ancestor_data_id"]] == n.time
                     num_ancestor_nodes += 1
             assert num_ancestor_nodes == ancestor_data.num_ancestors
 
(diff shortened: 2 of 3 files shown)


hyanwong:
This has the new metadata scheme, where the tsinfer-specific metadata is in a "tsinfer" key.

hyanwong:
This is ready for review. I reckon if we merge it (or whatever we end up with) at the same time as #607 then we make 2 changes to the formats at roughly the same time.

tsinfer/formats.py
```diff
@@ -74,6 +74,43 @@ def permissive_json_schema():
...
+    return tskit.MetadataSchema(
+        {
+            "codec": "struct",
+            "type": ["object", "null"],
+            "properties": {
+                "tsinfer": {
+                    "description": "Information about node identity "
+                    "from the tsinfer inference process",
+                    "type": "object",
+                    "default": {"ancestor_data_id": -1, "sample_data_id": -1},
```
jeromekelleher, line 88: What's this default for? I'm not sure we can do this.

tsinfer/formats.py
```diff
@@ -74,6 +74,43 @@ def permissive_json_schema():
...
 
 
+def node_metadata_schema():
+    # This is fixed by tsinfer: users cannot add to the node metadata
+    return tskit.MetadataSchema(
+        {
+            "codec": "struct",
+            "type": ["object", "null"],
+            "properties": {
+                "tsinfer": {
```
jeromekelleher, line 84: I don't see the point in the "tsinfer" top key here.

jeromekelleher (commented):
Looks good, but I'm not sure about the schema. I think we're over complicating it.

The metadata column should be interpretable as a 2D numpy array of int32s, so all the values are explicitly filled in as (-1,-1). There's no defaults then. I think this is much more useful in practice than using nulls and saving a few bytes.

tsinfer/formats.py
hyanwong: I thought it was cleaner, because then when we e.g. add values inferred from `tsdate` we can store them in a `tsdate` property. That way it is clear where the different metadata values have come from, and they don't really risk overwriting each other etc.

tsinfer/formats.py
hyanwong: This is needed because if a top level object is not marked as "required", it needs a top level default. This does work, and I've checked it with @benjeffery, although perhaps he could verify?

jeromekelleher:
As I say, the main point of using struct here for me would be to enable direct access to the metadata as numpy arrays. Let's get some info on doing that first before discussing the other details any further.

hyanwong:
Ah, that's not the main reason I would be using a struct, TBH. I didn't even think about wanting to get the entire array out.

BTW, and not that it matters, but the `sample_data_id` is mostly unfilled, I think.

jeromekelleher:
If the info is sparse and you mainly plan on leaving things blank then there's no point at all in struct, I think. But being able to process these things efficiently in Python seems attractive to me.

hyanwong:
The "sample_data_id" value is sparse (I think). The "ancestor_data_id" is not, and storing 27 million copies of the string "ancestor_data_id" in the public tree sequences seems pretty wasteful to me.

tsinfer/formats.py
hyanwong: Addendum: I see this as a nice advantage of `struct`: we can be verbose in the explanation (and nesting) of the properties in the metadata, and it doesn't take up extra space in the encoding of the metadata for each row. That's my understanding, anyway.

jeromekelleher:
Maybe some concrete numbers for chr1 would be helpful.

hyanwong:
OK, for the p-arm of chr1, which is 123.4 MB (i.e. about 1/25th of the whole), we are storing 30Mb in node metadata, of which we have just over a million copies of the byte string "ancestor_data_id" (= 16 Mb), and no copies of the string "sample_data_id" (because we only use this for historically inserted samples, which we don't do in the unified ts).

jeromekelleher:
So what would that be for 8 bytes per node?Around 8MB I guess?

tsinfer/formats.py
benjeffery: Thinking about it, I'm not sure this is what you want. For the lat/lng case it made sense - doing the default in this way says "Either both properties are the default or both are specified". For this case I think you want them to be independent, and the default should be specified separately at the deeper level.

tsinfer/formats.py
benjeffery: The nice thing is that there is no storage cost to the additional level.

hyanwong: > The nice thing is that there is no storage cost to the additional level.

Yes, that's what I was trying to say (but you said it better)

tsinfer/formats.py
hyanwong: I just thought this through and indeed we can delete the higher-level default, I think. I've updated the PR.

benjeffery: Yes, the proprty becomes required - but it can just be an empty object.

hyanwong:
> So what would that be for 8 bytes per node?Around 8MB I guess?

Yep, 8.9MB. Note that the released unified genealogy tree sequences also don't contain the means and variances associated with node times (I'm not sure why they don't). If we included those it would add json info such as

```
b'"mn":11039.99, "vr":1218813.29'
```

to the JSON, which I estimate would multiply the metadata storage requirements in the JSON format by ~ 2.5 times, and would either increase the struct version by another 8.9 MB if we store mean and var as 4 byte floats, or 17.8 if we use doubles.

hyanwong:
So if we store tsdate mean and var, for all chromosomes in total, we use up roughly 1.9 GB in a JSON encoding, and either 445 MB (if storing floats) or 667MB (if storing doubles) for the struct encoding.

If we don't store tsdate means and variances, we use up ~ 750MB for JSON and 220MB for struct.

benjeffery:
I've come and gone several times on the "top-level key with name of generating software" concept several times, but I think I've convinced myself that it is a bad idea. Here's the thinking - if downstream software wants to use a metric stored in metadata it will have to know what software wrote it, likely hard-coding the name of that software. This would prevent a different upstream software from replacing the first, or even worse cause it to write metrics to a key that had the name of the first piece of software in order to be compatible! 
In other words the identity of the metric is it's key, it's name, and the generating software is incidental  (but should absolutely be stored in the description of the field). 
Tagging @hyanwong and @jeromekelleher as I think this is something we need to decide for the ecosystem and put into the tskit docs at some point.

hyanwong:
Thanks @benjeffery. I guess there might be a difference between metadata that is generalisable, and metadata that is very tightly linked to the software package. In the case of the `sample_data_id` and `ancestor_data_id` from _tsinfer_, these values reference specific internal IDs created by the _tsinfer_ inference process, so I can't imagine they would be useful outside that context.

Having said which, I'm perfectly happy to label them simply `sample_data_id` and documenting their origin in the `description` field (or even use `tsinfer_sample_data_id`, although that slightly ruins the justification of not hard-coding the software name)!

benjeffery:
Yes, good point. What downstream uses are there of `sample_data_id` and `ancestor_data_id`? Is it just for analysis of `tsinfer`'s performance?

hyanwong:
Some of the rationale is at https://github.com/tskit-dev/tsinfer/issues/301 and https://github.com/tskit-dev/tsinfer/issues/604#issuecomment-968979060

Basically, it is useful for manipulating the tree sequences for matching historical individuals into a tree sequence, and  inserting them etc. This could either be done in the standard internal tsinfer routines, or as part of a post-processing step

hyanwong:
Closing as we decided simply to use permissive json, at least for the time being

PR #644, opened 2022-03-07
Add fill_value to object arrays args to bypass zarr issue

benjeffery:
Zarr will fix the indexing issue https://github.com/zarr-developers/zarr-python/pull/974 on next release, but not the `fill_value` issue. I think the best way forward is to specify fill values as in this PR then blocklist the bad zarr version for indexing once we unpin on the next zarr release.

### DIFF
--- a/tsinfer/formats.py
+++ b/tsinfer/formats.py
@@ -406,6 +406,7 @@ class DataContainer:
             compressor=self._compressor,
             dtype=object,
             object_codec=self._metadata_codec,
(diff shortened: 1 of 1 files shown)


hyanwong:
There's no harm specifying the fill value here anyway, right? So we should just merge this now.

jeromekelleher (commented):
LGTM!

jeromekelleher (approved):

Issue #647, opened 2022-03-09
Library errors encountered when using make_ancestors_ts()


szhan:
I'm trying to make an ancestor tree sequence out of a standard tree sequence generated using msprime using the unexposed function make_ancestors_ts(), which is part of eval_util.py.

I'm trying out the function using the following block of code:

```
import tsinfer
from tsinfer import make_ancestors_ts
import tskit
import msprime

print(f"tsinfer {tsinfer.__version__}")
print(f"tskit {tskit.__version__}")
print(f"msprime {msprime.__version__}")

rate_map = msprime.RateMap.uniform(
    sequence_length = 1_000_000,
    rate = 1e-8
)

ts = msprime.sim_ancestry(
    samples = 1_000,
    population_size = 10_000,
    ploidy = 2,
    model = "hudson",
    recombination_rate = rate_map
)

mts = msprime.sim_mutations(
    ts,
    rate = 1e-8
)

ts_anc = make_ancestors_ts(samples = mts.samples(), ts = mts, remove_leaves = True)
```

I've run the block of code multiple times. Sometimes it works, but often it leads to two errors (usually the first):
1) LibraryError: Specified parent mutation is at a different site.
2) LibraryError: Mutation out of bounds.

jeromekelleher:
Huh - I *thought* the problem was that we weren't updating mutation parents, but it looks like we are:

 https://github.com/tskit-dev/tsinfer/blob/6ed592f62d3c92c00eb918ef2bcb50c8c8171bad/tsinfer/eval_util.py#L481]

@szhan can you make a reproducible example here please by setting the random seeds, and pasting in the stack trace?

szhan:
This pair of random seeds led to the stack trace below:
(1) sim_ancestry - 354053995
(2) sim_mutations - 4054152783

```
---------------------------------------------------------------------------
LibraryError                              Traceback (most recent call last)
Input In [22], in <module>
----> 1 ts_anc = make_ancestors_ts(samples = mts.samples(), ts = mts, remove_leaves = True)

File ~/opt/anaconda3/envs/tskit-play/lib/python3.8/site-packages/tsinfer/eval_util.py:458, in make_ancestors_ts(samples, ts, remove_leaves)
    456     if np.sum(var.genotypes) > 1:
    457         position.append(var.site.position)
--> 458 reduced = subset_sites(ts, position)
    459 minimised = inference.minimise(reduced)
    461 tables = minimised.dump_tables()

[...]

PR #648, opened 2022-03-09
Add support for multiple mutations

jeromekelleher:
I fixed the initial issue causing #647, but it's after opening up another can of worms with the evaluations.

Looks like we should just try to limit ourselves to two alleles when using tsinfer? @hyanwong, do you remember where this was at?

### DIFF
--- a/tests/test_evaluation.py
+++ b/tests/test_evaluation.py
@@ -1,5 +1,5 @@
 #
-# Copyright (C) 2018-2020 University of Oxford
+# Copyright (C) 2018-2022 University of Oxford
 #
 # This file is part of tsinfer.
 #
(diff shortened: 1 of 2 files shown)


hyanwong:
When there are >2 alleles we simply place the mutations using parsimony, and don't use them for inference. So normally, it's fine using triallelic alleles.

We should advise people that if they really need to use them for inference, they *can* use tri- and tetra-allelic sites by picking some of the alleles and setting the samples that have those alleles to "missing data", leaving samples with only 2 allelic states. I guess in general you want to chose the ancestral state then the next most frequent allele as the two to keep.

I'm unsure as to why @szhan was having problems with triallelic sites. Perhaps because he was porting them directly into the ancestors TS, rather than allowing the `match_ancestors` routine to mark those sites as not-for-inference?

hyanwong:
> I'm unsure as to why @szhan was having problems with triallelic sites in https://github.com/tskit-dev/tsinfer/issues/647

Ah, he was using `make_ancestors_ts`. So that (undocumented) function simply needs to omit the non-biallelic sites. So perhaps we should change the name of this issue to "make_ancestors_ts should omit non-biallelic sites", @jeromekelleher ?

jeromekelleher:
Ah yes, that was it, thanks @hyanwong. Updated to remove all sites that either are singletons or have more than one mutation. Should be good to merge now and ready to use for @szhan 's application.

jeromekelleher:
Ah yes, that was it, thanks @hyanwong. Updated to remove all sites that either are singletons or have more than one mutation. Should be good to merge now and ready to use for @szhan 's application.

jeromekelleher:
Going to merge this so Shing can make progress, I think it's pretty uncontroversial.

jeromekelleher:
Going to merge this so Shing can make progress, I think it's pretty uncontroversial.

PR #660, opened 2022-04-12
Fix a typo in the docstring of subset()

szhan:
Corrected a minor typo. It should be "sample IDs" rather than "site IDs".


jeromekelleher (approved):
Thanks @szhan!

PR #675, opened 2022-06-10
Initial extend

jeromekelleher:
Early WIP, not ready for comments.

### DIFF
--- a/tsinfer/constants.py
+++ b/tsinfer/constants.py
@@ -37,6 +37,9 @@ NODE_IS_SAMPLE_ANCESTOR = 1 << 18
 # file
 NODE_IS_HISTORICAL_SAMPLE = 1 << 20
 
+# Undocumented.
+NODE_IS_IDENTICAL_SAMPLE_ANCESTOR = 1 << 21
+
 # What type of inference have we done at a site?
 INFERENCE_NONE = "none"
 INFERENCE_FULL = "full"
--- a/tsinfer/inference.py
+++ b/tsinfer/inference.py
@@ -1,5 +1,5 @@
 #
-# Copyright (C) 2018-2021 University of Oxford
+# Copyright (C) 2018-2022 University of Oxford
 #
 # This file is part of tsinfer.
 #
@@ -1096,6 +1096,7 @@ class Matcher:
         extended_checks=False,
         engine=constants.C_ENGINE,
         progress_monitor=None,
+        allow_multiallele=False,
     ):
         self.sample_data = sample_data
         self.num_threads = num_threads
@@ -1231,7 +1232,9 @@ class Matcher:
         # quickly be big enough even for very large instances.
         max_edges = 64 * 1024
         max_nodes = 64 * 1024
-        if np.any(num_alleles > 2):
+        if np.any(num_alleles > 2) and not allow_multiallele:
+            # Currently only used for unsupported extend operation. We can
+            # remove in future versions.
             raise ValueError("Cannot currently match with > 2 alleles.")
         self.tree_sequence_builder = self.tree_sequence_builder_class(
             num_alleles=num_alleles, max_nodes=max_nodes, max_edges=max_edges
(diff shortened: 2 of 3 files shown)


jeromekelleher:
Adding this to the 0.3.0 milestone as I'd like to get it in as undocumented functionality to enable some stuff downstream.

jeromekelleher:
I think this is ready to go in, so good to get some eyes on it. The basic idea is that this provides the basic infrastructure needed to do some inferences I'm working on, and which seem to be working reasonably well now. Most of the code is for working around awkward annoying problems which will hopefully go away in the 0.4.0 version of tsinfer with the more flexible API.

So, basically I'd like to get this in as an undocumented API to enable some upstream code, but aim to remove it for 0.4.0 when the new APIs will make it unnecessary.

@hyanwong - would you mind taking a quick look through please?

tsinfer/inference.py
```diff
@@ -1705,6 +1706,124 @@ def match_samples(self, sample_indexes, sample_times):
                 progress_monitor.update()
             progress_monitor.close()
 
+    def extend(self, samples, known_haplotypes, node_metadata):
+        """
+        Runs the "extend" operation by matching for an exemplar sequence
+        from each of the distinct categories.
```
hyanwong, line 1712: What are the "categories" here?

tsinfer/inference.py
```diff
@@ -2179,3 +2298,93 @@ def minimise(ts):
         filter_individuals=False,
         filter_populations=False,
     )
+
+
+def solve_num_mismatches(ts, k):
```
hyanwong, line 2303: I'm not quite following here, but I think that calculating the low-level recombination probability depends on the distance between the two sites, and I can't see that being used here. That might be OK for your usage, however.

hyanwong (approved):
I'm not quite following the general gist of this PR (it looks like you are adding the "exemplar" sample nodes as ancestors into the and_ts, right?), but I'm happy to take it on trust, especially if it is a temporary fix for something that can be done properly later. Perhaps you can talk me through the general idea in person, @jeromekelleher ?

tsinfer/constants.py
```diff
@@ -37,6 +37,9 @@
 # file
 NODE_IS_HISTORICAL_SAMPLE = 1 << 20
 
+# Undocumented.
+NODE_IS_IDENTICAL_SAMPLE_ANCESTOR = 1 << 21
```
hyanwong, line 41: I think we call these "proxy ancestors" in other code (i.e. an ancestor identical to a sample inserted  a little earlier in time, to help matching)

tsinfer/inference.py
hyanwong: Do you mean:

> Runs the "extend" operation by taking a set of samples, creating "exemplar" proxy samples for those which have multiple identical sequences, then matching the exemplar (or the sample if unique) into the existing tree sequence, returning a new tree sequence.

tsinfer/inference.py
```diff
@@ -2179,3 +2298,93 @@ def minimise(ts):
...
+            root_haplotype = next(sample_data.haplotypes())[1].copy()
+            root_haplotype[:] = 0
+            self.haplotypes = {root_haplotype.tobytes()}
+        else:
+            self.ancestors_ts = ancestors_ts
+            if time_units is not None and self.ancestors_ts.time_units != time_units:
+                raise ValueError(
+                    f"Mismatched time_units: {time_units} != ancestors_ts.time_units",
+                )
+            # Add in the existing haplotypes. Note - this will probably
```
hyanwong, line 2356: At scale, we probably don't want to match against the sample haplotypes in the next round of extend, if there is a simple "exemplar" we can match against instead. But there's no reason why we can't have these in the actual TS. Is there an argument here for a flag "NODE_HAS_OLDER_PROXY", which would simply cause the matching algorithm to skip matching against those nodes, since we are guaranteed that we can match against an older haplotype anyway?

hyanwong (approved):
This all LGTM, and given that's it's not documented, I think we should merge and released tsinfer 0.3. We can always improve iteratively, like the `NODE_HAS_OLDER_PROXY` flag I suggested above.

My main reservation is actually the naming. I was originally confused by the term "extend", which could be tend to mean extending the ts along the genome. I'm not sure it's immediately obvious what "extending an existing inferred TS" means. Can we think of alternative terminology, perhaps? This needn't stop merging, however.

tsinfer/inference.py
jeromekelleher: Yes, updated.

jeromekelleher:
Great, thanks @hyanwong. I've updated the documentation a bit.

I agree on the terminology, it's not great. It's not worth changing now though, as I think it'll be subsumed in to the more general "match" operation that's we're hoping to use eventually.

Also agreed on the exemplars thing, we'll want to figure out a better way of doing that in general.

PR #679, opened 2022-06-19
Set time units on ancestors and sample data files

### LINKED ISSUE #469: Add time_units attribute to AncestorData file
At the moment we get all the metadata for the final tree sequence from the SampleData file. However, there is at least one piece of metadata which makes sense to associate with an AncestorData file, which is the meaning of time (as yet unimplemented, but see https://github.com/tskit-dev/tskit/issues/495 and https://github.com/tskit-dev/tskit/issues/1051). In other words, should the ancestors file "know" whether the time units are years, generation, freq measures, etc?

This came up because of https://github.com/tskit-dev/tsinfer/pull/467, where it can be useful to know if the ancestors that we are dealing with are using frequency as time, or something else. It might also be sensible to be able to check whether the AncestorData file is using the same meaning of time as its associated SampleData file, if we ever allow this to occur. when e.g. running `insert_proxy_samples(.., require_same_sample_data=False)`.

hyanwong:
Fixes #469, but is a little involved. This adds a time_units attribute to both sample data and ancestor data files. By default, the value is `tskit.TIME_UNITS_UNKNOWN` for sample data files, but `tskit.TIME_UNITS_UNCALIBRATED` for ancestor data files.

When build_ancestors is called, if some of the site times are set in the sample data file, the default ancestor data file time units are instead taken from the sample data file rather than left to the default `tskit.TIME_UNITS_UNCALIBRATED`.

The time units in the ancestors_ts are taken from the ancestors file (not the sample data file), which I think is right. The time units in the final ts are taken from the ancestors ts.

I have also bumped the minor version numbers of both sample data and ancestor data files: I'm not sure if that is needed for attribute changes, but I figured it was worth it.

### DIFF
--- a/tests/test_formats.py
+++ b/tests/test_formats.py
@@ -162,6 +162,7 @@ class TestSampleData(DataContainerMixin):
         input_file.finalise()
         assert input_file.format_version == formats.SampleData.FORMAT_VERSION
         assert input_file.format_name == formats.SampleData.FORMAT_NAME
+        assert input_file.time_units == ts.time_units
         assert input_file.num_samples == ts.num_samples
         assert input_file.sequence_length == ts.sequence_length
         assert input_file.metadata_schema == ts.metadata_schema.schema
@@ -210,7 +211,9 @@ class TestSampleData(DataContainerMixin):
         with tempfile.TemporaryDirectory(prefix="tsinf_format_test") as tempdir:
             filename = os.path.join(tempdir, "samples.tmp")
             input_file = formats.SampleData(
-                path=filename, sequence_length=ts.sequence_length
+                path=filename,
+                sequence_length=ts.sequence_length,
+                time_units=ts.time_units,
             )
             self.verify_data_round_trip(ts, input_file)
             compressor = formats.DEFAULT_COMPRESSOR
@@ -285,14 +288,19 @@ class TestSampleData(DataContainerMixin):
 
     def test_defaults_no_path(self):
         ts = tsutil.get_example_ts(10)
-        with formats.SampleData(sequence_length=ts.sequence_length) as sample_data:
+        with formats.SampleData(
+            sequence_length=ts.sequence_length,
+            time_units=ts.time_units,
+        ) as sample_data:
             self.verify_data_round_trip(ts, sample_data)
             for _, array in sample_data.arrays():
                 assert array.compressor == formats.DEFAULT_COMPRESSOR
 
     def test_with_metadata_and_individuals(self):
         ts = tsutil.get_example_individuals_ts_with_metadata(5, ploidy=2)
-        with formats.SampleData(sequence_length=ts.sequence_length) as sample_data:
+        with formats.SampleData(
+            sequence_length=ts.sequence_length, time_units=ts.time_units
+        ) as sample_data:
             self.verify_data_round_trip(ts, sample_data)
 
     def test_access_individuals(self):
@@ -364,10 +372,22 @@ class TestSampleData(DataContainerMixin):
 
     def test_from_tree_sequence_simple(self):
         ts = tsutil.get_example_ts(10)
-        sd1 = formats.SampleData(sequence_length=ts.sequence_length)
+        sd1 = formats.SampleData(
+            sequence_length=ts.sequence_length, time_units=ts.time_units
+        )
         self.verify_data_round_trip(ts, sd1)
         sd2 = formats.SampleData.from_tree_sequence(ts, use_sites_time=True)
         assert sd1.data_equal(sd2)
+
+    def test_time_units(self):
+        ts = tsutil.get_example_ts(10)
+        tables = ts.dump_tables()
+        tables.time_units = "random time units"
+        ts = tables.tree_sequence()
+        sd1 = formats.SampleData(
+            sequence_length=ts.sequence_length, time_units=ts.time_units
+        )
+        self.verify_data_round_trip(ts, sd1)
 
     def test_from_tree_sequence_variable_allele_number(self):
         ts = tsutil.get_example_ts(10)
@@ -393,7 +413,9 @@ class TestSampleData(DataContainerMixin):
         ts = tables.tree_sequence()
         assert len(ts.site(0).mutations) > 1
         assert len(ts.site(ts.num_sites - 1).mutations) == 0
-        sd1 = formats.SampleData(sequence_length=ts.sequence_length)
+        sd1 = formats.SampleData(
+            sequence_length=ts.sequence_length, time_units=ts.time_units
+        )
         self.verify_data_round_trip(ts, sd1)
         num_alleles = sd1.num_alleles()
         for var in ts.variants():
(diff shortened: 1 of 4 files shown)


hyanwong:
I'm not sure what's going on with the coverage here. I can't see what I've missed, but it seems down by 3.20%

tests/test_formats.py
```diff
@@ -364,11 +372,23 @@ def test_from_tree_sequence_bad_populations(self):
 
     def test_from_tree_sequence_simple(self):
         ts = tsutil.get_example_ts(10)
-        sd1 = formats.SampleData(sequence_length=ts.sequence_length)
+        sd1 = formats.SampleData(
```
jeromekelleher, line 375: Why change all these tests? It doesn't look like ``verify_data_round_trip`` checks time_units, so why alter these tests? We want to know that old code will continue to work, so *not* changing existing tests would be much more reassuring.

tests/test_inference.py
```diff
@@ -774,6 +774,36 @@ def test_three_sites(self):
...
 
 
+class TestTimeUnits:
+    """
+    Tests that time units get set properly
+    """
+
+    def test_normally_uncalibrated(self):
+        ts = msprime.simulate(10, mutation_rate=5, recombination_rate=4, random_seed=21)
+        assert ts.time_units != tskit.TIME_UNITS_UNCALIBRATED
```
jeromekelleher, line 784: Might as well make this concrete, ``ts.time_units == "generations"`` as it's not going to change.

tests/test_inference.py
```diff
@@ -774,6 +774,36 @@ def test_three_sites(self):
         self.verify([[1, 1], [1, 1], [1, 1]])
 
 
+class TestTimeUnits:
+    """
+    Tests that time units get set properly
+    """
+
+    def test_normally_uncalibrated(self):
+        ts = msprime.simulate(10, mutation_rate=5, recombination_rate=4, random_seed=21)
```
jeromekelleher, line 783: If we're not testing the data here I don't see any value in calling ``msprime.simulate``. We're just adding extra complexity here from the effects of ``from_tree_sequence`` which is very complicated and isn't the actual path that users will use. It would be better to have a super-simple minimal example, and check the individual steps in isolated tests.

Probably good to test from_tree_sequence by itself, as I guess that may inherit the time_units in different ways as well.

tests/test_inference.py
```diff
@@ -3781,6 +3811,36 @@ def test_standard_pipeline(self):
                 sample = inf_ts.node(u)
                 assert sample.time == t
 
+    def test_insert_proxy_time_units_mismatch(self):
+        samples = [msprime.Sample(population=0, time=t) for t in (0.0, 0.0, 0.1, 1.5)]
```
jeromekelleher, line 3815: I don't find this test particularly enlightening - it hinges on me knowing precisely what ``from_tree_sequence`` does as a side effect in terms of the ``time_units`` variable. We should be aiming for tests that someone can read in isolation. There's no point in simulating a bunch of data here as we're never testing the output of that process.

tsinfer/formats.py
```diff
@@ -915,6 +924,9 @@ class SampleData(DataContainer):
         site coordinates must be less than this value.
     :param str path: The path of the file to store the sample data. If None,
         the information is stored in memory and not persistent.
+    :param str time_units: If times are input for sites or individuals, these
```
jeromekelleher, line 927: It would be simpler to *always* set the time_units to this value if specified.

tsinfer/formats.py
```diff
@@ -933,16 +945,17 @@ class SampleData(DataContainer):
...
     # State machine for handling automatic addition of samples.
     ADDING_POPULATIONS = 0
     ADDING_SAMPLES = 1
     ADDING_SITES = 2
 
-    def __init__(self, sequence_length=0, **kwargs):
-
-        super().__init__(**kwargs)
+    def __init__(self, sequence_length=0, *, time_units=None, **kwargs):
+        if time_units is None:
```
jeromekelleher, line 956: Don't need to do this, the superclass should handle time_units automatically.

tsinfer/formats.py
```diff
@@ -933,16 +945,17 @@ class SampleData(DataContainer):
     """
 
     FORMAT_NAME = "tsinfer-sample-data"
-    FORMAT_VERSION = (5, 0)
+    FORMAT_VERSION = (5, 1)
```
jeromekelleher, line 948: I don't think this is a minor version bump - don't we need to do any upgrade to handle missing ``time_units`` values on older files?

jeromekelleher (commented):
This seems like quite a big change to me, and one that's likely to have unforeseen consequences. I have a lot of trouble seeing how it fits together now, with ``sites_time``, ``individuals_time`` etc.

How about we just add a ``time_units`` argument to ``infer`` or ``match_samples``/``match_ancestors`` which sets the time units to a given value or otherwise just results in ``uncalibrated``? It's the same thing ultimately, because people will need to understand what these things mean when writing their scripts and it saves a whole bunch of plumbing and complexity.

hyanwong:
Thanks a lot for looking at this @jeromekelleher, really appreciated.

> This seems like quite a big change to me, and one that's likely to have unforeseen consequences. I have a lot of trouble seeing how it fits together now, with `sites_time`, `individuals_time` etc.

Yes, I agree, it does make it annoyingly complex.

> How about we just add a `time_units` argument to `infer` or `match_samples`/`match_ancestors` which sets the time units to a given value or otherwise just results in `uncalibrated`? It's the same thing ultimately, because people will need to understand what these things mean when writing their scripts and it saves a whole bunch of plumbing and complexity.

That seems a good idea for me in the short term, although maybe we don't want to document it, in case we come up with a different solution.

What I was trying to do in this PR was to associate the time units with the *ancestors* file, as it feels like that's where the meaning of the time units should apply. For example, if we call `tsinfer.SampleData.from_tree_sequence(ts use_sites_time=True)`, when inference is run on the resulting sample data file, it should have the `time_units` set to that of `ts`, whereas when it is run with `use_sites_time=False`, it should set `time_units` to "uncalibrated". With your scheme I guess it would be down to the user to know where the sample_data file has come from?

hyanwong:
Just a follow-up: what we *really* want to do, I think, is detect whether when inference was done, we set times-as-frequencies either [here](https://github.com/tskit-dev/tsinfer/blob/34bd0648d24eb0550f64cb86086b994a9abe8a8c/tsinfer/inference.py#L359) or in the depths of the ancestor builder [here](https://github.com/tskit-dev/tsinfer/blob/34bd0648d24eb0550f64cb86086b994a9abe8a8c/tsinfer/inference.py#L832)

jeromekelleher:
I don't know. We basically just want ``time_units`` to be set to uncalibrated in the majority of cases. I'd be surprised if anyone who's not already an expert in all this stuff is using the ``sites_times`` stuff, so maybe we should just assume that they can fix their time units afterwards. I think my vote here would be to do the very simplest possible thing and to simply set ``time_units`` as uncalibrated unconditionally, and deal with exceptions to this later in follow-ups.

Presumably tsdate just sets ``time_units`` to "generations" unconditionally at the end of inference?

@awohns, any thoughts?

hyanwong:
> Presumably tsdate just sets `time_units` to "generations" unconditionally at the end of inference?

Yes, there's a parameter `time_units` that I introduced which if None, sets the `time_units` to generations.

jeromekelleher:
Well, I think we then get 99% of the value of this by just setting ``time_units`` to uncalibrated unconditionally for all tree sequences output by tsinfer and dealing with the complicated edge cases later.

hyanwong:
Yep, I agree with the 99%. I still think we should not document this for the moment though. The simplest route to me would be to have a `time_units` parameter to match_ancestors, which sets the time units on the ancestors_ts (if None, use uncalibrated). I think this then gets ported through to the final ts automatically.

hyanwong:
Keeping this open as a draft. Alternative simple approach in #680

PR #681, opened 2022-06-21
Allow sequence length to be specified in sd.subset

hyanwong:
Just hit this bug - I wanted to change the seq length when subsetting, which the docs imply should be possible.


tsinfer/formats.py
```diff
@@ -1356,7 +1356,9 @@ def subset(self, individuals=None, sites=None, **kwargs):
         sites = set(sites)
         if len(sites) != num_sites:
             raise ValueError("Duplicate site IDS")
-        with SampleData(sequence_length=self.sequence_length, **kwargs) as subset:
+        if "sequence_length" not in kwargs:
```
jeromekelleher, line 1359: What about adding ``sequence_length=None`` to the args above, and then:
```
sequence_length = sequence_length if sequence_length is not None else self.sequence_length
```
Manually looking into kwargs should be a last-resort option. (Should make sequence_length kw only though)

jeromekelleher (commented):
No need for messing with kwargs dict here.

tsinfer/formats.py
hyanwong: Done

jeromekelleher (approved):

Issue #622, opened 2022-01-08
Is the parameter "simplify" correctly named?


hyanwong:
It seems a little confusing, to those who don't know the details of `simplify`, that e.g. tsdate requires a "simplified" ts (i.e. [without unary nodes](https://github.com/tskit-dev/tsdate/pull/188)), but that specifying `tsinfer.infer(sample_data, simplify=True)` does *not* produce the right sort of tree sequence and a further round of simplification is needed to remove the unary nodes.

Therefore, I wonder if it is worth changing the parameter name to something else. Here's one suggestion: we could reverse the meaning, and have `keep_all_nodes`, with `None` doing the current behaviour, `False` doing a full simplify (i.e. removing unary nodes), and True omitting all simplification. I'm not sure if this is a bit too opaque, though, and if so, whether a better parameter name would fix it? Other suggesting welcome.

jeromekelleher:
That sounds like a pretty aggressively breaking change to me @hyanwong - wouldn't this completely break existing code?

If we change this we need to make sure that the default behaviour remains the same (I think we're happy that this is the output we want people to do their evaluations against?), and that we don't break existing code.

Without thinking too hard about it, it seems like a lot of work for not a whole pile of gain?

hyanwong:
I think we would retain (but deprecate) the `simplify` parameter? And yes, keep the default the same. Perhaps this isn't needed so much if tsdate errors out on tree sequences with unary nodes, but I do think it might be a rather confusingly named parameter? Happy to leave it though.

Issue #673, opened 2022-06-02
Split the all-zeros root node into pieces


hyanwong:
As discussed in various places (slack, [GitHub issues](https://github.com/tskit-dev/tsinfer/issues/672#issuecomment-1142328987)), we think that having a single ancestor with all the ancestral states onto which everything matches is potentially misleading, and certainly problematic for dating.

After discussion with @jeromekelleher, a reasonable compromise seems to be to take all the edges which reference this node, take the right and left coordinates and uniqueify them, then create a new root parent node within each interval. This will probably require cutting up some edges, but it is less extreme than creating a new root node for each tree. We can do this as the final part of outputting the tree sequence (we want to address #649 first, though).

Incidentally, we need not create a new node if in that interval the root node is entirely unary. But perhaps we would want to add a redundant node in this case for consistency?

Hopefully this will make the "ignore_oldest_root" option in tsdate redundant (https://github.com/tskit-dev/tsdate/pull/138)

hyanwong:
We also think that it should not be necessary to do anything with unary nodes under the all-zeros root, because if there has been matching against one of these, rather than the all-zeros root, it implies that it matched against one of the derived mutations further along the sequence.

hyanwong:
It would be useful to see how many extra nodes this approach generates when run on the unified trees in [zenodo](https://zenodo.org/record/5512994#.YpimtC8w08R).

hyanwong:
I have coded up the grand-MRCA-splitting code in #687 (defaults to `keep_single_mrca=False`), and it makes quite a large different when running tsdate (pinging @awohns). Here's an example from the [Brandt paper code](https://github.com/deboraycb/ARGsims) (done by running a modified version of `python3 ./2_tsinfer_date.py l5mb 8 2e-8 > tsinfer_date_l5mb.out2` then `Rscript 5_plotTc_tsdate.R -p l5mb -N 20000 -g priorDef`

### With "ignore_oldest_root=True" in `tsdate` and "keep_single_mrca"=True in `tsinfer`
(this is like what it was before)
![sim_l5mb_priorDef_inf_pmf_clean_singleMRCA_omit_oldest](https://user-images.githubusercontent.com/4699014/178232756-8db48ced-bddf-4cda-ae2f-507a8b659c78.png)

### With "ignore_oldest_root=True" in `tsdate` and "keep_single_mrca"=False in `tsinfer`
(splits the grand MRCA into pieces)
[...]

Issue #649, opened 2022-03-11
Remove ultimate ancestor using subset


hyanwong:
Since we never actually need the ultimate ancestor once we have run match_ancestors, can we simply remove it from the ancestors tree sequence using `tables.subset()`? It's a bit odd leaving it in. Is it necessary for matching samples against the ancestors TS? I think not.

jeromekelleher:
I think we could probably just not add it in the first place, but likely this will break some tests.

gtsambos:
If this global CA is actually informative -- which maybe it could be, idk, if you trusted its tsdated age, and it's an MRCA? -- then it could even be useful to keep. But it's unlikely that this is what it actually represents though, right? You have no information about mutations that have arisen before each local MRCA, so I imagine you wouldn't be doing a good job of dating this ultimate ancestor.

hyanwong:
I think I agree with you @jeromekelleher that neither are things that we want in the final tree sequence, even though we do normally want to keep unary nodes. We would need to check that tsdate is happy with this (which I think it should be).

As a halfway house, we could remove these nodes from the final output tree sequence by post processing? Then at some future point we could remove them from within the algorithm itself, should that be necessary?

jeromekelleher:
> As a halfway house, we could remove these nodes from the final output tree sequence by post processing? Then at some future point we could remove them from within the algorithm itself, should that be necessary?

I think we could get rid of one of them all right. I think it would be good if we could say that node 0 is the ultimate ancestor consisting of all zeros, in the standard pipeline, as this is what we assume in any case.

hyanwong:
> I think we could get rid of one of them all right. I think it would be good if we could say that node 0 is the ultimate ancestor consisting of all zeros, in the standard pipeline, as this is what we assume in any case.

What if node 0 is not coalescent in a tree though. Do we remove the edge(s) from the oldest MRCA to node 0, so that the haplotype given by 0 in the tree sequence might not span the whole sequence? I.e. in Georgia's picture below

![Screen Shot 2022-05-31 at 4 07 14 pm](https://user-images.githubusercontent.com/4699014/171189430-6d3c0503-c844-4e75-943c-e0aa0f4ca472.png)

Do we delete 12 (which I assume is the all-zero CA) from the left hand tree? And how about node 11 in the left hand tree?

hyanwong:
[...]

Issue #275, opened 2020-06-15
Root and ultimate ancestor oddities


hyanwong:
Investigating the tree sequences produced by the latest tsinfer version reveals some oddities in the two oldest ancestors. Now that times are set as *frequencies* by default, it seems rather odd that all; the ancestors are at times 0...1 except the 2 oldest, which are at times ~2.0 and ~3.0, because we insert a standard root at max_time+1 and an "ultimate ancestor" at time max_time+2. Since it's only the relative time that is important, this shouldn't be a problem, but it might be helpful (e.g. for plotting reasons) if the max_ancestor_time is < 1, to set the root_time to 1.

Additionally, in https://github.com/tskit-dev/tsinfer/blob/master/tsinfer/inference.py#L587 the comments say

"Add the ultimate ancestor. This is an awkward hack really; we don't ever insert this ancestor."

I suggest that if this ancestor is never expected to be inserted, we should set the time to `+inf`, so that if it *is* inserted we get an error. I did this, and it turn out it *is* inserted and trust up in the final tree sequence, for instance:

```
ts = msprime.simulate(10, recombination_rate=1, mutation_rate=2, random_seed=123)
sd = tsinfer.SampleData.from_tree_sequence(ts, use_times=False)
anc = tsinfer.generate_ancestors(sd)
print("Ultimate time (should not be used):", anc.ancestors_time[0])
print("Universal root (should be oldest):", anc.ancestors_time[1])

a_ts =  tsinfer.match_ancestors(sd, anc)
infer_ts =  tsinfer.match_samples(sd, a_ts)
print("Oldest time in ts is:", infer_ts.max_root_time)
```

giving

```
Ultimate time (should not be used): 2.8
Universal root (should be oldest): 1.8
Oldest time in ts is: 2.8
```

jeromekelleher:
Yes, the oldest ancestor stuff is hacky. We should fix this for the next release.

jeromekelleher:
Pushing this back to the 0.3 series, the ultimate ancestor assumption is buried deep in the guts of the LS algorithm and it's not worth messing with it now for 0.2.

hyanwong:
Just to note - I suspect that the oldest root is present in the final inferred ts because when we simplify we now `keep_unary=True`. Previously it was simplified out.

jeromekelleher:
That's a good point.

hyanwong:
[...]

PR #687, opened 2022-06-29
Remove the ultimate ancestor and split the grand MRCA

hyanwong:
In a [few](https://github.com/tskit-dev/tsinfer/issues/649#issuecomment-1143628417) [issues](https://github.com/tskit-dev/tsinfer/issues/275#issuecomment-1065012704) we decided to remove the "ultimate" ancestors (the hack that means that the ancestors IDs in the ancestors file and the tree sequence align. This is needed as a precursor to #673.

This does this, removing the assert, but on my laptop I'm finding this now fails in unexpected ways in find_path etc, which makes me think that we are removing one of the ancestors required for LS matching.

### DIFF
--- a/requirements/CI-tests-complete/requirements.txt
+++ b/requirements/CI-tests-complete/requirements.txt
@@ -16,5 +16,5 @@ pytest-xdist==2.5.0
 seaborn==0.11.2
 sortedcontainers==2.4.0
 tqdm==4.64.0
-tskit==0.5.0
+tskit==0.5.3
 zarr==2.11.3
--- a/tests/test_inference.py
+++ b/tests/test_inference.py
@@ -205,6 +205,7 @@ class TestRoundTrip:
         ]
         for params in test_params:
             ts = tsinfer.infer(sample_data, **params)
+
             self.assert_lossless(
                 ts, genotypes, positions, alleles, sample_data.sequence_length
             )
@@ -1018,12 +1019,11 @@ class TestMetadataRoundTrip:
             t1 = output_ts.simplify(subset).dump_tables()
             assert len(t1.individuals.metadata) > 0
             assert len(t1.individuals.location) > 0
-            t1.provenances.clear()
             t2 = tsinfer.match_samples(
                 sample_data, ancestors_ts, indexes=subset
             ).dump_tables()
             t2.simplify()
-            assert t1.equals(t2, ignore_provenance=True)
+            t1.assert_equals(t2, ignore_provenance=True)
 
     def test_individual_location(self):
         ts = msprime.simulate(12, mutation_rate=5, random_seed=16)
@@ -1920,9 +1920,10 @@ class TestMatchSamples:
     """
 
     def test_partial_samples(self):
-        ts = msprime.simulate(
-            10, mutation_rate=2, recombination_rate=2, random_seed=233
+        ts = msprime.sim_ancestry(
+            10, sequence_length=1e4, recombination_rate=2e-4, random_seed=233
         )
+        ts = msprime.sim_mutations(ts, rate=2e-4, random_seed=233)
         for tree_seq in [ts, eval_util.strip_singletons(ts)]:
             sd = tsinfer.SampleData.from_tree_sequence(tree_seq, use_sites_time=False)
             ts1 = tsinfer.infer(sd)
@@ -1971,6 +1972,17 @@ class TestMatchSamples:
             sample_data.add_site(0.5, [1, 1])
         ts = tsinfer.infer(sample_data, time_units="generations")
         assert ts.time_units == "generations"
+
+    def test_ultimate_ancestor_removed(self):
+        ts = msprime.simulate(10, mutation_rate=10, recombination_rate=2, random_seed=1)
+        assert ts.num_sites > 0
+        sd = tsinfer.SampleData.from_tree_sequence(ts)
+        anc = tsinfer.generate_ancestors(sd)
+        ultimate_ancestor = anc.ancestor(0)
+        ancestors_ts = tsinfer.match_ancestors(sd, anc)
+        assert ancestors_ts.num_sites > 0
+        final_ts = tsinfer.match_samples(sd, ancestors_ts)
+        assert ultimate_ancestor.time not in final_ts.tables.nodes.time
 
 
 class AlgorithmsExactlyEqualMixin:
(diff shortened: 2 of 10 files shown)


hyanwong:
Note that the ancestors IDs and the nodes in the ancestors tree sequence won't align anyway if we allow path compression during match_ancestors.

hyanwong:
Actually the problem is [here](https://github.com/tskit-dev/tsinfer/blob/87f7b54c4ac6294af46cf37095f6881d1df59995/tsinfer/algorithm.py#L873). I think the issue is that at the matching algorithm assumes that all the nodes against which we match have an edge above it, we don't have this if we remove the ultimate ancestor.

jeromekelleher:
Yes, it's not straightforward - it's not obvious how to fix easily.

jeromekelleher:
We also make a hard assumption that zero if the ultimate root, which makes it very tricky to use arbitrary tree sequences as input. We'd need to fix that too. My best guess on the simplest way to fix this is to make an "virtual root" node at the end of the node table, and to add in edges from all the actual roots to this, and then change the places where we're assuming 0 to this node. But it's highly optimised code and the logic is very tricky (it's mainly about nodes that fall out of the tree, and keeping track of that) so it's not an easy thing to do.

hyanwong:
So are we still of the opinion that this is the best thing to do, or should we go to the fallback of deleting the ultimate ancestor from the resulting tree sequence?

If we leave it in during ancestor generation, we could set the time of the ultimate ancestor in tsinfer to `np.inf`, then it can act as a sort of sentinel, to ensure that we can't put it into a tree sequence (we can remove it before calling `tables.tree_sequence()`?

jeromekelleher:
I don't know to be honest. It's not really that much of a problem and we have things that urgently need to be done, so I'd rather park it.

hyanwong:
It's stopping me addressing #673 which does actually make a difference. So I'm happy to take an executive decision and leave it in with `np.inf` as the time (or at least, test whether this works)

hyanwong:
So I think I've figured out the easiest way to do this: we require the ultimate ancestor when matching. But we don't do matching on a simplified tree sequence. So the easiest thing is to only remove the ultimate ancestor when we simplify, during the final output stage. The nicest way to do this, IMO, is to remove the (single) edge connecting the ultimate ancestor (ancestor ID 0) to the grand MRCA (ancestor ID 1), then let `simplify` take care of the removal of the node.

I've adjusted this PR to do that. The only reasons it is now failing is the test where we compare the simplified output to the output when running the standard `.simplify(keep_unary=True)`, as these outputs now differ because of the removal of the ultimate ancestor. I guess that's fine, and we can adjust the tests.

Comments as to whether this is the best approach are very welcome.

hyanwong:
I have this now in satisfactory shape, so it's ready for review, hurrah.

hyanwong:
This will also fix https://github.com/tskit-dev/tsinfer/issues/622

tsinfer/inference.py
```diff
@@ -1924,6 +1912,156 @@ def get_mutations(self, node_id):
...
+        new_root = tables.nodes.append(genomewide_mrca)
+        for c in tree.children(genomewide_mrca_id):
+            tables.edges.add_row(parent=new_root, child=c, left=left, right=right)
+        while mutation_pos < right:
+            tables.mutations[mutation_id] = tables.mutations[mutation_id].replace(
+                node=new_root
+            )
+            mutation_id, mutation_pos = next(mut_iter, (None, ts.sequence_length))
+        left = right
+    return True
```
jeromekelleher, line 2062: What's the return for?

tsinfer/inference.py
```diff
@@ -1924,6 +1912,156 @@ def get_mutations(self, node_id):
...
+    )
+    logger.info(
+        "Finished simplify; now have {} nodes and {} edges".format(
+            tables.nodes.num_rows, tables.edges.num_rows
+        )
+    )
+    return tables.tree_sequence()
+
+
+def detach_node_over_grand_root(tables):
```
jeromekelleher, line 1991: Split this into a function ``has_single_edge_over_grand_root`` . The actual taking off is trivial and can be done like:
```
if has_single_edge_over_grand_root(tables):
    tables.edges.truncate(tables.edges.num_rows - 1)
```

jeromekelleher (commented):
Looks good generally. Needs to be decomposed into simpler functions that we can hopefully test well in isolation.

hyanwong:
> Looks good generally. Needs to be decomposed into simpler functions that we can hopefully test well in isolation.

Thanks *a lot* for looking at this. I restructured the code to avoid all that sorting and turning backwards and forwards into tables/tree sequences, although that's unavoidable to loop through the trees after detaching the "virtual root-like" ancestor. I guess one or two more tests of the `has_single_edge_over_grand_root` and `has_same_root_everywhere` helper functions (should these be prefixed with an underscore?) might be good, but is this new approach OK?

Comments on algorithms etc in replied to the feedback above.

tsinfer/inference.py
```diff
@@ -1924,6 +1912,156 @@ def get_mutations(self, node_id):
...
+    # efficient way to do this using the quintuply linked data structure
+    roots = {root for tree in ts.trees() for root in tree.roots}
+    if len(roots) > 1:
+        return False
+
+    logger.info("Splitting the all zeros grand MRCA into separate nodes")
+    genomewide_mrca_id = roots.pop()
+    assert genomewide_mrca_id == tables.edges[tables.edges.num_rows - 1].parent
+    genomewide_mrca = tables.nodes[genomewide_mrca_id]
+    edges_to_remove = tables.edges.parent == genomewide_mrca_id
```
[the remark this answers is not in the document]
hyanwong: It's neater to reverse the `ts.edges()` iterator, I think, so that we can use the IDs (and if we do it on the ts rather than the tables, we are assured that the order is correct):

```python
    root_breaks = set()
    keep_edges = 0
    for e in reversed(ts.edges()):
        # all the edges at the end are the ones to the genomewide_mrca
        if e.parent == genomewide_mrca_id:
            root_breaks |= {e.left, e.right}
        else:
            keep_edges = e.id + 1
            break
    tables.edges.truncate(keep_edges)
```

tsinfer/inference.py
hyanwong: Removed anyway with new code arrangement

tsinfer/inference.py
hyanwong: Done, although we need to move the mutations afterwards too, of course

hyanwong:
NB currently failing because I've started to use the negative node indexing stuff in the upcoming `tskit 0.5.3`. We can wait to merge until that's released, perhaps (as long as the approach is correct)

tsinfer/inference.py
```diff
@@ -1907,6 +1909,149 @@ def get_mutations(self, node_id):
         return self.mutations[node_id]
 
 
+def has_single_edge_over_grand_root(ts):
+    # Internal function to check if this is a "raw" inferred tree sequence.
+    if ts.num_edges < 2:
+        # must have edge to grand root and above grand root
+        return False
+    last_edge = ts.edge(-1)
```
jeromekelleher, line 1917: I'd find it easier to follow the logic here (and it would be easier to test) if the full-span requirement was tested first. This clause will also catch split edges like (0, x, u, vr), (x, L, u, vr).

tsinfer/inference.py
```diff
@@ -1907,6 +1909,149 @@ def get_mutations(self, node_id):
...
+        split_mrca = True
+
+    tables = ts.dump_tables()
+
+    if has_single_edge_over_grand_root(ts):
+
+        logger.info("Removing the oldest edge to detach the virtual-root-like ancestor")
+        last_edge = ts.edge(
+            -1
+        )  # Edge with oldest parent must be last in the edge table
```
jeromekelleher: I would either remove this comment or put it on the line above so we don't get this line-splitting from Black.

jeromekelleher (commented):
I've had a good look through and I *think* the logic is correct. Needs some tests for the helper functions, (i.e. class TestHasSameRootEverywhere ... try on some different wacky cases)

tsinfer/inference.py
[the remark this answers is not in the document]
jeromekelleher: I'd find the while-loop version easier (and shorter) to read tbh, but :shrug:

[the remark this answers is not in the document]
hyanwong: There's a bug in that one: I think we want to `truncate(j+1)`, which was another reason for spelling out `keep_edges` explicitly. But happy to switch back.

jeromekelleher:
LMK when this is ready for final review please @hyanwong

tests/test_inference.py
```diff
@@ -2508,6 +2518,228 @@ def test_many_trees(self):
...
+    oldest = tables.nodes.add_row(time=3)
+    tables.edges.add_row(parent=oldest, child=1, left=0, right=ts.sequence_length)
+    u = tables.nodes.add_row(flags=tskit.NODE_IS_SAMPLE)  # add an isolated node too
+    yield tables.tree_sequence(), "Single_root_edge"
+
+    tables.edges.add_row(parent=oldest, child=u, left=0, right=ts.sequence_length)
+    yield tables.tree_sequence(), "Multi_root_edge"
+
+
+def ts_root_test_params():
```
jeromekelleher, line 2582: Nice!

tests/test_inference.py
```diff
@@ -2508,6 +2518,228 @@ def test_many_trees(self):
...
+        roots = {root for tree in ts.trees() for root in tree.roots}
+        if len(roots) == 1 and tskit.NULL not in roots:
+            assert tsinfer.has_same_root_everywhere(ts)
+        else:
+            assert not tsinfer.has_same_root_everywhere(ts)
+
+    @pytest.mark.parametrize("ts", ts_root_test_params())
+    def test_has_single_edge_over_grand_root(self, ts, request):
+        if request.node.callspec.id == "Single_root_edge":
+            assert tsinfer.has_single_edge_over_grand_root(ts)
```
jeromekelleher, line 2603: This is brittle and needs the reader to understand the innards of pytest. Better to break up the list of examples (or to make this property a parameter of ``ts_root_test_params()``?)

tsinfer/inference.py
```diff
@@ -1912,6 +1908,146 @@ def get_mutations(self, node_id):
...
+    :func:`match_samples`. It involves removing the oldest node if it sits above a
+    grand MRCA ancestor (this "virtual-root-like node" is used to enable LS matching)
+    then splitting up the grand MRCA (if split_mrca is ``True``), and finally
+    reordering the sample nodes such that they are the first nodes listed in the
+    node table,  removing tree nodes and edges that are not on a path between the
+    root and any of the samples (by applying the :meth:`tskit.TreeSequence.simplify`
+    method with ``keep_unary`` set to True but ``filter_sites``, ``filter_populations``
+    and ``filter_individuals`` set to False).
+
+    :param bool split_mrca: If the oldest node is only a parent to a single
```
jeromekelleher, line 1952: Move the bulk of this discussion to a paragraph rather than within the ``split_mrca`` parameter description - these should be short and focus on things like types, etc.

jeromekelleher (commented):
LGTM. Needs another pass to make sure we're clear on the compatibility semantics going forward, and we deal with the single-tree case nicely.

hyanwong:
> LGTM. Needs another pass to make sure we're clear on the compatibility semantics going forward, and we deal with the single-tree case nicely.

Great, thanks a lot for looking over this: a few people will benefit from this new feature (I mentioned it to at least 2 people at the ARG symposium). All your comments seems good improvements. I'll close them off one by one now.

tsinfer/inference.py
```diff
@@ -1912,6 +1908,146 @@ def get_mutations(self, node_id):
...
+    )
+    return tables.tree_sequence()
+
+
+def split_grand_mrca(tables):
+    # Internal function: if a single oldest node is a root across the entire genome,
+    # split it up into a set of contemporaneous nodes whenever the node children change
+
+    ts = tables.tree_sequence()
+    if not has_same_root_everywhere(ts):
```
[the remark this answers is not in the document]
hyanwong: We haven't done that though, have we? I can only find `has_same_root_everywhere` being called once in the entire file. Above we check whether there is a "virtual-root-like" ancestor, but that's not the same. For instance, this would fail if there were isolated nodes in the TS.

tsinfer/inference.py
```diff
@@ -221,22 +221,24 @@ def infer(
...
-    simplify=True,
     exclude_positions=None,
+    post_process=None,
     num_threads=0,
     # Deliberately undocumented parameters below
     precision=None,
     engine=constants.C_ENGINE,
     progress_monitor=None,
     time_units=None,
+    simplify=True,  # Deprecated
```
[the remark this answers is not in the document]
hyanwong: Yes, this all seems good to me apart from:

> If the user explicitly sets simplify=True, we do the same thing as before (we don't do the new things - which we can, right?)

There's no path to do that right now, but I can add it.

tsinfer/inference.py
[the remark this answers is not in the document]
jeromekelleher: Right, sorry, I mixed this up with has_single_edge_over_grand_root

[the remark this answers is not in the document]
jeromekelleher: Needs a test though, currently not covered.

tests/test_inference.py
```diff
@@ -2508,6 +2518,228 @@ def test_many_trees(self):
...
+        ts = tsinfer.post_process(ts)
+        # After post-processing, samples should be 0..n
+        for i, (orig_id, inferred_node_id) in enumerate(zip(idx, ts.samples())):
+            assert i == inferred_node_id
+            sd_individual = medium_sd_fixture.individual(
+                medium_sd_fixture.sample(orig_id).individual
+            )
+            inferred_individual = ts.individual(ts.node(inferred_node_id).individual)
+            assert sd_individual.metadata["id"] == inferred_individual.metadata["id"]
+
```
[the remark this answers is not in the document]
hyanwong: > We need a test here for the case where we _infer_ a single tree (recombination rate = ~0), verifying that we don't change the input at all. 

Sorry, I'm being dense and don't quite understand your comments here. You mean that we don't change the genotypes, or that we infer exactly the same tree as we used to generate the variant data. If you mean the same tree, then we do need to drop the oldest edge (as that won't be in the original, simulated tree)

tsinfer/inference.py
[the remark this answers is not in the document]
hyanwong: I think it's in `TestRootCheckingFunctions::test_has_same_root_everywhere`?

hyanwong:
Apart from addressing you comments above, I'm adding one extra thing, which is that I now call `logger.warning` when we post process within the inference step (i.e. in match samples) and we can't find the virtual-root-like-ancestor and/or the grand MRCA. I've done this because at the moment, we simply invisibly avoid splitting etc if the inferred tree sequence is not in the expected format, and I'dm prefer to know that there's something unexpected about the format of the output inferred TS in this case.

tsinfer/inference.py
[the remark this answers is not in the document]
jeromekelleher: Seems worthwhile - no point in breaking people's code for no reason.

jeromekelleher:
SGTM. Might also be worth issuing a warning when users specify ``simplify=True`` suggesting they change over to ``post_process=True``.

tsinfer/inference.py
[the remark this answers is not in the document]
hyanwong: Yep, I hadn't thought of this avenue. Added in my local changes. Will push in a tick.

tests/test_inference.py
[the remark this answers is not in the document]
hyanwong: Thanks for clarifying just now @jeromekelleher  - I've added a check so that we simply return without doing anything if there is only a single edge to the grand MRCA, so we don't delete and recreate a node.

hyanwong:
I hope I've addressed all the comments now: I've done them as a separate commit so the changes are obvious. Let me know if they look OK, @jeromekelleher, and I can squash the latest commit into the previous one.

CHANGELOG.rst
```diff
@@ -13,6 +13,14 @@
 
 **Breaking changes**:
 
+- The ``simplify`` parameter is now deprecated in favour of ``post_process``, which
+  prior to simplification, removes the "virtual-root-like" ancestor (inserted by
+  tsinfer to aid the matching process) then splits the grand MRCA into separate pieces.
+  If splitting is not required, the ``post_process` step can also be called as a
```
jeromekelleher, line 19:
```suggestion
  If splitting is not required, the ``post_process`` step can also be called as a
```

requirements/CI-tests-complete/requirements.txt
```diff
@@ -16,5 +16,5 @@ pytest-xdist==2.5.0
 seaborn==0.11.2
 sortedcontainers==2.4.0
 tqdm==4.64.0
-tskit==0.5.0
+tskit==0.5.3
```
jeromekelleher, line 19: Also need to update our installation requirements, probably in setup.cfg.

tsinfer/inference.py
```diff
@@ -589,25 +592,41 @@ def match_samples(
...
         of the sample.
     :rtype: tskit.TreeSequence
     """
+    simplify_only = False  # if true, carry out "old" (deprecated) simplify behaviour
+    if simplify is None:
+        if post_process is None:
+            post_process = True
+    else:
+        if post_process is not None:
+            raise ValueError("Can't specify both `simplify` and `post_process`")
```
jeromekelleher, line 619: Needs a test

tsinfer/inference.py
```diff
@@ -1750,6 +1759,8 @@ def get_samples_tree_sequence(self, map_additional_sites=True):
             # root.
             assert left.shape[0] == 0
             root = tables.nodes.add_row(flags=0, time=tables.nodes.time.max() + 1)
+            ultimate = tables.nodes.add_row(flags=0, time=tables.nodes.time.max() + 2)
```
jeromekelleher, line 1762: No point in running max on tables.nodes.time twice

tsinfer/inference.py
```diff
@@ -640,7 +659,7 @@ def match_samples(
         # we sometimes assume they are in the same order as in the file
 
     manager.match_samples(sample_indexes, sample_times)
-    ts = manager.finalise(simplify=simplify)
+    ts = manager.finalise(do_post_process=post_process, simplify_only=simplify_only)
```
jeromekelleher, line 662: I'm not sure there's much point in sending these parameters to the finalise method - could we just do
```
if post_process:
    ts = post_process(...)
return ts
```

jeromekelleher (commented):
LGTM, spotted a few final things.

Need to be careful with our updated requirement on tskit getting carried through properly to the final packages.

tsinfer/inference.py
hyanwong: Oh yes, good to have the extra eyes on this. Although it's a nice idea, it does create a variable name conflict between the `post_process` parameter and the identically named function. Do we want to rename anything to get around this?

hyanwong: I can work around it by defining `_post_process` as an internal alias for the `post_process()` function, but it's a bit messy.

jeromekelleher (approved):
LGTM let's merge. Should probably wait for the JupyterBook one to land first though

jeromekelleher:
Some conflicts to fix @hyanwong , we can auto merge then

hyanwong:
> Some conflicts to fix @hyanwong , we can auto merge then

Oh yes, sorry - didn't see this as I was checking on mobile. Should be OK now. How do I auto merge?

Issue #694, opened 2022-07-10
Replace `permissive_json_schema` with `tskit.MetadataSchema.permissive_json()`


hyanwong:
If I assign metadata to individuals using `tskit.MetadataSchema.permissive_json()`, tsinfer dies when trying to extend the schema using [`add_to_schema`](https://github.com/tskit-dev/tsinfer/blob/b87fbea7ff423e6480574f829858a4d84e96aa9e/tsinfer/inference.py#L74)

```
import msprime
import tskit
import tsinfer

nm = ["a", "b", "c", "d", "e"]
ts = msprime.sim_ancestry(len(nm), ploidy=1, sequence_length=100, random_seed=1)
ts = msprime.sim_mutations(ts, rate=0.1, random_seed=1)
tables = ts.dump_tables()
tables.individuals.metadata_schema = tskit.MetadataSchema.permissive_json()
for i, n in enumerate(nm):
   tables.individuals[i] = tables.individuals[i].replace(metadata={"name": n})
sd = tsinfer.SampleData.from_tree_sequence(tables.tree_sequence())
tsinfer.infer(sd)
```

```
---> 76     if name in schema["properties"]:
     77         raise ValueError(f"The metadata {name} is reserved for use by tsinfer")
     78     if definition is None:

KeyError: 'properties'
```

hyanwong:
The reason is that this is how a permissive schema is defined in tsinfer, and it doesn't match the tskit version:

https://github.com/tskit-dev/tsinfer/blob/b87fbea7ff423e6480574f829858a4d84e96aa9e/tsinfer/formats.py#L66

benjeffery:
The error is that line 76 above should be:
```python
try:
    if name in schema['properties']
       raise ValueError(f"The metadata {name} is reserved for use by tsinfer")
except KeyError:
    pass

As schemas don't have to have properties defined.

**Comment by hyanwong** [comment]
Thanks Ben. The formats.py file says:

```
def permissive_json_schema():
    # A version of this should be in tskit, but we keep it here for now
    # for convenience
    return {
        "codec": "json",
        "type": "object",
        "properties": {},
        "required": [],
        "additionalProperties": True,
    }
```

I'm not sure what the rationale for a different sort of permissive_json_schema in tskit vs tsinfer is? I guess we are happy to leave this here for now, anyway.

benjeffery:
I think that was put in place before tskit had that schema. It should be replaced with the tskit one.

[...]

PR #695, opened 2022-07-11
Fix adding to metadata schema if no properties key

hyanwong:
See https://github.com/tskit-dev/tsinfer/issues/694#issuecomment-1180208186


hyanwong:
incorporated into #687

PR #697, opened 2022-07-12
Remove an unused parameter from make_ancestors_ts

hyanwong:
make_ancestors_ts is not exposed, but it's still confusing to have a param that is never used.


benjeffery (approved):
LGTM if this is an undocumented, internal function. It has a docstring, but I can't find it in the tsinfer manual, so I assume it is undoc'd and this change isn't breaking?
Can you confirm @hyanwong?

hyanwong:
It is deliberately undocumented and used only internally. Shing is using it as a hack, but I've talked to him about that.

PR #700, opened 2022-08-09
Use new doc theme

benjeffery:


benjeffery:
Hmm, not working, I wonder if we have to upgrade to a jupyterbook to use the theme.

benjeffery:
Looks like we can only use the theme once we convert to jupyterbook.

hyanwong:
It would be fab to convert this to JB anyway. Number 4 on your to-do?

PR #710, opened 2022-09-21
Move stabilise_node_ordering into eval function

### LINKED ISSUE #709: Remove stabilise_node_ordering option from API
It's leading to complicated code which we hardly ever need. Bring the things it does into a function defined in eval_util (probably)

hyanwong:
Fixes #709. I guess we don't need to add this to the changelog as it was never documented in the first place?


jeromekelleher (approved):
Perfect - much better!

PR #717, opened 2022-10-08
Switch to the default tskit permissive JSON schema

### LINKED ISSUE #694: Replace `permissive_json_schema` with `tskit.MetadataSchema.permissive_json()`

hyanwong:
Fixes #694. Note that the error type has now changed from `ValueError` to `KeyError` when trying to access a badly formed struct schema, here:

https://github.com/tskit-dev/tsinfer/blob/7079672f91e4e4306ef0a644016ce041c3ca9e97/tests/test_formats.py#L1282

we now get 

```
    def required_validator(validator, required, instance, schema):
        # Do the normal validation
        yield from jsonschema._validators.required(validator, required, instance, schema)
    
        # For struct codec if a property is not required, then it must have a default
>       for prop, sub_schema in instance["properties"].items():
E       KeyError: 'properties'
```

I guess that's fine, though?

### DIFF
--- a/tests/test_inference.py
+++ b/tests/test_inference.py
@@ -36,6 +36,7 @@ import numpy as np
 import pytest
 import tskit
 import tsutil
+from tskit import MetadataSchema
 
 import _tsinfer
 import tsinfer
@@ -879,7 +880,7 @@ class TestMetadataRoundTrip:
 
     def test_ts_metadata(self):
         metadata = {f"x_{j}": j for j in range(10)}
-        schema = tsinfer.permissive_json_schema()
+        schema = tskit.MetadataSchema.permissive_json().schema
         for j in range(10):
             name = f"x_{j}"
             metadata[name] = j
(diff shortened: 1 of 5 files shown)


tests/test_inference.py
```diff
@@ -4156,41 +4144,39 @@ def test_sample_too_old(self):
 
 class TestAddToSchema:
     def test_is_copy(self):
-        schema = tsinfer.permissive_json_schema()
```
jeromekelleher, line 4159: A less noisy option would have been to change the definition of ``permissive_json_schema`` to return tskit's version...

jeromekelleher (approved):
LGTM - I don't think there's any real material change?

jeromekelleher:
@benjeffery - any thoughts on the type change on error mentioned above?

benjeffery:
> @benjeffery - any thoughts on the type change on error mentioned above?

What does the schema that is generating that error look like? It's not ideal and probably the tskit code shouldn't be assuming that `properties` exists on that object.

tests/test_inference.py
hyanwong: Yes, but it felt like leaving cruft in the tsinfer codebase, which could have been stripped out.

hyanwong:
> What does the schema that is generating that error look like? It's not ideal and probably the tskit code shouldn't be assuming that `properties` exists on that object.

We're simply doing the equivalent of this:

```python
bad_schema = tskit.MetadataSchema.permissive_json().schema  #  simply {'codec': 'json'}
bad_schema["codec"] = "struct"
tables.populations.metadata_schema = tskit.MetadataSchema(bad_schema)
```

benjeffery:
I think the error message should be better from tskit here, and also be a `MetadataSchemaValidationError`. I'll raise an issue over there.

benjeffery:
https://github.com/tskit-dev/tskit/issues/2601

hyanwong:
> I think the error message should be better from tskit here, and also be a `MetadataSchemaValidationError`. I'll raise an issue over there.

So do we want to check for either a KeyError or a MetadataSchemaValidationError in the tests in this PR, so that it doesn't break on the next tskit release, or shall we just fix it when it does break in the future?

benjeffery:
Just catch any error in the test, then add an issue to make it specific after the next tskit release.

jeromekelleher (approved):

PR #718, opened 2022-10-08
Allow ancestral allele to be set

### LINKED ISSUE #686: Add an "ancestral_state_index" parameter to SampleData.add_sites()?
At the moment we require people to order their allele lists such that the first allele is the ancestral state. This can be a little fiddly, and often the ancestral state isn't known anyway, so it would be good to have a way for users to indicate this, while keeping the site in the final TS. Currently the way to incorporate a site with an unknown ancestral state is to pick one at random as the first one to put, and include the site in the `exclude_positions` parameter of `generate_ancestors()`, which seems pretty obscure to me.

I suggest that a non-breaking change that could allow this would be to have an extra parameter "ancestral_state_index", which defaults to 0, meaning that the zeroth allele in the `alleles` parameter is treated as the ancestral state. This would be similar to the sort of format we expect to see when we switch to using sgkit as the input format. We could allow `tskit.NULL` to be passed to mean that we don't know the ancestral state, and it should be placed using parsimony.

This seems pretty intuitive to me, but there are some decisions to be made here. When we add a site with (say) `ancestral_state_index=1`, do we record the genotypes and allele list before sticking them into the sample data file (which seems a bit hacky, and is not likely to be portable to sgkit). Or do we switch the genotypes when we build the ancestors, in which case we will need to swap the genotypes when we match against the ancestors using `match_samples` or `match_ancestors`. If the latter, I suspect that it might be helpful for sanity checking purposes to store the meaning of the genotype values (i.e. the reordered alleles list) in the AncestorData instance.

hyanwong:
Fixes #686 by altering the genotypes returned by the variants and haplotypes iterator on the fly. Probably worth testing the performance implications. Tests not complete, obviously.

I think the main slowness will be the double `np.where` call when getting haplotypes, which is called for each haplotype returned. But it may be that getting the sample haplotypes is not a time-critical stage of the inference. And we'll probably need to do something like this with `sgkit` storage anyway.

```
    def __all_haplotypes(self, sites=None):
        ...
        for j in range(self.num_samples):
            if j % chunk_size == 0:
                chunk = self.sites_genotypes[:, j : j + chunk_size].T
            a = chunk[j % chunk_size]
            # Remap the genotypes at all sites, depending on the ancestral_allele index    
            a = np.where(a==aa_indexes, 0, np.where(
                np.logical_and(a != MISSING_DATA, a < aa_indexes), a + 1, a
            ))
            yield j, a if sites is None else a[sites]
```


hyanwong:
Still needs to add the feature of excluding sites from inference if the ancestral state is MISSING_DATA, and tweaking the parsimony function to deduce the most parsimonious ancestral state, but those are both easy fixes once the general approach is deemed OK.

tsinfer/formats.py
```diff
@@ -1778,17 +1820,23 @@ def add_site(
 
         if alleles is None:
             alleles = ["0", "1"]
+        if ancestral_state is None:
+            ancestral_allele_index = 0
+        elif ancestral_state == MISSING_DATA:
+            ancestral_allele_index = MISSING_DATA
+        else:
+            ancestral_allele_index = list(alleles).index(ancestral_state)
```
jeromekelleher, line 1828: What's the conversion to ``list`` for here? I think tuples also support the ``index`` operation.

tsinfer/formats.py
```diff
@@ -1971,14 +2020,22 @@ def sites(self, ids=None):
         alleles = self.sites_alleles[:]
         metadata = self.sites_metadata[:]
         time = self.sites_time[:]
+        ancestral_allele_index = self.sites_ancestral_allele_index[:]
         if ids is None:
             ids = np.arange(0, self.num_sites, dtype=int)
         for j in ids:
+            aa = ancestral_allele_index[j]
```
jeromekelleher, line 2027: I don't think we should reorder the alleles here, let's just keep them in the order specified by the user.

tsinfer/formats.py
```diff
@@ -1971,14 +2020,22 @@ def sites(self, ids=None):
...
+            al_list = alleles[j]
+            if aa > 0:
+                # reorder the alleles to put the ancestral one first
+                al_list = tuple([al_list[aa]] + al_list[:aa] + al_list[aa + 1 :])
             site = Site(
                 id=j,
                 position=position[j],
-                ancestral_state=alleles[j][0],
-                alleles=tuple(alleles[j]),
+                ancestral_state=(
```
jeromekelleher, line 2035: I don't think giving the ancestral state back here helps much. Let's just return the straight ``ancestral_allele_index`` that we store.

tsinfer/formats.py
```diff
@@ -2007,30 +2064,69 @@ def num_alleles(self, sites=None):
     def variants(self, sites=None):
```
jeromekelleher, line 2064: Let's add a ``recode_ancestral=True`` option here. If False, we just return the alleles/genotypes exactly as they are.

tsinfer/formats.py
```diff
@@ -2007,30 +2064,69 @@ def num_alleles(self, sites=None):
...
+        :param array sites: A numpy array of ascending site ids for which to return
+            data. If None (default) return all sites.
+        :return: An interator over the variants in the sample data file.
+        :rtype: iter(Variant)
+
         """
         all_genotypes = chunk_iterator(self.sites_genotypes, indexes=sites)
-        for genotypes, site in zip(all_genotypes, self.sites(ids=sites)):
-            yield Variant(site=site, alleles=site.alleles, genotypes=genotypes)
+        ancestral_allele_index = self.sites_ancestral_allele_index[:]
```
jeromekelleher, line 2080: Don't need to do this, as the ``site`` returns ``ancestral_allele_index``.

Within the loop:
```python
alleles = site.alleles
if recode_ancestral:
    alleles = site.remap_alleles()
    genotypes = site.remap_genotypes(genotypes)
# NOTE: if recode_ancestral is True then variant.alleles and site.alleles may differ.
yield Variant(site=site, alleles=alleles, genotypes=genotypes)
```

tsinfer/formats.py
```diff
@@ -2007,30 +2064,69 @@ def num_alleles(self, sites=None):
...
+            if aa_index > 0:
+                # re-map the genotypes
+                geno_map = np.arange(len(site.alleles) + 1)
+                geno_map[MISSING_DATA] = MISSING_DATA
+                geno_map[aa_index] = 0
+                geno_map[0:aa_index] += 1
+                genos = geno_map[genos]
+            yield Variant(site=site, alleles=site.alleles, genotypes=genos)
 
     def __all_haplotypes(self, sites=None):
```
jeromekelleher, line 2097: Also add ``remap_ancestral`` option here.

tsinfer/formats.py
```diff
@@ -2007,30 +2064,69 @@ def num_alleles(self, sites=None):
...
         # the chunk iterator to handle this.
+        aa_indexes = self.sites_ancestral_allele_index[:]
+        # If ancestral allele is missing, keep the order unchanged (aa_index of zero)
+        aa_indexes[aa_indexes == MISSING_DATA] = 0
         chunk_size = self.sites_genotypes.chunks[1]
         for j in range(self.num_samples):
             if j % chunk_size == 0:
                 chunk = self.sites_genotypes[:, j : j + chunk_size].T
             a = chunk[j % chunk_size]
+            # Remap the genotypes at all sites, depending on the ancestral_allele index
```
jeromekelleher, line 2108: Nice - not totally following what's going on, but can't see this being a bottleneck.

tsinfer/formats.py
```diff
@@ -2007,30 +2064,69 @@ def num_alleles(self, sites=None):
...
+            ancestral_allele_index = ancestral_allele_index[np.array(sites)]
+        for genos, site, aa_index in zip(
+            all_genotypes,
+            self.sites(ids=sites),
+            ancestral_allele_index,
+        ):
+            if aa_index > 0:
+                # re-map the genotypes
+                geno_map = np.arange(len(site.alleles) + 1)
+                geno_map[MISSING_DATA] = MISSING_DATA
```
jeromekelleher, line 2091: I think this depends on MISSING_DATE being -1, right? Might be good to drop a comment saying that.

tsinfer/formats.py
```diff
@@ -1699,6 +1733,7 @@ def add_site(
         metadata=None,
         inference=None,
         time=None,
+        ancestral_state=None,
```
jeromekelleher, line 1736: I'm not sure tskit.MISSING_DATA is helpful here as an API, as we're mixing strings and integers on input. What if we had a special argument value "unknown" here to mark it? It would make for more readable code I think,
```python
sd.add_site(alleles="ACGT", genotypes=[0, 0, 0, 1], ancestral_state="unknown")
```

since we're not actually storing ``ancestral_state`` there's no advantage to using -1, and (I think) it makes sense to keep the types of the parameter clear.

jeromekelleher (commented):
LGTM, some suggestions for changing around the API.

tsinfer/formats.py
hyanwong: Happy to convert to tuple instead. Since `alleles` is passed in, it could be e.g. a numpy array, so I wanted to ensure art was something with an `index` operator.

tsinfer/formats.py
hyanwong: OK. This was to match the tskit `Site` object, though.

tsinfer/formats.py
hyanwong: Yes, good call. I was going to suggest this as an update too.

tsinfer/formats.py
hyanwong: Yes, good point

tsinfer/formats.py
hyanwong: I'm worried that people will use this allele list to index into the genotypes array (it's confusing that you can get the alleles using `variant.site.alleles` and `variant.alleles`. Perhaps the site should return a `set()` of allelic states, like `ts.site(0).alleles` does? In this case I think we do want to return the actual value for the ancestral_allele, rather than the index, as you can't us it to index into a set.

jeromekelleher: That's OK though, isnt it, as we won't reorder the genotypes unless specifically asked to (easy to update internal methods to do this)

hyanwong: I thought we were setting `recode_ancestral=True` by default? But I can see that we might want *not* to do that by default, and set it explicitly to `True` in the tsinfer internal routines.

jeromekelleher: Let's set recode_ancestral=False by default, probably no good reason to do this outside of tsinfer.

tsinfer/formats.py
hyanwong: I could even put an assert in, actually, then it's self documenting.

tsinfer/formats.py
hyanwong: Hmm, isn't it a bit ugly (e.g. what if the allele actually is the string "unknown"!). I would prefer something that can't be interpreted as a string. None would be an obvious one, but we are using that as the default to mean `0`.

Alternatively we could ask for the ancestral state index, and disallow negative values, other than `-1`, but then (as you pointed out above, I think), the user might reasonably expect `-1` to mean the last allele.

Finally, we could ask the user to add `None` to the list of alleles, and indicate that for the index, but that seems a bit convoluted to me.

tsinfer/formats.py
hyanwong: Agree.

tsinfer/formats.py
jeromekelleher: Hah, good point. Hmm, what if we did something like
```python
ANCESTRAL_STATE_IS_FIRST = object() # Standard singleton pattern
...
def add_site(self, ..., ancestral_state=ANCESTRAL_STATE_IS_FIRST):
    if ancestral_state is ANCESTRAL_STATE_IS_FIRST:
        ancestral_state_index = 0
    elif ancestral_state is None:
        ancestral_state_index = -1 # Unknown
    else:
         ancestral_state_index = alleles.index(ancestral_state)
```

The idea is that we're only really supporting old code using the default case, so we can use ``None`` in a more useful way like this.

Thoughts?

hyanwong: If we don't mind having something other than `None` as the default, then it might be simpler to have the parameter being the `ancestral_state_index` (default `0`), and have `None` to mean unknown? Then that matches exactly what we store in the SampleData file anyway.

hyanwong: Also, if we want a shorter name, rather than use the term `ancestral_state_index`, we could use `ancestral_allele` to mean the allele index, and reserve the term `ancestral_state` (which matches the tskit terminology) to mean the `str` value?

hyanwong:
Needs to bump minor format number, and test older files somehow

tsinfer/formats.py
jeromekelleher: I like ``ancestral_allele`` - we generally store indexes and don't put the name on the variable, and this avoids confusion with the tskit 'state'. So let's do that.

I think specifying "ancestral_state" as a string is probably the easiest for users, but then maybe we're over thinking this as the plan is to remove this step of the workflow (add_site) ASAP.

So: let's make the API this:
```python
def add_site(self, ..., ancestral_allele=None):
     if ancestral_allele is None:
          ancestral_allele = 0
     # do the stuff.
```

We can always add another mutually exclusive parameter ``ancestral_state`` later if we want it.

hyanwong: OK, great. A minor point though: I thought your idea was to have the default as (explicitly) "0", and to have `None` mean "no ancestral allele known". So:

```python
def add_site(self, ..., ancestral_allele=0):
     if ancestral_allele is None:
          # store tskit.MISSING_DATA internally
     # do the stuff.
```

jeromekelleher: Having ``ancestral_allele=-1`` as missing data seems most logical now that we're making it an integer? It's good to reserve ``None`` as "I didn't specify anything". Otherwise, we won't know if the user didn't specify anything for ancestral_allele or whether they specifically said "it must be 0".

jeromekelleher: The tskit.MISSING_DATA isn't particularly helpful here I think - sgkit also uses -1 as missing data, and that's a general rule that a user is going to need to understand.

hyanwong: As long as we are happy with `-1` not being used by the user to mean the last allele in the list. I'm not sure how we would catch that.

jeromekelleher: That's just a general gotcha, can't be helped.

tests/test_formats.py
```diff
@@ -2593,3 +2618,16 @@ class TestBufferedItemWriterThreads2(BufferedItemWriterMixin):
...
     num_threads = 20
+
+
+class TestOldFormats:
+    """
+    Test that we can read old formats
+    """
+
+    def test_read_sd_format_5_0(self):
+        # generated by tsinfer 0.2.3
```
jeromekelleher, line 2629: nice

tsinfer/formats.py
```diff
@@ -1197,8 +1210,29 @@ def sites_time(self, value):
 
     @property
     def sites_alleles(self):
+        """
+        The "raw" alleles list for each site, as passed in when adding sites. Note that
+        if an ancestral allele is defined, it need not be the zeroth element. The
+        allele list for a site may therefore differ in order from that returned by the
```
jeromekelleher, line 1216: The last sentence is incorrect isn't it? Also it should be the variants() method we're talking about, not sites I think?

tsinfer/formats.py
```diff
@@ -2004,33 +2058,92 @@ def num_alleles(self, sites=None):
                 num_alleles[j] -= 1
         return num_alleles[sites]
 
-    def variants(self, sites=None):
+    def variants(self, sites=None, recode_ancestral=None):
         """
         Returns an iterator over the Variant objects. This is equivalent to
-        the TreeSequence.variants iterator.
-        """
+        the TreeSequence.variants iterator. The alleles are guaranteed to be
```
jeromekelleher, line 2064: Missing a ``If the ``recode_ancestral`` parameter is True, the ..." here

jeromekelleher: Less than clear what happens when there is no ancestral allele defined here. Would help to rephrase the main description of the parameter as a paragraph in the body of the docstring, and reserve the param string for very basic stuff.

tsinfer/formats.py
```diff
@@ -2004,33 +2058,92 @@ def num_alleles(self, sites=None):
...
+        returned in an order such that the ancestral state (if know) is the
+        first item in the list (and hence a ``0`` in the genotypes array) corresponds
+        to a sample with the ancestral state. This may mean that the indexes in the
+        genotypes array are recoded compared to the array passed in to :meth:`add_site`.
+
+        :param array sites: A numpy array of ascending site ids for which to return
+            data. If None (default) return all sites.
+        :param bool recode_ancestral: Should we recode the genotypes so that the
```
jeromekelleher, line 2072: i.e., "If True recode genotypes so that the ancestral state is coded as 0 as described above. Otherwise return genotypes in the input allele encoding"

tsinfer/formats.py
```diff
@@ -2004,33 +2058,92 @@ def num_alleles(self, sites=None):
...
+            ancestral state (if it is defined) has a genotype of ``0``, and is
+            listed first in the alleles list. Default: ``None`` treated as ``False``.
+        :return: An interator over the variants in the sample data file.
+        :rtype: iter(Variant)
+
+        .. warning::
```
jeromekelleher, line 2078: I think this should go above the param and return block (conventionally at the end of a doscstring)

tsinfer/formats.py
```diff
@@ -2004,33 +2058,92 @@ def num_alleles(self, sites=None):
...
+                )
             yield j, a if sites is None else a[sites]
 
-    def haplotypes(self, samples=None, sites=None):
+    def haplotypes(self, samples=None, sites=None, recode_ancestral=None):
         """
-        Returns an iterator over the (sample_id, haplotype) pairs.
+        Returns an iterator over the (sample_id, haplotype) pairs. Each haplotype is
+        an array of indexes, where the ``i``th value is an index into the
+        ``site.alleles`` list for the ``i`` specified site. A value of ``0`` in the
```
jeromekelleher, line 2128: Same comments as docstring for variants above. Could just do a "See variants for details" to avoid duplication.

jeromekelleher (commented):
LGTM, just some documentation cleanups needed

tsinfer/formats.py
hyanwong: Yes, sorry - still on the final pass to check through docs and add extra tests

hyanwong:
It turns out that in 2 places in the inference step (e.g. when placing mutations onto the TS) require knowledge of the order of alleles at the site without using the variants, so I've put that as a separate (calculated) attribute of the `Site` class.

Also, I can't get 

```
:rtype: iter(int, numpy.ndarray)
```

to work in sphinx for the return value of the `.haplotypes` method, so I've just nixed it for the moment.

jeromekelleher:
Ready for final review then?

hyanwong:
> Ready for final review then?


> Ready for final review then?

Yes, I just added `item` to the nitpicks, as suggested by Ben.

Untitled.ipynb
```diff
@@ -0,0 +1,712 @@
+{
```
jeromekelleher, line 1: File added

tests/test_inference.py
```diff
@@ -740,12 +740,15 @@ def setup_class(cls):
...
-    def verify(self, genotypes):
+    def verify(self, genotypes, anc_alleles=None):
         genotypes = np.array(genotypes, dtype=np.int8)
         m = genotypes.shape[0]
+        if anc_alleles is None:
+            anc_alleles = np.zeros(m, dtype=np.int8)
         with tsinfer.SampleData(sequence_length=m + 1) as sample_data:
             for j in range(m):
-                sample_data.add_site(j, genotypes[j])
+                print("Verify:", j, genotypes[j], anc_alleles[j])
```
jeromekelleher, line 750: stray print

tests/test_inference.py
```diff
@@ -4210,6 +4231,23 @@ def test_sample_too_old(self):
             tsinfer.match_samples(sd_copy, ancestors_ts, force_sample_times=True)
 
 
+class TestAncestralAlleles:
+    def test_recoded_equivalents(self):
+        with tsinfer.SampleData(sequence_length=1) as sd:
+            sd.add_site(0, [1, 1, 0], alleles=("A", "T"), ancestral_allele=0)
+        print(list(sd.variants(recode_ancestral=True)))
```
jeromekelleher, line 4238: Lots of prints

tsinfer/formats.py
```diff
@@ -765,11 +766,30 @@ class Site:
...
     time = attr.ib()
     alleles = attr.ib()
 
+    @property
+    def ancestral_state(self):
+        if self.ancestral_allele == MISSING_DATA:
+            return None
+        return self.alleles[self.ancestral_allele]
+
+    @property
```
jeromekelleher, line 780: I think this is a function rather than a property. ``site.reorder_alleles()`` or something?

tsinfer/inference.py
```diff
@@ -1296,14 +1305,16 @@ def convert_inference_mutations(self, tables):
...
-                ancestral_state=site.alleles[0],
+                ancestral_state=site.ancestral_state,
                 metadata=metadata,
             )
             while mutation_id < num_mutations and mut_site[mutation_id] == site_id:
                 tables.mutations.add_row(
                     site_id,
                     node=node[mutation_id],
-                    derived_state=site.alleles[derived_state[mutation_id]],
+                    derived_state=site.alleles_ancestral_first[
```
jeromekelleher, line 1315: Yes, it would be clearer that we'd doing computation in the background there if we used a function on site for getting the ordered alleles.

jeromekelleher (commented):
Looks good, but a few things need cleanup.

It's not clear to me that we want to start documenting Variant and Site and what needs review in terms of compatibility now.

hyanwong:
> It's not clear to me that we want to start documenting Variant and Site and what needs review in terms of compatibility now.

No, this was simply to get the docs to pass CI, so that the return types were being documented correctly. I guess I can simply remove the documentation of return type from the `sd.variants()` iterator? Or exclude documenting `variants()`z at all.

jeromekelleher (approved):

jeromekelleher:
Makes sense. Let's not worry about it, probably doesn't matter. Thanks for getting this done @hyanwong, this is the first important piece in our new data input puzzle!

PR #720, opened 2022-10-09
Delete topology in flanking regions

### LINKED ISSUE #483: Chop off edges before first site and after last site
Material before the first site and after the last site is uninformative so we should probably prevent edges from spanning those regions. This can be an issue for downstream analyses when telomeres are large (in practice we manually remove these regions with `tskit.keep_intervals` anyway).
Making this automatic in `tsinfer` involve minimal changes and prevent downstream problems

hyanwong:
Fixes #483. I have added an option to `post_process` not to do this, partially because it makes testing easier. But there's no way to skip the flanking topology deletion unless you call post_process explicitly.

### DIFF
--- a/tsinfer/inference.py
+++ b/tsinfer/inference.py
@@ -1985,29 +1985,35 @@ def post_process(
     ts,
     *,
     split_mrca=None,
+    erase_flanks=None,
     # Parameters below deliberately undocumented
     warn_if_unexpected_format=None,
     simplify_only=None,
 ):
     """
-    post_process(ts, *, split_mrca=None)
+    post_process(ts, *, split_mrca=None, erase_flanks=None)
 
     Post-process a tsinferred tree sequence into a more conventional form. This is
     the function run by default on the final tree sequence output by
-    :func:`match_samples`. It involves the following 3 steps:
+    :func:`match_samples`. It involves the following 4 steps:
 
     #. If the oldest node is connected to a single child via an edge that spans the
        entire tree sequence, this oldest node is removed, so that its child becomes
        the new root (this step is undertaken to remove the "virtual-root-like node"
        which is added to ancestor tree sequences to enable matching).
     #. If the oldest node is removed in the first step and the new root spans the
-       entire genome, it is treated as the "grand MRCA" and (if split_mrca is ``True``)
-       the node is split into multiple coexisiting nodes with the splits
+       entire genome, it is treated as the "grand MRCA" and (unless split_mrca is
+       ``False``) the node is split into multiple coexisiting nodes with the splits
        occurring whenever the children of the grand MRCA change. The rationale
        is that tsinfer creates a grand MRCA consisting of a single ancestral haplotype
        with all inference sites in the ancestral state: this is, however, unlikely
        to represent a single ancestor in the past. If nodes in the tree sequence are
        then dated, these MRCA nodes can be pushed to different times.
+    #. Often, extensive regions of genome exist before the first defined site and after
+       the last defined site. Since no data exists in these sections of the genome, post
+       processing by default erases the inferred topology in these regions. However,
+       if ``erase_flanks`` is False, the flanking regions at the start and end will be
+       assigned the same topology as inferred at the first and last site respectively.
     #. The sample nodes are reordered such that they are the first nodes listed in the
        node table,  removing tree nodes and edges that are not on a path between the
        root and any of the samples (by applying the :meth:`~tskit.TreeSequence.simplify`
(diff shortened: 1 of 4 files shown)


tsinfer/inference.py
```diff
@@ -1950,29 +1950,35 @@ def post_process(
...
     *,
     split_mrca=None,
+    keep_flanks=None,
     # Parameters below deliberately undocumented
     warn_if_unexpected_format=None,
     simplify_only=None,
 ):
     """
-    post_process(ts, *, split_mrca=None)
+    post_process(ts, *, split_mrca=None, keep_flanks=None)
```
jeromekelleher, line 1959: It's a bit odd having positive and negative things done by default here (i.e, you do ``split_mrca`` by default but you don't ``keep_flanks``). It would be easier to remember if these were all positive statements which are things you do by default, so maybe rename to ``trim_flanks`` instead?

tsinfer/inference.py
```diff
@@ -1982,6 +1988,9 @@ def post_process(
     :param bool split_mrca: If ``True`` (default) and the oldest node is the only
         parent to a single "grand MRCA", split the grand MRCA into separate nodes
         (see above). If ``False`` do not attempt to identify or split a grand MRCA.
+    :param bool keep_flanks: If ``True``, output regions of topology inferred before
+        the first site and after the last site. Otherwise (default) keep only the
+        inferred topology between the first and last sites.
     :return: The post-processed tree sequence.
     :rtype: tskit.TreeSequence
     """
```
jeromekelleher, line 1996: Missing default assignment for ``trim_flanks`` here, ``trim_flanks = True if trim_flanks is None else trim_flanks``.

tsinfer/inference.py
```diff
@@ -2011,6 +2020,17 @@ def post_process(
                 "Cannot find a virtual-root-like ancestor during preprocessing"
             )
 
+        if not keep_flanks and tables.sites.num_rows > 0:
+            logger.info("Removing topology in flanking regions with keep_intervals")
+            # So that the last site falls within a tree, we must add one to the
+            # site position (or simply extend to the end of the ts)
+            upper_cutoff = min(tables.sites.position[-1] + 1, tables.sequence_length)
```
jeromekelleher, line 2027: Need to mindful that creating these column arrays is not a constant time operation and we'll be working with very large tables. You've created two copies within a couple of lines here.

jeromekelleher (commented):
LGTM - I think we should change to ``trim_flanks`` though.

tsinfer/inference.py
hyanwong: Yes, this is a good point and one of the things I was wondering about. Do we want `split_mrca=None, trim_flanks=None` or the inverse (something like `keep_mrca=None, keep_flanks=None`). I guess the first is fine?

tsinfer/inference.py
hyanwong: Ah, very helpful comment, thanks. I thought that once we had the tables the arrays gave direct memory access, so weren't costly. In which case, what's the quickest way to find the last position? I also assumed that `tables.sites.num_rows` was the same cost as `ts.num_sites`, but I see that it might not be.

tsinfer/inference.py
jeromekelleher: Positive things that we do by default and we can optionally switch off is the simplest here I think.

tsinfer/inference.py
jeromekelleher: Currently they create a new copy of the array each time you access. That's what ``ts.sites_position`` is for - **that** is a cheap view on the underlying memory.

Just store the array in a variable, ``pos = tables.sites.position`` to avoid needlessly recreating it.

hyanwong: Ah, I thought `ts.sites_position` and tables.sites.position were equivalent. Thanks for the heads-up

tsinfer/inference.py
hyanwong: I think we shouldn't use the word "trim" though, as the "normal" tskit trimming functions `.ltrim` and `.rtrim` actually change the coordinate system. Here we are simply removing topology. So perhaps "delete_flanks" or "erase_flanks" or "omit_flanks"? Do you have any preference @jeromekelleher ?

tsinfer/inference.py
hyanwong: No need to create the array anyway, I just need the first and last position, and I have the tree sequence, so I can do `ts.sites_position[0]` and `ts.sites_position[-1]` (or `ts.site(0).position`, etc I guess)

tsinfer/inference.py
hyanwong: I've switched to `erase_flanks` as I think erasing gives the right mental picture.

tsinfer/inference.py
jeromekelleher: As a general rule so, if you have the ts it's *much* better to use one of those arrays than the equivalent table column.

hyanwong: Useful to know, thanks. This is ready to merge, so.

tsinfer/inference.py
```diff
@@ -1982,11 +1988,17 @@ def post_process(
...
+    :param bool erase_flanks: If ``True`` (default), keep only the
+        inferred topology between the first and last sites. If ``False``,
+        output regions of topology inferred before the first site and after
+        the last site.
     :return: The post-processed tree sequence.
     :rtype: tskit.TreeSequence
     """
     if split_mrca is None:
         split_mrca = True
+    if erase_flanks is None:
```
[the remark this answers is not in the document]
hyanwong: But it's not doing `trim`: the `trim` method actually removes that region and shifts the coordinates. We don't want to do that.

jeromekelleher (approved):

PR #722, opened 2022-10-10
Fix doc build and prevent failing build from succeeding.

benjeffery:


benjeffery:
Merging as this is a small change to the docs and I need it on `main` to proceed with work on `tskit-site`.

PR #725, opened 2022-10-10
Update cli to print ts

### LINKED ISSUE #124: tsinfer ls on tree sequence does not include 'nodes'

hyanwong:
Fixes #124. Not sure if this even needs a test?


hyanwong:
I removed the path, as this will have been passed in to the CLI command anyway, so it's not unknown info

jeromekelleher (approved):

Issue #721, opened 2022-10-10
Update CLI to use use ``post_process`` rather than simplify


jeromekelleher:
Need to consider the compatibility issues for this.

jeromekelleher:
We can probably map this directly to a new ``no--postprocess`` and say the old ``--no-simplify`` is a deprecated for this.

PR #727, opened 2022-10-10
Add --no-post-process as an alias

hyanwong:
This is the simplest way to fix https://github.com/tskit-dev/tsinfer/issues/721, although it doesn't tell the user that the `--no-simplify` option is deprecated. However, it does save hassle working out what to do when conflicting options are passed, etc.


jeromekelleher (approved):
Nice!

PR #731, opened 2022-10-11
Add recombination and mismatch params to cli

### LINKED ISSUE #435: Add mismatch and recombination_rate arguments to the command line
The CLI currently does not allow control of these parameters.

hyanwong:
Fixes #435

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -119,7 +119,7 @@ class TestCli(unittest.TestCase):
         )
         self.output_trees = str(pathlib.Path(self.tempdir.name, "input-data.trees"))
         self.input_ts = msprime.simulate(
-            10, mutation_rate=10, recombination_rate=10, random_seed=10
+            10, length=10, mutation_rate=1, recombination_rate=1, random_seed=10
         )
         sample_data = tsinfer.SampleData(
             sequence_length=self.input_ts.sequence_length, path=self.sample_file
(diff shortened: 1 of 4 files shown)


jeromekelleher (commented):
Very nice, clean and easy. 

I guess we need a dependency on msprime now, which we should raise as an issue to make sure we track the dependencies in (e.g.) the conda package.

hyanwong:
> Very nice, clean and easy.
> 
> I guess we need a dependency on msprime now, which we should raise as an issue to make sure we track the dependencies in (e.g.) the conda package.

It's already in CI-tests-complete and CI-tests-conda. Not in CI-tests-pip though, but presumably is pulled in automatically when installing from pip?

jeromekelleher:
> It's already in CI-tests-complete and CI-tests-conda. Not in CI-tests-pip though, but presumably is pulled in automatically when installing from pip?

This is for the actual packaging, not just tests. Need to be much more careful about adding run-time dependencies.

We just need to track it as an issue, no need to get into it here.

hyanwong:
https://github.com/tskit-dev/tsinfer/issues/736

hyanwong:
I've no idea how I can test this, because calling the CLI command in the test suite doesn't end up the the mismatch etc parameters dumped into the tree sequence. So I still have no way to check from the ts file generated within the test suite that the mismatch etc parameters were passed correctly.

jeromekelleher:
Mocking out some key functions is the right way I think, as this will tell you whether values were passed through in the expected way.

hyanwong:
OK, I used mocking. Thanks for the tip. I think this should be ready to merge

hyanwong:
Also fixes https://github.com/tskit-dev/tsinfer/issues/736

tests/test_cli.py
```diff
@@ -329,6 +329,106 @@ def test_chain(self):
...
+        command = [
+            "infer",
+            self.sample_file,
+            "--recombination_map",
+            ratemap,
+        ]
+        with mock.patch("tsinfer.infer") as infer:
+            self.run_command(command)
+            args, kwargs = infer.call_args
+            assert type(kwargs["recombination_rate"]) == msprime.RateMap
```
jeromekelleher, line 405:
```suggestion
            assert isinstance(kwargs["recombination_rate"]), msprime.RateMap)
```

jeromekelleher (approved):
LGTM, one minor tweak.

tests/test_cli.py
hyanwong: Ah yes, much better, thanks.

PR #738, opened 2022-10-14
Fix bug in missing ancestral states

hyanwong:
A silly typo ("ancestral_allele" -> "ancestral_state") revealed that the test suite wasn't comprehensive enough for missing ancestral states, and needs a bit of tweaking to be able to test this corner case.


hyanwong:
It would be good to merge this bugfix: the codebase change is minimal, although the test suite change is more involved, sadly. I have simply called `return` from `TestSparseAncestorsRoundTrip:: verify_data_round_trip` for the newly introduced case where we have unknown ancestral states (and therefore can't generate sties without placing some on using parsimony), but perhaps I should call some form of `skip` instead?

jeromekelleher:
The test suite changes take a bit of time to digest I'm afraid, looking at it now.

hyanwong:
Yes, sorry about that. It would probably be easier if we had #739 , but I though it better to fix ASAP.

jeromekelleher (approved):
LGTM

PR #741, opened 2022-10-17
Do not record underlying calls in provenance

### LINKED ISSUE #740: Disable low-level provenance recording when using CLI
We currently record provenance from API calls which aren't informative from the CLI. 

Just add ``record_provenance=False`` to API calls from CLI module when #558 is merged.

hyanwong:
Fixes #740  - also adds provenance recording to generate-ancestors CLI call

### DIFF
--- a/tsinfer/cli.py
+++ b/tsinfer/cli.py
@@ -167,6 +167,7 @@ def run_infer(args):
         num_threads=args.num_threads,
         recombination_rate=get_recombination_map(args),
         mismatch_ratio=args.mismatch_ratio,
+        record_provenance=False,
     )
     output_trees = get_output_trees_path(args.output_trees, args.samples)
     write_ts(ts, output_trees)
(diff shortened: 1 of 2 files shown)


tsinfer/cli.py
```diff
@@ -177,13 +178,19 @@ def run_generate_ancestors(args):
...
-    tsinfer.generate_ancestors(
+    ancestors = tsinfer.generate_ancestors(
         sample_data,
         progress_monitor=args.progress,
-        path=ancestors_path,
         num_flush_threads=args.num_flush_threads,
         num_threads=args.num_threads,
+        record_provenance=False,
     )
+    new_ancestors = ancestors.copy(path=ancestors_path)
```
jeromekelleher, line 188: This is a big cost just for provenance, and will lead to a huge pause after the progress bar is done. I don't think it's worth it.

Maybe just skip this one and open an issue to track? There's no useful options in generate ancestors to record anywy

jeromekelleher (commented):
Looks good, but the match-ancestors stuff is a big performance and usabilty regression.

tsinfer/cli.py
hyanwong: Good point: I hadn't thought about the perf issues of making a copy. I assume there's no way to add provenance to the existing ancestors file without copying?

As you say, not worth spending time over, if we are re-doing the pipeline anyway, so we can just open an issue to track.

jeromekelleher (approved):
LGTM

PR #748, opened 2022-10-20
Minimum viable sgkit dataset

benjeffery:
Round trips an sgkit dataset.


benjeffery:
@jeromekelleher I'm thinking about how to cope with the `uuid` method here. We originally thought to make this a read-only operation on the sgkit dataset, but the only way i can see to support `uuid` is to generate one and write it back to the dataset. Otherwise, the dataset will have a different uuid each time it is read.

jeromekelleher:
Let's get rid of UUID, it's pointless. We've never used it and it's based on an unrealistic idea that datasets get made once and never changed. 

So, just hack around it in whatever way is simplest in the short term.

benjeffery:
Ok I've removed `uuid` as that was the simplest way.

For ancestral allele - I'm thinking we provide a method on the dataset that takes either an array of ancestral allele _values_ and converts it to the needed index. Could also accept a sequence that the alleles are extracted from at the site's positions?

hyanwong:
> For ancestral allele - I'm thinking we provide a method on the dataset that takes either an array of ancestral allele _values_ and converts it to the needed index. Could also accept a sequence that the alleles are extracted from at the site's positions?

Both would be useful, as ancestral alleles are sometimes provided as a FASTA file by ensembl (e.g. http://ftp.ensembl.org/pub/release-108/fasta/ancestral_alleles/). I guess we would need to check that they were of the same length as the sequence length. The approach would be similar to the reference_sequence stuff, I guess, other than we don't have to store it.

benjeffery:
> I guess we would need to check that they were of the same length as the sequence length.

There is no sequence length stored in sgkit (makes sense as a VCF has no such concept. We should check that the sequence value at the site is one of the alleles though, right?

FASTA won't work for indels right? Can't get alleles from the FASTA for them without knowing their length.

benjeffery:
> Either way, we need some way to put the seq len into the new format, right?

At the moment the concept is read-only from sgkit. Any use case that requires changing the dataset will need to be thought through as where possible that should be an sgkit function/workflow unless it is very tsinfer specific.

hyanwong:
> > Either way, we need some way to put the seq len into the new format, right?
> 
> At the moment the concept is read-only from sgkit. Any use case that requires changing the dataset will need to be thought through as where possible that should be an sgkit function/workflow unless it is very tsinfer specific.

Right, so the question is whether SGkit should have `seqlen` stored somewhere, e.g. when read from a VCF line like

```
##contig=<ID=20,length=62435964,assembly=B36,md5=f126cdf8a6e0c7f379d618ff66beb2da,species="Homo sapiens",taxonomy=x>
```

(https://samtools.github.io/hts-specs/VCFv4.2.pdf). I guess at the moment that data is not stored in an SGkit data file, although is is returned by htslib/cyvcf2 (annoyingly undocumented, though: https://github.com/brentp/cyvcf2/commit/a138fadd1d2b0b0a58568a00c4c42b09312037c5)

tomwhite:
> Right, so the question is whether SGkit should have `seqlen` stored somewhere

We should change sgkit to store it if it's in the VCF. Some previous discussion here: https://github.com/pystatgen/sgkit/issues/464

jeromekelleher:
We should be able to derive VCF-spec defined assembly and contig information (including length) from the sgkit dataset, so let's follow that up as an upstream improvement.

tomwhite:
I've opened https://github.com/pystatgen/sgkit/pull/946 to add contig lengths to the dataset. Assembly information is not exposed by cyvcf2 so I haven't added that yet.

benjeffery:
Trying to sort out dependency issues here as sgkit has upper bounds on `pandas` and `cyvcf2` needs to be in the conda part for windows compatibility. Apart from that this is ready to review!

tsinfer/formats.py
```diff
@@ -703,7 +694,6 @@ def __str__(self):
             ("format_name", self.format_name),
             ("format_version", self.format_version),
             ("finalised", self.finalised),
-            ("uuid", self.uuid),
```
jeromekelleher, line 706: We should document this in the CHANGELOG I guess, but totally fine to day "this is gone, deal with it"

tsinfer/formats.py
```diff
@@ -2225,6 +2215,256 @@ def populations(self):
...
 
+class SgkitSampleData(SampleData):
+
+    FORMAT_NAME = "tsinfer-sgkit-sample-data"
+    FORMAT_VERSION = (0, 1)
+
+    def __init__(self, path):
+        self.path = path
+        self.data = zarr.open(path, mode="r")
+        self._num_sites, self._num_individuals, self.ploidy = self.data[
```
jeromekelleher, line 2226: Maybe
```
genotypes_arr = self.data["call_genotypes"]
self._num_sites, self._num_individuals, self.ploidy = genotypes_arr.shape
```

Currently line breaking is a bit eye hurty

tsinfer/formats.py
```diff
@@ -2225,6 +2215,256 @@ def populations(self):
...
+    def metadata_schema(self):
+        try:
+            return self.data.attrs["metadata_schema"]
+        except KeyError:
+            None
+
+    @property
+    def metadata(self):
+        try:
+            return self.data.attrs["metadata_schema"]
```
jeromekelleher, line 2339: Typo, should be "metadata"

tsinfer/formats.py
```diff
@@ -2225,6 +2215,256 @@ def populations(self):
...
+        return self.data["variant_position"]
+
+    @property
+    def sites_alleles(self):
+        return self.data["variant_allele"]
+
+    @property
+    def sites_ancestral_allele(self):
+        try:
+            return self.data["sites/ancestral_allele"]
```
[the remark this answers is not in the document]
benjeffery: I think it likely we end up with `variant_ancestral_allele_index` so will put that for now, but final decision will be part of #764

tsinfer/formats.py
benjeffery: https://github.com/tskit-dev/tsinfer/pull/748/commits/0ee3018cefa4d1ce2721b0b29ce7bc8006a22766

tsinfer/formats.py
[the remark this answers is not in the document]
benjeffery: https://github.com/tskit-dev/tsinfer/pull/748/commits/48af57f6d63547af83df8c51822b374c1e4e6b81

tsinfer/formats.py
```diff
@@ -2225,6 +2215,256 @@ def populations(self):
...
+        except KeyError:
+            # Maintains backwards compatibility: in previous tsinfer versions the
+            # ancestral allele was always the zeroth element in the alleles list
+            return np.zeros(self.num_sites, dtype=np.int8)
+
+    @property
+    def sites_genotypes(self):
+        gt = self.data["call_genotype"]
```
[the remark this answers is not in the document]
benjeffery: https://github.com/tskit-dev/tsinfer/pull/748/commits/f322ae4907ab1f996d0027bd36edaef3c6e1b98e

tsinfer/formats.py
benjeffery: https://github.com/tskit-dev/tsinfer/pull/748/commits/e0df71ad5a99ca54f8f17c0920bc76aba8812843

benjeffery:
> It's probably simplest to drop python 3.7 here from CI I'd say?

It's not that simple, the circle CI tests are also failing, I can recreate locally and it seems to be an issue with xarray on Py3.7. Using Py3.8 and doing an unpinned install of the circle CI deps results in:
```
ERROR: sgkit 0.5.0 has requirement numpy<1.22, but you'll have numpy 1.23.4 which is incompatible.
ERROR: sgkit 0.5.0 has requirement pandas<1.4.0, but you'll have pandas 1.5.1 which is incompatible.
ERROR: sgkit 0.5.0 has requirement zarr<2.11.0,>=2.10.0, but you'll have zarr 2.13.3 which is incompatible
```
Along with failing tests.

Fixing these versions then passes - so I need to upgrade the circle CI tests to 3.8 and then be very careful about the pinning of sgkits deps - this should fix circle CI. Then remove the 3.7 tests on the github workflow.

jeromekelleher:
I'm going to unsubscribe here @benjeffery - can you ping me for final review when ready please?

jeromekelleher:
(Looks like a fun time trying to get the package version problem solved! :vomiting_face: )

benjeffery:
@jeromekelleher Have you seen this error before? "The process cannot access the file because it is being used by another process" It's in the CLI tests that I haven't touched!

jeromekelleher:
Yes, I think I'm having the same problem over in #769 ...

benjeffery:
The `sgkit[vcf]` format doesn't seem to work with conda as `cyvcf2` is missing. I could add this to the conda specification, but that would then fail on windows as there is no `cyvcf2` for windows on conda. With the previous `pip` setup I could specify not to install a module on a specific OS - I'll try to find the conda way of doing this.

jeromekelleher:
I'm not sure we need vcf support in sgkit here?

benjeffery:
It's nice to confirm that the version of sgkit creates a dataset we can read, we could store the dataset, but would need to update it when sgkit changed anything.

jeromekelleher:
Right, fair enough.

The sgkit on conda should be "batteries included" anyway and have all the functionality. There's no advantage to getting it via pip because cyvcf still won't be available.

benjeffery:
@jeromekelleher Would be good to get this merged before #778 as it contains the sgkit testing infrastructure. We have issues filed for the follow-up work.

jeromekelleher (approved):

jeromekelleher:
Go for it!

benjeffery:
@mergifyio rebase

benjeffery:
Manual merge due to mergify config change.

PR #750, opened 2022-10-22
Switch parameter name to "ultimate ancestor"

### LINKED ISSUE #749: Change "split_mrca" to "split_ultimate"?
When writing up the new changes in tsinfer 0.3, I realised that perhaps "split_mrca" is badly named, because the "all zeros" ancestor need not be an MRCA anywhere in the inferred tree sequence. I wonder if "split_ultimate" is less misleading? If we think so, I can adjust the docs to refer to the "ultimate ancestor" rather than the "grand MRCA".. I guess we can do this now and not worry about backwards compatibility for parameter names, as we haven't released a new version yet. What do you think, @jeromekelleher ?

hyanwong:
Fixes #749

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,9 +33,9 @@
 
 - The `simplify` parameter is now deprecated in favour of `post_process`, which
   prior to simplification, removes the "virtual-root-like" ancestor (inserted by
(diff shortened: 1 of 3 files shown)


jeromekelleher (approved):
Great catch, this is much better. 

Documentation driven development FTW!

PR #754, opened 2022-10-25
Add --keep-intermediates CLI option

### LINKED ISSUE #105: Add --keep-intermediates option to ``infer`` command
Many times we'll just want to run the full chain with default-ish params, but also want to keep the ancestors files lying around.

hyanwong:
Thought I would fix this niggle while fiddling with the CLI files. Fixes #105

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -226,6 +226,42 @@ class TestCommandsDefaults(TestCli):
         self.verify_output(output_trees)
 
 
+class TestCommandsExtra(TestCli):
+    """
+    Test miscellaneous extra options for standard commands
+    """
+
+    def test_filenames_without_keeping_intermediates(self):
+        output_anc = os.path.join(self.tempdir.name, "test1")
+        output_anc_ts = os.path.join(self.tempdir.name, "test2")
+        with pytest.raises(ValueError, match="--keep-intermediates"):
+            self.run_command(["infer", self.sample_file, "-a", output_anc])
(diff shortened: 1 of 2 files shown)


tests/test_cli.py
```diff
@@ -226,6 +226,37 @@ def test_augment_ancestors(self):
...
+                "infer",
+                self.sample_file,
+                "--keep-intermediates",
+                "-a",
+                output_anc,
+                "-A",
+                output_anc_ts,
+            ]
+        )
+        assert os.path.exists(output_anc)
```
benjeffery, line 256: Can we go one better here and check that the right files go to the right place? Maybe by loading the ts?

benjeffery (approved):
Sorry for the delayed review here @hyanwong! Just one comment on testing.

tests/test_cli.py
hyanwong: Good call. I loaded both files and checked num_ancestors/ samples

hyanwong:
I think we can merge this, right @benjeffery ?

benjeffery:
> I think we can merge this, right @benjeffery ?

I'm happy for it - was hoping for a nod from JK - but I think his plate is full. Let's merge and ask for forgiveness if necessary ;)

benjeffery:
Gah #788. Will fix.

hyanwong:
> Will fix.

Thank you for all the hard work you are doing!

benjeffery:
@mergifyio rebase

PR #761, opened 2022-10-26
Release traceback memory if gets > 2GiB

jeromekelleher:
Otherwise many threads will end up with very large allocations as sequences that result in large tracebacks are randomly assigned to different threads.


jeromekelleher:
Merging - note I didn't squash the commits here as I want to keep the first commit hash in the history for provenance as I'm using it for an analysis.

hyanwong:
SGTM

PR #768, opened 2022-11-04
Extend greedy mutation coalescing

jeromekelleher:
WIP

### DIFF
--- a/tsinfer/inference.py
+++ b/tsinfer/inference.py
@@ -1754,6 +1754,9 @@ class SampleMatcher(Matcher):
         """
         Runs the "extend" operation matching a set of samples in the existing
         tree, returning a new tree sequence.
+
+        NOTE: this is not part of the public API is likely to be removed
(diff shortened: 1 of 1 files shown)


jeromekelleher:
This seems to be working well in practise now, so ready to merge. It's pretty clear that ultimately this code doesn't belong in tsinfer  and will be moved out into the dedicated package sc2ts when we get the required infrastructure in tskit (although we really should be doing something like this with path-compression, which currently doesn't reason about mutations at all).

@hyanwong - would you like to take a look through or shall I merge?

hyanwong:
I'm happy to look through if you want extra eyes, but also happy for you to merge if you need it right now.

jeromekelleher:
@mergifyio rebase

jeromekelleher:
I think we'll just merge, seems to be working well and I'd like to have a stable git hash to point at in provenance.

PR #769, opened 2022-11-07
Fix numpy pip

jeromekelleher:
We were pulling in pandas, matplotlib, seaborn etc from pip which seemed a bit mad to me (and was resulting in incompatible numpy versions) so I've had a go at simplifying things by pulling everything from conda. Let's see how it goes!


jeromekelleher:
This is ready for a look @benjeffery - I was hitting some problems in #768 which looked like clashing numpy versions between pip and conda packages, which lead me to ask why we were using pip at all. All of the packages we need are now in conda-forge, so I don't see much point in using pip here.

A simplification I made was to drop the pins on things like numpy where the version differs between 37 and 310. Definitely for numpy this is fine, because conda-forge to a lot of work to make sure things are pinned to given numpy versions. The other packages aren't are quite peripheral functionality so I'm not worried at all about them affecting user experience downstream.

What do you think?

jeromekelleher:
Oh yeah, disabling parallel fixed  those weird windows permission errors. Another option would be to disable the CLI tests on windows.

benjeffery:
I guess it makes sense to let conda figure out the whole dependency chain. It's easy enough to bring back pip if we need to one  day.

> Oh yeah, disabling parallel fixed those weird windows permission errors.

Ah glad that's sorted.

benjeffery (approved):

Issue #771, opened 2022-11-08
CI: windows tests failing with parallel running


jeromekelleher:
See eg [here](https://github.com/tskit-dev/tsinfer/actions/runs/3422191487/jobs/5699288681)

I thought I'd turned off parallel test running in #769 but it appears not?

The simplest thing is probably to run the tests in sequential mode. Or someone could figure out why tempdir.TemporaryDirectory doesn't seem to be working properly on Windows...

benjeffery:
Ahhh, you need `-n 0` as https://github.com/tskit-dev/tsinfer/blob/main/setup.cfg#L68 defaults testing to `-n 4`

jeromekelleher:
Sigh, still failing: https://github.com/tskit-dev/tsinfer/actions/runs/3435485226/jobs/5727901725

benjeffery:
We run the tests in serial now, so hopefully this doesn't occur again.

PR #775, opened 2022-11-10
Remove pytest-xdist

benjeffery:
Investigating #771

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -106,4 +106,4 @@ jobs:
         run: |
           source ~/.profile
           conda activate anaconda-client-env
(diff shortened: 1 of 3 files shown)


benjeffery:
Still fails with complete removal of `pytest-xdist`. Looking into the test.

benjeffery:
Haha test passes when solely that test is run.

jeromekelleher:
Maybe we should just disable those tests on Windows, probably not worth the effort in find out what's going on.

benjeffery:
I've re-run this a successfully a few times.

jeromekelleher (approved):
Great, thanks for sorting this out.

PR #776, opened 2022-11-10
Parsimoniously explain direct reversions

jeromekelleher:
Insert a new node to avoid A->B->A mutation transitions along inheritance paths.


jeromekelleher:
Happens quite a lot when you have sequences that are sampled close together in time, and you happen to sample one that has an extra mutation a day earlier.

PR #777, opened 2022-11-17
Remove "extend" code (ported to sc2ts)

jeromekelleher:
Removes the code added to facilitate using the sample matching engine in a downstream package.

Nearly all the logic was external to tsinfer, and although it's brittle to depend on internal APIs we plan to swap out the LS matching engine in sc2ts for tskit's engine ASAP, so this seems like the less bad option.


hyanwong:
Good plan. LGTM.

PR #778, opened 2022-11-22
Sgkit compatible ancestors

benjeffery:
Work in progress.

Tests finally pass with a genotype array that is sgkit-compatible.
Next I need to remove the lmdb store (without breaking SampleData which shares the storage base class) and then do some perf tests.

### DIFF
--- a/tests/test_formats.py
+++ b/tests/test_formats.py
@@ -1929,15 +1929,16 @@ class TestAncestorData(DataContainerMixin):
         stored_start = ancestor_data.ancestors_start[:]
         stored_end = ancestor_data.ancestors_end[:]
         stored_time = ancestor_data.ancestors_time[:]
-        stored_ancestors = ancestor_data.ancestors_haplotype[:]
+        # Remove the ploidy dimension
+        stored_ancestors = ancestor_data.ancestors_full_haplotype[:, :, 0]
         stored_focal_sites = ancestor_data.ancestors_focal_sites[:]
         stored_length = ancestor_data.ancestors_length[:]
-        for j, (start, end, t, focal_sites, haplotype) in enumerate(ancestors):
+        for j, (start, end, t, focal_sites, full_haplotype) in enumerate(ancestors):
             assert stored_start[j] == start
             assert stored_end[j] == end
             assert stored_time[j] == t
             assert np.array_equal(stored_focal_sites[j], focal_sites)
-            assert np.array_equal(stored_ancestors[j], haplotype[start:end])
+            assert np.array_equal(stored_ancestors[:, j], full_haplotype)
             assert np.array_equal(ancestors_list[j], haplotype[start:end])
         pos = list(ancestor_data.sites_position[:]) + [ancestor_data.sequence_length]
         for j, anc in enumerate(ancestor_data.ancestors()):
@@ -1945,7 +1946,10 @@ class TestAncestorData(DataContainerMixin):
             assert stored_end[j] == anc.end
             assert stored_time[j] == anc.time
             assert np.array_equal(stored_focal_sites[j], anc.focal_sites)
-            assert np.array_equal(stored_ancestors[j], anc.haplotype)
+            assert np.array_equal(stored_ancestors[:, j], anc.full_haplotype)
+            assert np.array_equal(
+                stored_ancestors[anc.start : anc.end, j], anc.haplotype
+            )
             length = pos[anc.end] - pos[anc.start]
             assert stored_length[j] == length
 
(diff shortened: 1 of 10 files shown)


benjeffery:
I thought it worth doing some perf now, so the effect of the lmdb removal can be seen separately, this is initially just for `generate-ancestors`, on sample data from this ts:
```
║TreeSequence               ║
╠═══════════════╤═══════════╣
║Trees          │      61779║
╟───────────────┼───────────╢
║Sequence Length│ 50000000.0║
╟───────────────┼───────────╢
║Time Units     │generations║
╟───────────────┼───────────╢
║Sample Nodes   │      30000║
╟───────────────┼───────────╢
║Total Size     │   17.8 MiB║
╚═══════════════╧═══════════╝
╔═══════════╤══════╤═════════╤════════════╗
║Table      │Rows  │Size     │Has Metadata║
╠═══════════╪══════╪═════════╪════════════╣
║Edges      │294197│  9.0 MiB│          No║
╟───────────┼──────┼─────────┼────────────╢
║Individuals│     0│ 24 Bytes│          No║
╟───────────┼──────┼─────────┼────────────╢
║Migrations │     0│  8 Bytes│          No║
╟───────────┼──────┼─────────┼────────────╢
║Mutations  │ 65362│  2.3 MiB│          No║
╟───────────┼──────┼─────────┼────────────╢
║Nodes      │102737│  2.7 MiB│          No║
╟───────────┼──────┼─────────┼────────────╢
║Populations│     1│ 16 Bytes│          No║
╟───────────┼──────┼─────────┼────────────╢
║Provenances│     1│955 Bytes│          No║
╟───────────┼──────┼─────────┼────────────╢
║Sites      │ 65362│  1.6 MiB│          No║
╚═══════════╧══════╧═════════╧════════════╝
```

`main`:
```
ga-add   (1/2)100%|██████████████████████████████████████████████████████| 65.4k/65.4k [00:07, 8.78kit/s]
ga-gen   (2/2)100%|████████████████████████████████████████████████████████| 46.4k/46.4k [02:23, 323it/s]
```
`branch`:
```
ga-add   (1/2)100%|██████████████████████████████████████████████████████| 65.4k/65.4k [00:07, 8.34kit/s]
ga-gen   (2/2)100%|████████████████████████████████████████████████████████| 46.4k/46.4k [02:58, 260it/s]
```
So this branch is slightly slower at `ga-gen`

File sizes (number is chunk size in variant dimension):
```
95M 100k.main.ancestors
62M 100k.branch.2048.ancestors
59M 100k.branch.16384.ancestors
59M 100k.branch.65536.ancestors
```
This branch gives smaller ancestor files, decreasing the chunk size (variant dimension) has a small increase in file size, and almost no change in runtime (~1s)

hyanwong:
That's looking good. The `ga` step was never much of a time-blocker anyway, so although 25% increase in time isn't ideal, it's not something worth spending huge amounts of time trying to change, either.

tests/test_formats.py
```diff
@@ -1932,29 +1932,34 @@ def verify_data_round_trip(self, sample_data, ancestor_data, ancestors):
         stored_start = ancestor_data.ancestors_start[:]
         stored_end = ancestor_data.ancestors_end[:]
         stored_time = ancestor_data.ancestors_time[:]
-        stored_ancestors = ancestor_data.ancestors_haplotype[:]
+        # Remove the ploidy dimension
+        stored_ancestors = ancestor_data.ancestors_full_haplotype[:, :, 0]
```
hyanwong, line 1936: should we assert here that the ploidy is 1, (i.e. ancestor_data.ancestors_full_haplotype.shape[2] == 1), just to make sure?

hyanwong (commented):
Looks like a good start. Re perf, just for kicks I wonder if it's worth perf testing ancestor generation with e.g. a 100K sites and 1M samples? Or instead doing it with some missing data, which will generate lots of ancestors.

benjeffery:
@jeromekelleher Now rebased onto #779 so is a bit simpler.

jeromekelleher:
Will I take a look once #779 is merged?

benjeffery:
The ancestors load into sgkit, but you can't do anything with them yet as xarray is turning the genotypes into a floating-point array, investigating.

jeromekelleher:
xarray seems keen on that sort of thing - time to start filing some upstream bugs?

benjeffery:
> xarray seems keen on that sort of thing - time to start filing some upstream bugs?

It was the same one I hit before, so already filed, but I can work around it.

jeromekelleher:
Can you link to the upstream issue please?

benjeffery:
https://github.com/pydata/xarray/issues/7292 as zarr's fill_value is `0` by default.

benjeffery:
One more wrinkle, I thought that `call_genotype_mask` was optional in `sgkit`, but appears that it is needed. This means creating and storing an additional array of the same shape and almost identical content to the genotypes.

benjeffery:
So far `sgkit.load_dataset`, `sgkit.variant_stats`, `sgkit.display_genotypes` are working on the ancestors. As I added them, `display_genotypes` required `sample_id` and `variant_stats` did not. @tomwhite what other methods should I add that give the full set of requirements on a dataset? Or is there a better way of testing "sgkit compatible"-ness?

tomwhite:
> One more wrinkle, I thought that `call_genotype_mask` was optional in `sgkit`, but appears that it is needed.

What function were you using that requires it?

> This means creating and storing an additional array of the same shape and almost identical content to the genotypes.

True, but on the other hand the mask compresses very well so isn't much storage overhead...

I note that https://github.com/pystatgen/vcf-zarr-spec/blob/main/vcf_zarr_spec.md says masks are optional ("An array called `<name>` may have an accompanying array called `<name>_mask`..."), so we might want to revisit. E.g. compute the mask on demand perhaps.

tomwhite:
> So far `sgkit.load_dataset`, `sgkit.variant_stats`, `sgkit.display_genotypes` are working on the ancestors. As I added them, `display_genotypes` required `sample_id` and `variant_stats` did not. @tomwhite what other methods should I add that give the full set of requirements on a dataset? Or is there a better way of testing "sgkit compatible"-ness?

That seems like a good test.

sgkit doesn't really mandate a particular set of variables - it depends on what you want to do. It has certain conventions (which you can see by looking at the variables that are populated by the various functions to load from plink/bgen/vcf), but each function implementing a particular genetic method will typically depend on different variables.

benjeffery:
> What function were you using that requires it?

`variant_stats` required the mask.

tomwhite:
> `variant_stats` required the mask.

Right. You could perhaps add it as needed with 

```python
data_vars["call_genotype_mask"] = ([DIM_VARIANT, DIM_SAMPLE, DIM_PLOIDY], call_genotype < 0)
```

benjeffery:
Yikes, segfault on circle CI. Adding more test output to see which one it was.

benjeffery:
Now that we have something that works in sgkit (to a first approximation) I've re-run the perf numbers. To recap this was the state when we were just writing `call_genotype`:
`main`:
```
ga-add   (1/2)100%|██████████████████████████████████████████████████████| 65.4k/65.4k [00:07, 8.78kit/s]
ga-gen   (2/2)100%|████████████████████████████████████████████████████████| 46.4k/46.4k [02:23, 323it/s]
```
`branch`:
```
ga-add   (1/2)100%|██████████████████████████████████████████████████████| 65.4k/65.4k [00:07, 8.34kit/s]
ga-gen   (2/2)100%|████████████████████████████████████████████████████████| 46.4k/46.4k [02:58, 260it/s]
```
Now with writing `call_genotype_mask` in addition:
```
ga-add   (1/2)100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 65.4k/65.4k [00:07, 8.78kit/s]
ga-gen   (2/2)100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 46.4k/46.4k [03:29, 221it/s]
```
Sadly this is quite a bit slower, but is expected seeing that we're essentially writing out the genotypes twice.
The mask is very compressible though, its information content being equal to the lengths of the ancestors, only taking up 2MB.

@jeromekelleher Is this slow down ok, or should we look at making the mask optional in sgkit?

jeromekelleher:
Overall slowdown is about 1/3 I think? If so let's not worry about it, this bit isn't a bottleneck anyway.

benjeffery:
@jeromekelleher I've factored out the common `create_dataset` args as discussed in our meeting. Would appreciate a review here now.

tsinfer/formats.py
```diff
@@ -242,29 +262,39 @@ def zarr_summary(array):
...
+            yield chunk[j % chunk_size]
+    elif dimension == 1:
+        for j in indexes:
+            chunk_id = j // chunk_size
+            if chunk_id != prev_chunk_id:
+                chunk = array[:, chunk_id * chunk_size : (chunk_id + 1) * chunk_size][:]
+                prev_chunk_id = chunk_id
+            yield chunk[:, j % chunk_size]
+    else:
+        raise ValueError("Only first two dimensions supported")
```
jeromekelleher, line 297: Needs a test (or just assert?)

jeromekelleher (approved):
LGTM, one minor thing

PR #784, opened 2022-12-07
Error if sgkit sample data is unphased

### LINKED ISSUE #772: sgkit: Error/Warning on unphased data when ploidy > 1.
I'm leaning towards this being an error - with an override flag, unless there are arguments for this being a warning?

benjeffery:
Fixes #772

This is doing unnecessary checks as not all sites may be used - but this is the simplest approach. If we think there are common use cases where used are only using a subset of sites and the rest are unphased, then could consider only checking used sites.


benjeffery:
@jeromekelleher Review changes made - should be good to go.

benjeffery:
@mergifyio rebase

PR #791, opened 2023-01-17
Add mask to sgkit SampleData

### LINKED ISSUE #786: sgkit: Support `call_genotype_mask`?
Often when running  tsinfer, a subset of the variants are wanted for inference. Making a subset sgkit dataset isn't ideal, as it means making a copy of the dataset and rechunking (see https://github.com/pystatgen/sgkit/issues/981).

One option is to have tsinfer support a mask as part of the input dataset. I think I'm right in saying that for tsinfer `variant_mask` would make the most sense - but in sgkit `call_genotype_mask` is standard. We could use `call_genotype_mask` and error if a variant has a mixture of masked and unmasked calls?

The main changes to enable this are in the chunk iterator, which would need to skip the masked rows.

benjeffery:
Fixes #786

### DIFF
--- a/tsinfer/formats.py
+++ b/tsinfer/formats.py
@@ -305,15 +305,20 @@ def zarr_summary(array):
     return ret
 
 
-def chunk_iterator(array, indexes=None, dimension=0):
+def chunk_iterator(array, indexes=None, mask=None, dimension=0):
     """
     Utility to iterate over closely spaced rows in the specified array efficiently
     by accessing one chunk at a time (normally used as an iterator over each row)
     """
     # Only the first two dimensions are supported.
     assert dimension < 2
+    if mask is None:
+        mask = np.ones(array.shape[dimension], dtype=bool)
+    else:
+        assert len(mask) == len(array)
+
     if indexes is None:
-        indexes = range(array.shape[dimension])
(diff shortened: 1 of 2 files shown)


tsinfer/formats.py
```diff
@@ -243,21 +243,32 @@ def zarr_summary(array):
...
 
-def chunk_iterator(array, indexes=None):
+def chunk_iterator(array, indexes=None, mask=None):
     """
     Utility to iterate over closely spaced rows in the specified array efficiently
     by accessing one chunk at a time (normally used as an iterator over each row)
     """
+    if mask is not None:
+        assert len(mask) == len(array)
+
```
jeromekelleher, line 253: Maybe simpler to create default mask (keep everything)?

jeromekelleher (commented):
Looks nice, not particularly invasive!

benjeffery:
This needs a few more tests before merging.

benjeffery:
Added all the other arrays that the existing sample data used, including fallbacks for when they are missing.
Everything that isn't made by sgkit's `vcf_to_zarr` is optional.

jeromekelleher (approved):
LGTM - looks like we're missing coverage on one error condition, good to catch that before merge

PR #794, opened 2023-01-18
Filter out empty string alleles

### LINKED ISSUE #792: Variant.alleles needs to filter out empty string for sgkit
Sgkit always returns a fixed length array of alleles, padded by empty srtings

benjeffery:
Fixes #792


tsinfer/formats.py
```diff
@@ -2448,6 +2451,19 @@ def variants(self, sites=None, recode_ancestral=None):
                 geno_map[aa] = 0
                 geno_map[0:aa] += 1
                 genos = geno_map[genos]
+            # Filter out empty alleles, as sgkit pads with them so that all sites have
+            # the same number of alleles. This is only safe if the empty
+            # alleles are at the end of the list, so check this.
+            non_empty_alleles = []
+            empty_seen = False
+            for allele in alleles:
+                if allele != b"" and allele != "":
```
jeromekelleher, line 2460: How about ``len(allele) == 0``? Above also - seems more robust?

jeromekelleher (commented):
LGTM, one suggestion

tsinfer/formats.py
benjeffery: Alleles can be `None` though, right? For missing data?

benjeffery:
@jeromekelleher I've had to change an existing test here that generated empty string alleles - I note that if someone wanted to model deletions, then the allele could be the empty string, and this change prevents that.

tsinfer/formats.py
hyanwong: `hmm, that's a good point. We stuck `None` as the last allele to stop people using the `-1` genotype as an index. Perhaps there's a better way to do this? It would be nice to allow (if possible) `""` to be used to indicate a deletion.

Note that this is all on the "presenting to tsinfer" side of things. If SGkit allows `None` in the alleles list, it would, I guess, be possible for SGkit to return `None` for an "empty" allele, and then, in the tsinfer format parsing, to remove those `None` values and tag another onto the end if there is missing data. Seems a bit convoluted though.

hyanwong:
> @jeromekelleher I've had to change an existing test here that generated empty string alleles - I note that if someone wanted to model deletions, then the allele could be the empty string, and this change prevents that.

In principle, I think deletions are often encoded by prepending the letter before, e.g. "ATTT" -> "A", but it *would* be nice to allow an empty string. Could we return something that is never a valid allele instead, like "\0" perhaps?

tsinfer/formats.py
jeromekelleher: Right... Fair enough.

benjeffery:
I don't think we can change sgkit's output easily as it would be a breaking change. The only safe way to do this is to remove alleles that are not referred from the genotypes - but that's some hefty processing.
Pragmatically I think we remove empty string alleles, but maybe provide a flag to disable this behaviour?

hyanwong:
> Pragmatically I think we remove empty string alleles, but maybe provide a flag to disable this behaviour?

Sure. Pragmatic solutions are good!

hyanwong:
P.s. this does make me think that there should be a method that does the tedious thing of going through the genotype indices and deleting unreferenced alleles from the list. That would also be useful after subsetting, I guess.

benjeffery:
> P.s. this does make me think that there should be a method that does the tedious thing of going through the genotype indices and deleting unreferenced alleles from the list. That would also be useful after subsetting, I guess.

Yes, feel free to raise this over at sgkit, as it doesn't belong in tsinfer.

benjeffery:
@jeromekelleher I think this is ready to merge. There's quite a bit of plumbing to make this an option, so I think we wait and see if this is an issue.

benjeffery:
This should be merged if we're happy to have it be the behaviour without an option to disable.

jeromekelleher (approved):

PR #804, opened 2023-02-21
Write only the extent of ancestors to improve generation perf

benjeffery:
This takes the `ga-gen` step from `3:41` to `0:18`. Writing the entire ancestor was slow. Flame graph before:
Screenshot at 2023-02-21 13-46-29

And after:
Screenshot at 2023-02-21 13-46-07
The large amount of time formatting strings is due to the logging lines, but I think this os only used in the synchronous code.

### DIFF
--- a/tsinfer/formats.py
+++ b/tsinfer/formats.py
@@ -21,6 +21,7 @@ Manage tsinfer's various file formats.
 """
 import collections.abc as abc
 import datetime
+import functools
 import itertools
 import json
 import logging
@@ -152,6 +153,13 @@ class BufferedItemWriter:
             for j in range(self.num_buffers):
                 self.buffers[key][j] = np.empty_like(np_array)
                 self.buffers[key][j].resize(*shape)
+                # We need to initialise the buffers for the arrays where only the extent
+                # of the ancestor is written
+                if key == "full_haplotype":
+                    self.buffers[key][j][...] = MISSING_DATA
+                elif key == "full_haplotype_mask":
+                    self.buffers[key][j][...] = True
+
             # Make sure the destination array is zero sized at the start.
             shape[chunked_dimension] = 0
             array.resize(*shape)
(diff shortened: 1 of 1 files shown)


tsinfer/formats.py
```diff
@@ -247,8 +259,15 @@ def add(self, **kwargs):
             self._queue_flush_buffer()
         offset = self.num_buffered_items[self.write_buffer]
         for key, value in kwargs.items():
-            if "full_haplotype" in key:
-                self.buffers[key][self.write_buffer][:, offset, 0] = value
+            # Here we have to special case the haplotype for performance
+            # reasons, as writing the full haplotype is expensive.
+            if key == "haplotype":
```
jeromekelleher, line 264: Could make this a bit more readable by pulling start and end out into local vars (rather then looking up kwargs repeatedly)

tsinfer/formats.py
```diff
@@ -2735,7 +2755,9 @@ def num_sites(self):
         """
         The number of inference sites used to generate the ancestors
         """
-        return self.sites_position.shape[0]
+        if not hasattr(self, "_num_sites"):
```
jeromekelleher, line 2758: Using hasattr feels a bit icky (and slow I think?). How about initialising ``self._num_sites = None`` and using that as the caching tag?

jeromekelleher (approved):
LGTM, a couple of minor comments

tsinfer/formats.py
benjeffery: I wrote it that way first - but `load` calls `__new__` and not `__init__`. The best way is probably a caching decorator. Will see if our python version supports using the one in functools.

tsinfer/formats.py
benjeffery: Fixed

tsinfer/formats.py
benjeffery: Woop! Now that we have dropped 3.7 we can use `@functools.cached_property`!

PR #808, opened 2023-03-03
Alloc genotype memory from mmap 

jeromekelleher:
Here's a quick stab that should basically work; fancy trying it out @benjeffery?

### DIFF
--- a/lib/ancestor_builder.c
+++ b/lib/ancestor_builder.c
@@ -16,6 +16,15 @@
 ** You should have received a copy of the GNU General Public License
 ** along with tsinfer.  If not, see <http://www.gnu.org/licenses/>.
 */
+/* It's not worth trying to get mmap'd genotypes working on windows,
+ * and is just a silent no-op if it's tried.
+ */
+#if defined(_WIN32)
+#else
+/* Needed for ftruncate */
+#define _XOPEN_SOURCE 700
+#define MMAP_GENOTYPES 1
+#endif
 
 #include "tsinfer.h"
 #include "err.h"
@@ -24,6 +33,17 @@
 #include <stdio.h>
 #include <string.h>
 #include <stdbool.h>
+
+#include <errno.h>
+#include <fcntl.h>
+#include <stdlib.h>
+#include <string.h>
+
+#ifdef MMAP_GENOTYPES
+#include <sys/mman.h>
+#include <unistd.h>
+#include <sys/types.h>
+#endif
 
 #include "avl.h"
 
@@ -135,6 +155,7 @@ ancestor_builder_print_state(ancestor_builder_t *self, FILE *out)
 
     fprintf(out, "Ancestor builder\n");
     fprintf(out, "flags = %d\n", (int) self->flags);
+    fprintf(out, "mmap_fd = %d\n", self->mmap_fd);
     fprintf(out, "num_samples = %d\n", (int) self->num_samples);
     fprintf(out, "num_sites = %d\n", (int) self->num_sites);
     fprintf(out, "num_ancestors = %d\n", (int) self->num_ancestors);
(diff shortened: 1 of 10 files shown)


benjeffery:
Speedy! My only option at the moment is to try this out on an network drive. I will also do the same thing on my machine, using an NMVe drive with the same ram/disk ratio to get a feel for if the performance is viable.

jeromekelleher:
Top level interface for Python is ``mmap_temp_dir``.

benjeffery:
Some performance numbers:
|  s | time | page faults | 
| ------------- | ------------- | --- |
| main  | 8m11s | 2974 |
| mmap, 100%RAM, NMVe | 8m10s  | 2613 |
| mmap, 50%RAM, NMVe | 8m48s  | 199121 |
| mmap, 50%RAM, magnetic | 1hr39min  | 1080637 |

I'm not sure why the magnetic run had so many more page faults. Will do some more testing after it's easier to set the mmap location.

hyanwong:
That's looking feasible, right. What is the non-mmap time?

benjeffery:
> That's looking feasible, right. What is the non-mmap time?

First line of the table - 8:11.

hyanwong:
> > That's looking feasible, right. What is the non-mmap time?
> 
> First line of the table - 8:11.

Oops. Didn't realise. Doh! Thanks :facepalm:

benjeffery:
I'll take a look at this once #808 is in and the diff is easy to review.

jeromekelleher:
Updated

lib/ancestor_builder.c
```diff
@@ -181,23 +202,62 @@ ancestor_builder_print_state(ancestor_builder_t *self, FILE *out)
...
-    memset(self, 0, sizeof(ancestor_builder_t));
-    if (num_samples <= 1) {
-        ret = TSI_ERR_BAD_NUM_SAMPLES;
+    self->mmap_size = self->max_sites * self->encoded_genotypes_size;
+    if (ftruncate(self->mmap_fd, (off_t) self->mmap_size) != 0) {
+        ret = TSI_ERR_IO;
         goto out;
     }
+    self->mmap_buffer = mmap(
+        NULL, self->mmap_size, PROT_READ | PROT_WRITE, MAP_SHARED, self->mmap_fd, 0);
```
benjeffery, line 219: This is probably unnecessary complexity, but once the genotypes are loaded into the mmap, we could speed things up by re-opening it with only `PROT_READ`?

tests/test_inference.py
```diff
@@ -1721,6 +1721,38 @@ def test_encodings_identical_results(self, genotype_encoding, num_threads):
...
+        with tempfile.TemporaryDirectory() as tmpdir:
+            a2 = tsinfer.generate_ancestors(
+                sample_data,
+                genotype_encoding=genotype_encoding,
+                num_threads=num_threads,
+                mmap_temp_dir=tmpdir,
+            )
+        a1.assert_data_equal(a2)
+
+    def test_mmap_missing_dir(self):
```
benjeffery, line 1746: Test mmap to dir with the wrong perms to check we get a nice message?

tsinfer/inference.py
```diff
@@ -385,6 +411,11 @@ def generate_ancestors(
     :param int genotype_encoding: The encoding to use for genotype data internally
         when generating ancestors. See the :class:`.GenotypeEncoding` class for
         the available options. Defaults to one-byte per genotype.
+    :param str mmap_temp_dir: The directory to create the temporary backing file
```
benjeffery, line 414:
```suggestion
    :param str mmap_temp_dir: The directory within which to create the temporary backing file
```

tests/test_inference.py
```diff
@@ -1721,6 +1721,38 @@ def test_encodings_identical_results(self, genotype_encoding, num_threads):
...
+        # and non-mmaped storage?
+        a1 = tsinfer.generate_ancestors(sample_data)
+        with tempfile.TemporaryDirectory() as tmpdir:
+            a2 = tsinfer.generate_ancestors(
+                sample_data,
+                genotype_encoding=genotype_encoding,
+                num_threads=num_threads,
+                mmap_temp_dir=tmpdir,
+            )
+        a1.assert_data_equal(a2)
```
benjeffery, line 1744: Can we also test that mmap file has been removed after generating the ancestors?

benjeffery (approved):
Awesome, just a couple of comments.

lib/ancestor_builder.c
jeromekelleher: Good idea. I think the benefit would be marginal, and it might cause a full flush of the file.

Worth following up if you like though, we can open an issue?

tests/test_inference.py
jeromekelleher: Done

tests/test_inference.py
jeromekelleher: Nice, done

jeromekelleher: Nice, done

jeromekelleher:
Comments addressed, should be ready to go

PR #809, opened 2023-03-04
One-bit encoding for genotypes 

jeromekelleher:
WIP 

Doesn't look too bad if we abstract things a bit, will have a look at the C API to see if it ports across.

### DIFF
--- a/_tsinfermodule.c
+++ b/_tsinfermodule.c
@@ -92,13 +92,13 @@ AncestorBuilder_init(AncestorBuilder *self, PyObject *args, PyObject *kwds)
 {
     int ret = -1;
     int err;
-    static char *kwlist[] = {"num_samples", "max_sites", NULL};
-    int num_samples, max_sites;
+    static char *kwlist[] = {"num_samples", "max_sites", "genotype_encoding", NULL};
+    int num_samples, max_sites, genotype_encoding;
     int flags = 0;
 
     self->builder = NULL;
-    if (!PyArg_ParseTupleAndKeywords(args, kwds, "ii", kwlist,
-                &num_samples, &max_sites)) {
+    if (!PyArg_ParseTupleAndKeywords(args, kwds, "ii|i", kwlist,
+                &num_samples, &max_sites, &genotype_encoding)) {
         goto out;
     }
     self->builder = PyMem_Malloc(sizeof(ancestor_builder_t));
@@ -106,6 +106,7 @@ AncestorBuilder_init(AncestorBuilder *self, PyObject *args, PyObject *kwds)
         PyErr_NoMemory();
         goto out;
     }
+    flags = genotype_encoding;
     Py_BEGIN_ALLOW_THREADS
     err = ancestor_builder_alloc(self->builder, num_samples, max_sites, flags);
     Py_END_ALLOW_THREADS
@@ -274,34 +275,6 @@ AncestorBuilder_ancestor_descriptors(AncestorBuilder *self)
         }
         PyTuple_SET_ITEM(descriptors, j, py_descriptor);
     }
-    /* j = 0; */
-    /* /1* It's not great that we're breaking encapsulation here and looking */
-    /*  * directly in to the builder's data structures. However, it's quite an */
-    /*  * awkward set of data to communicate, so it seems OK. *1/ */
-    /* for (f = self->builder->num_samples; f > 0; f--) { */
-    /*     for (a = self->builder->frequency_map[f].head; a != NULL; a = a->next) { */
-    /*         map_elem = (pattern_map_t *) a->item; */
-    /*         dims = map_elem->num_sites; */
-    /*         site_array = (PyArrayObject *) PyArray_SimpleNew(1, &dims, NPY_INT32); */
-    /*         if (site_array == NULL) { */
-    /*             goto out; */
-    /*         } */
-    /*         site_array_data = (int32_t *) PyArray_DATA(site_array); */
-    /*         /1* The elements are listed backwards, so reverse them *1/ */
-    /*         k = map_elem->num_sites - 1; */
-    /*         for (s = map_elem->sites; s != NULL; s = s->next) { */
-    /*             site_array_data[k] = (int32_t) s->site; */
-    /*             k--; */
-    /*         } */
-    /*         descriptor = Py_BuildValue("kO", (unsigned long) f, site_array); */
-    /*         if (descriptor == NULL) { */
-    /*             Py_DECREF(site_array); */
-    /*             goto out; */
-    /*         } */
-    /*         PyTuple_SET_ITEM(descriptors, j, descriptor); */
-    /*         j++; */
-    /*     } */
-    /* } */
     ret = descriptors;
     descriptors = NULL;
 out:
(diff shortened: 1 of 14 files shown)


benjeffery:
Looking good - let me know when I should run some quick tests.

jeromekelleher:
Use like ``generate_ancestors(params as before, genotype_encoding=1)``

jeromekelleher:
Aha, that should be it. Curios to see what the perf cost/benefit of this is!

benjeffery:
Not looking good so far, `ancestor_builder_finalise` is taking a _lot_ longer. I'm not sure how much as it is still running after 8min, (usually that step takes a few seconds). Attaching `gdb` shows the process to be in `unpackbits`. I'll try a more optimised version of that function.

jeromekelleher:
> Not looking good so far, ancestor_builder_finalise is taking a lot longer. 

Hmm, I can see how that might happen. OK, I'll see what I can do.

benjeffery:
This seems to be better:
```c
void unpackbits(const uint8_t *restrict source, size_t len, int8_t *restrict dest) {
    uint64_t MAGIC = 0x8040201008040201ULL;
    uint64_t MASK  = 0x8080808080808080ULL;
    size_t dest_index = 0;
    for (size_t i = 0; i < len; i++) {
        uint64_t t = ((MAGIC*source[i]) & MASK) >> 7;
        *(uint64_t*)&dest[dest_index] = t;
        dest_index += 8;
    }
}
```
It's running now, and has got past the `ancestor_builder_finalise` step.

benjeffery:
Whoops, maybe that isn't doing the right thing as some tests are failing now, was running fast until it failed though!!

jeromekelleher:
Currently refactoring to do selective decoding in some places. We can drop a more optimised function in later.

jeromekelleher:
Looks like this is working pretty well. Here's my informal perf testing based on a simulation with 10K samples and 78K sites:
- main: time 1:20s, max resident: 760M
- this branch, genotype_encoding=0, time=1:31m max resident: 1.0G
- this branch, genotype_encoding=1, time=1:20, max_resident: 460M

So, we're about as fast as the current head and save quite a bit of memory! The memory isn't optimised at the moment, I'll need to take a bit more care with that.

But, I think the approach basically works now and we don't really need any major changes, so I'll try to get this mergable ASAP.

jeromekelleher:
Good to get some confirmation on real data as well though @benjeffery, if you can get it working on the datasets you're working on?

jeromekelleher:
Hmm, might have made a retrograde step with the memory usage, will have another look.

benjeffery:
I also get:
```
CRITICAL:tsinfer.threads:Exception occured in thread; exiting
CRITICAL:tsinfer.threads:Traceback (most recent call last):
  File "/home/benj/projects/tsinfer/tsinfer/threads.py", line 59, in thread_target
    worker(index)
  File "/home/benj/projects/tsinfer/tsinfer/inference.py", line 1214, in build_worker
    start, end = self.ancestor_builder.make_ancestor(focal_sites, a)
_tsinfer.LibraryError: Bad focal site.
```

jeromekelleher:
Right, crap, we're reusing a decode buffer.

jeromekelleher:
that should fix it

benjeffery:
Right, here's some results. Generating ancestors for 1000g with 8 threads (not all fully used):
(793802 sites * 3202 samples * 2 ploidy = 5.08GB/635MB

| Code version  | wallclock | builder RAM| process resident RAM |
| ------------- | ------------- | ------------- | ------------- |
| 1kg(byte)  | 8:31  | n/a | 7.2GB |
| bitpack(byte) | 8:58 | 4.7GB | 8.7GB |
| bitpack(1bit) | 8:56 | 608.7MB | 4.3GB |
| bitpack-magic(1bit) | 6:58 | 608.7MB | 4.0GB |

So with the "magic" unpack I posted in https://github.com/tskit-dev/tsinfer/pull/809#issuecomment-1460172739 we're faster than the normal code, which I guess makes sense as you can hold more data in the CPU cache, and need less main memory bandwidth. The ~3.5GB ram is a constant term from the ancestor write buffers I think, so shouldn't worry us. 

I will now try this code on the GEL chr 22.

jeromekelleher:
Awesome!

Since the speed diff isn't massive we can add the bitpack-magic change in a follow-up. It needs some care about buffer sizes, I think, and probably a few choice tests to make sure we don't overflow on awkard cases.

benjeffery:
GEL chr22 in 3:05:51 using 76GB RAM using the bitpack-magic branch. Awesome work on this one, that's hopefully the last big blocker removed!

jeromekelleher:
Great! 

Don't use the magic version for actual inference though, there's definitely a buffer overflow and undefined behaviour. Easy fix though (just a case of making sure genotype encode buffer size is divisible by 64 rather than 8)

hyanwong:
BTW, presumably if we have missing data we can use a 2-bit encoding for twice the RAM cost? Would this be useful for others, or hard to code up? No reason to do it now, but maybe an issue for the future?

jeromekelleher:
Yeah, that's the plan

hyanwong:
> Yeah, that's the plan

Shall I open an issue?

benjeffery:
Ok will re-run the benchmarks on 1kg.

jeromekelleher:
Can you open an issue to track adding your 64 bit unpack implementation? It should be a nice simple follow up here

benjeffery:
https://github.com/tskit-dev/tsinfer/issues/816

jeromekelleher:
I think this should be ready to go @benjeffery?

tests/test_inference.py
```diff
@@ -1667,6 +1639,36 @@ def test_oldest_ancestors(self):
...
+            sample_data.add_site(position, genotypes)
+        sample_data.finalise()
+        # Do we get precisely the same result as the Python engine with no encoding?
+        a1 = tsinfer.generate_ancestors(sample_data, engine=tsinfer.PY_ENGINE)
+        a2 = tsinfer.generate_ancestors(
+            sample_data, genotype_encoding=genotype_encoding, num_threads=num_threads
+        )
+        a1.assert_data_equal(a2)
+
+    def test_one_bit_encoding_missing_data(self):
```
benjeffery, line 1659: Do we also need to test the case of no missingness but 3 states?

_tsinfermodule.c
```diff
@@ -274,34 +275,6 @@ AncestorBuilder_ancestor_descriptors(AncestorBuilder *self)
         }
         PyTuple_SET_ITEM(descriptors, j, py_descriptor);
     }
-    /* j = 0; */
```
benjeffery, line 277: I assume this is commented out as it might be useful for future debugging?

benjeffery (approved):
LGTM, couple of comments.
There is a block of uncovered code in the `data_equal` method. I assume that's because all the tests pass, but might be nice to add one unequal set that is unequal by the last criteria to be checked?

tests/test_inference.py
jeromekelleher: Good idea, but can't test because mulilti allelic don't get through to ancestor gen

_tsinfermodule.c
jeromekelleher: This is all deleted , just old crufr that was commented out

jeromekelleher:
Hopefully ready to go now

PR #812, opened 2023-03-22
Pass through options to tqdm

benjeffery:
Useful for when you need to redirect output to a file (For example when a batch system provides no stdout/err until the job has ended.


hyanwong:
LGTM

tsinfer/progress.py
```diff
@@ -35,6 +35,7 @@ def __init__(
         augment_ancestors=False,
         match_samples=False,
         verify=False,
+        **kwargs,
```
jeromekelleher, line 38: Maybe tqdm_kwargs would be more explanatory?

jeromekelleher (commented):
Slightly hostage to API name collisions here, maybe a dict of args to pass through to tqdm would be better? Or maybe just provide a stream option explicitly if that's the only immediate use?

tsinfer/progress.py
benjeffery: Yeah good idea - fixed.

benjeffery:
@Mergifyio rebase

PR #818, opened 2023-04-11
Dask experiments

benjeffery:
Sample matching using dask.
The overhead for using the default dask client with threads in the current process is very small ~2-3%, as the TSB doesn't need to be serialised.
For multi-process workers I get a ~30% longer runtime, this is worst case as I have only 1000 samples and 10000 sites.
The extra cost is mostly fixed setup cost, this is worth it for flexible multi-node worker clusters though I think. 

This needs some tidying up before merging, and reworking the testing, (the test_dask.py was more of a playground than actual testing).

### DIFF
--- a/tsinfer/inference.py
+++ b/tsinfer/inference.py
@@ -27,6 +27,7 @@ import dataclasses
 import heapq
 import json
 import logging
+import os
 import pickle
 import queue
 import sys
@@ -35,6 +36,8 @@ import threading
 import time
 import time as time_
 
+import dask
+import dask.bag as db
 import humanize
 import lmdb
 import numpy as np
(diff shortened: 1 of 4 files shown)


benjeffery:
Rebased onto main, will now try this out if the cluster is quiet.

benjeffery:
@jeromekelleher This is ready for a review - I'm not sure why the circleCI tests are getting killed. Will look into it.

tsinfer/inference.py
```diff
@@ -531,6 +562,7 @@ def match_ancestors(
         engine=engine,
         progress_monitor=progress_monitor,
         resume_lmdb_file=resume_lmdb_file,
+        dask=dask,
```
jeromekelleher, line 565: maybe ``use_dask`` instead of ``dask`` so we don't shadow the package name?

jeromekelleher (approved):
LGTM, let's merge and use it to see how we go

PR #827, opened 2023-05-15
Resume using LMDB

### LINKED ISSUE #826: Store resume buffer of match results
As an alternative to #825 and #793, we could keep an LMBD which maps the keys (child node IDs) to their results. The ``_find_path`` function should check this (if it's enabled?) and if the result exists we return that instead of running the match.

This should give us a pragmatic way of restarting long running computations.

benjeffery:
Fixes #826

### DIFF
--- a/tsinfer/inference.py
+++ b/tsinfer/inference.py
@@ -27,13 +27,16 @@ import dataclasses
 import heapq
 import json
 import logging
+import pickle
 import queue
+import sys
 import tempfile
 import threading
 import time
 import time as time_
 
 import humanize
+import lmdb
 import numpy as np
 import tskit
 
@@ -472,6 +475,7 @@ def match_ancestors(
     extended_checks=False,
     time_units=None,
     record_provenance=True,
+    resume_lmdb_file=None,
 ):
     """
     match_ancestors(sample_data, ancestor_data, *, recombination_rate=None,\
@@ -526,6 +530,7 @@ def match_ancestors(
         extended_checks=extended_checks,
         engine=engine,
         progress_monitor=progress_monitor,
+        resume_lmdb_file=resume_lmdb_file,
     )
     ts = matcher.match_ancestors()
     tables = ts.dump_tables()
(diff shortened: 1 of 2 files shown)


jeromekelleher:
Nice! Does it work???

benjeffery:
It does now!

Note that this won't work with the dask code path as it implements it's own `_find_path` so that it runs on the worker, but that's easily worked around in the dask PR.

tsinfer/inference.py
```diff
@@ -1391,7 +1410,36 @@ def _find_path(self, child_id, haplotype, start, end, thread_index=0):
         match = self.match[thread_index]
         missing = haplotype == tskit.MISSING_DATA
 
-        left, right, parent = matcher.find_path(haplotype, start, end, match)
+        key = f"{child_id}_{self.__class__.__name__}".encode()
+
+        if self.resume_lmdb is not None:
+            with self.resume_lmdb.begin() as txn:
```
jeromekelleher, line 1416: I'm having second thoughts about doing this in the worker threads. Although this is the place to put the dB lookup to minimise code changes, it introduces a db lookup in each worker. This will be hard to test, and introduce mutexes (if the dB does concurrent updates properly). Can we do the lookup in the main process, even if it means a bit of duplication between the samples and ancestors code paths?

This should also simplify doing it under dask

jeromekelleher (commented):
Very cool. See comment

tsinfer/inference.py
benjeffery: Yes, `lmdb` actually uses a file-based lock so not ideal. I'll try to think of a way that this can be abstracted to work for both dask and local paths.

jeromekelleher:
As discussed, let's change to pushing all the results for a given ancestor epoch into the DB in one go, before we go out to the worker threads (rebase on #829 )

benjeffery:
Here's a stab at whole epoch caching.

jeromekelleher:
Nice

PR #828, opened 2023-05-19
Add both linesweep and levelorder ancestor grouping

benjeffery:
Will be using this branch for testing and timing on the GeL data, after which I'll probably reduce it down to just the time and sweep based methods if the sweep is working well.


benjeffery:
An update here  - I've refactored to use `numba` and I've been trying to run this on our largest GeL ancestor files. It still takes over an hour to do the linesweep and grouping. The last few epochs also have many overlapping ancestors which is slow to resolve. However I realised we can retain time grouping for the last few, very large, epochs so I'm attempting a hybrid approach where the linesweep only groups the earlier epochs, this is looking promising.

benjeffery:
Doing the top 30% takes a couple of min for 400,000 ancestors.
I'm getting a `LibraryError: Bad path information: time` though, so clearly the grouping isn't exactly right.

benjeffery:
Testing with random sets of lines has revealed a mistake in pushing "overlapping same time" ancestors to the last of their groups, as this can push an ancestor into being in a group with one of it's dependencies. Working on a fix.

hyanwong:
> Testing with random sets of lines has revealed a mistake in pushing "overlapping same time" ancestors to the last of their groups, as this can push an ancestor into being in a group with one of it's dependencies. Working on a fix.

Well spotted. Worth the extensive testing on this one, as the logic is quite complex.

benjeffery:
I've refactored the algorithm heavily to merge overlapping same-time ancestors before dependency resolution, this now passes all the tests that I can throw at it.

Will now try this branch on the GeL data.

benjeffery:
For the ~400k ancestors in the GeL 5k subset:

Grouping by time 0.5s
Grouping by level order 92s
Grouping by linesweep 6s

Screenshot from 2023-06-07 18-36-04

I'll clean up the code to remove levelorder and then I think it's good to merge.

jeromekelleher:
Amazing! Great work 💪

benjeffery:
Would appreciate a review here now. CircleCI tests are getting killed, haven't looked into it yet.

tsinfer/algorithm.py
```diff
@@ -1118,3 +1119,166 @@ def run_traceback(self, start, end, match):
             parent[j] = e.parent
 
         return left, right, parent
+
+
+def merge_overlapping_ancestors(start, end, time):
```
jeromekelleher, line 1124: I guess we'd usually use docstrings here for implementation comments - but maybe you're consciously avoiding documenting so people don't think it's part of public API?

tsinfer/algorithm.py
```diff
@@ -1118,3 +1119,166 @@ def run_traceback(self, start, end, match):
...
+def run_linesweep(event_times, event_index, event_type, new_time):
+    # Run the linesweep over the ancestor start-stop events,
+    # building up the dependency graph as a count of dependencies for each ancestor,
+    # and a list of dependant children for each ancestor.
+    n = len(new_time)
+
+    # numba really likes to know the type of the list elements, so we tell it by adding
+    # a dummy element to the list and then popping it off.
+    # `active` is the list of ancestors that overlap with the current linesweep position.
+    active = [-1]
```
jeromekelleher, line 1174: FYI you can also do this by explicitly creating a TypedList (I think it's called). This is fine though

tsinfer/algorithm.py
```diff
@@ -1118,3 +1119,166 @@ def run_traceback(self, start, end, match):
...
+    # For a given set of ancestors, we want to group them for matching in parallel.
+    # For each ancestor, any overlapping, older ancestors must be in an earlier group,
+    # and any overlapping, younger ancestors in a later group. Any overlapping same-age
+    # ancestors must be in the same group so they don't match to each other.
+    # We do this by first merging the overlapping same-age ancestors. Then build a
+    # dependency graph of the ancestors by linesweep. Then form groups by topological
+    # sort. Finally, we un-merge the same-age ancestors.
+
+    assert len(start) == len(end)
+    assert len(start) == len(time)
```
jeromekelleher, line 1242: Good to add a few ``logging.info`` comments in this function so that we can see what's going on and the relative timings of the steps

tsinfer/inference.py
```diff
@@ -1444,85 +1443,116 @@ def __init__(self, sample_data, ancestor_data, time_units=None, **kwargs):
...
+        epoch_sizes = time_slices[:, 1] - time_slices[:, 0]
+        # To choose a cutoff point find the first epoch that is 50 times larger than
+        # the median epoch size. For a large set of human genomes the median epoch
+        # size is around 10, so we'll stop grouping by linesweep at 5000.
+        try:
+            first_large_epoch = np.where(epoch_sizes > 500 * np.median(epoch_sizes))[0][
+                0
+            ]
+            first_large_epoch_first_ancestor = time_slices[first_large_epoch, 0]
+        except IndexError:
```
jeromekelleher, line 1481: Not clear why we get an IndexError here. I think it's better to have a conditional on the actual cause here rather than letting exceptions handle things, as it can mask genuine errors.

I guess it's something like
```python
cutoff =  500 * np.median(epoch_sizes)
if np.max(epoch_sizes) < cutoff:
    first_large_epoch = np.where(..)
else:
   # other thing
```

jeromekelleher (approved):
Superb - a few minor bike shedding comments. Go ahead and merge when you're happy.

Good to add (a stub at least) this to the CHANGELOG, it's a major update

tsinfer/algorithm.py
```diff
@@ -1118,3 +1119,166 @@ def run_traceback(self, start, end, match):
             parent[j] = e.parent
 
         return left, right, parent
+
+
```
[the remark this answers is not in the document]
benjeffery: I had previously tried and got some issues such as `lexsort` not being supported. I bought that out to the outer function, but then there was as issue with the type of `old_dict` and as it was taking almost no time I cut my losses!

tsinfer/algorithm.py
benjeffery: Yes, that was my thought. But then I guess this won't get pulled into the actual compiled docs so even if it had a docstring it isn't actually "documented". Happy with either way.

tsinfer/algorithm.py
benjeffery: Yeah, I was put off by the "Experimental Feature" warning in the numba docs.

tsinfer/algorithm.py
```diff
@@ -1118,3 +1119,166 @@ def run_traceback(self, start, end, match):
             parent[j] = e.parent
 
         return left, right, parent
```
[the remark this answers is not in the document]
benjeffery: Great, done.

tsinfer/algorithm.py
```diff
@@ -1118,3 +1119,166 @@ def run_traceback(self, start, end, match):
...
+
+def merge_overlapping_ancestors(start, end, time):
+    # Merge overlapping, same-time ancestors. We do this by scanning along a single
+    # time epoch from left to right, detecting breaks.
+    sort_indices = np.lexsort((start, time))
+    start = start[sort_indices]
+    end = end[sort_indices]
+    time = time[sort_indices]
+    old_indexes = {}
+    # For efficiency, pre-allcate the output arrays to the maximum possible size.
```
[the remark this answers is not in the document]
benjeffery: Done

tsinfer/algorithm.py
benjeffery: Added logging for timing, and also info such as where the linesweep cutoff was.

tsinfer/inference.py
benjeffery: Done

jeromekelleher:
Happy to merge when you are

PR #829, opened 2023-05-22
Start on factoring out ResultBuffer

jeromekelleher:


benjeffery:
Just realised that `ThreadPoolExecutor.map` greedily consumes a generator, which won't work for this use case as we'll load all the sample haplotypes into memory before doing any work, will refactor.

jeromekelleher:
It should flow control though, shouldn't it? As in,  pop stuff off the input generator as work gets done?

benjeffery:
> It should flow control though, shouldn't it? As in, pop stuff off the input generator as work gets done?

Not from what I glean from the docs: "the iterables are collected immediately rather than lazily" from https://docs.python.org/3/library/concurrent.futures.html

tsinfer/inference.py
```diff
@@ -2060,40 +2011,15 @@ def get_augmented_ancestors_tree_sequence(self, sample_indexes):
...
-        with self.lock:
-            self.mutations[node_id] = site, derived_state
-
-    def get_path(self, node_id):
-        return self.paths[node_id]
-
-    def get_mutations(self, node_id):
-        return self.mutations[node_id]
+@dataclasses.dataclass
+class Result:
```
jeromekelleher, line 2015: MatchResult or something similar might be more descriptive @benjeffery. See [here](https://github.com/jeromekelleher/tsinfer/blob/f9070616b58221ab02f01738108e0ce19043ad35/tsinfer/matching.py#L104) for a rough version of how I was imagining it to work later on. It might be worth adding the Path class now, so we go ``result.path.left`` rather than ``result.path_left``.

jeromekelleher (commented):
Looks good, a few minor comments.

tsinfer/inference.py
benjeffery: Done

benjeffery:
@jeromekelleher Comments addressed

tsinfer/inference.py
```diff
@@ -1499,105 +1477,86 @@ def __start_epoch(self, epoch_index):
...
-            self.__complete_epoch(j)
+            if self.num_threads > 0:
+                mapper = lambda func, args: threads.threaded_map(  # noqa E731
+                    func, args, self.num_threads
+                )
+            else:
+                mapper = map
+            # We have to consume the iterator here, otherwise we'll match against
+            # ancestors in this epoch
+            results = list(
```
jeromekelleher, line 1548: Any reason not to do something like this?
```python
results = []
work_iter = zip(self.ancestors, range(start, end))
for result in mapper(worker_function, work_iter):
     results.append(result)
     match_progresss.update()
```

I always find it difficult to untangle nested iterators like that, and it will surely be helpful to take the progress update out of worker threads

jeromekelleher (commented):
LGTM, couple of minor comments

tsinfer/inference.py
benjeffery: I had a go in https://github.com/tskit-dev/tsinfer/pull/829/commits/03b61b67716bab558054a1ad00c0aa198a44a93c

tsinfer/inference.py
```diff
@@ -1747,92 +1706,73 @@ def restore_tree_sequence_builder(self):
...
+            if self.num_threads > 0:
+                mapper = lambda func, args: threads.threaded_map(  # noqa E731
+                    func, args, self.num_threads
+                )
             else:
-                self.__match_samples_multi_threaded(sample_indexes)
+                mapper = map
+            # We have to consume the iterator here, otherwise we'll match against
+            # ancestors in this epoch
+            results = list(mapper(worker_function, sample_haplotypes))
```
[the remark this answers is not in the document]
benjeffery: Yep, done in https://github.com/tskit-dev/tsinfer/pull/829/commits/03b61b67716bab558054a1ad00c0aa198a44a93c

benjeffery:
@jeromekelleher I'll squash if all looks good here.

jeromekelleher (commented):
squash n merge!

PR #844, opened 2023-06-27
Update installation.rst

eaton-lab:
This conda install command in the tsinfer docs does not work because `conda` must be followed by a command. See error from bash terminal call of `conda -c conda-forge install tsinfer`:

```bash
CommandNotFoundError: No command 'conda conda-forge'.
```


benjeffery (approved):
Thanks for the fix @eaton-lab!

PR #849, opened 2023-07-06
Rework sgkit ancestral allele 

benjeffery:
Fixes #847, #846

PR #853, opened 2023-08-03
Dask sample matching

benjeffery:
Stacked on #849 
There are some challenges here with reading the sample data off disk in a performant manner. Initially, each sample was an individual `dask` task, but this means that each task must read a "sample chunk width" amount of data off disk. This was changed to combine 10 contiguous matching tasks into one `dask` task.

### DIFF
--- a/tsinfer/inference.py
+++ b/tsinfer/inference.py
@@ -34,7 +34,7 @@ import sys
 import tempfile
 import threading
 import time
-import time as time_
+import zlib
 
 import dask
 import dask.bag as db
@@ -784,7 +784,7 @@ def match_samples(
         engine=engine,
         progress_monitor=progress_monitor,
         resume_lmdb_file=resume_lmdb_file,
-        use_dask=dask,
+        use_dask=use_dask,
     )
     sample_indexes = check_sample_indexes(sample_data, indexes)
     sample_times = np.zeros(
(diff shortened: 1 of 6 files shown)


tsinfer/inference.py
```diff
@@ -35,6 +35,7 @@
 import threading
 import time
 import time as time_
```
jeromekelleher, line 37: This is old, but I'm not sure why we're importing ``time`` twice?

tsinfer/inference.py
```diff
@@ -1856,6 +1893,57 @@ def __init__(self, sample_data, ancestors_ts, **kwargs):
...
+            else algorithm.AncestorMatcher
+        )
+        matcher = ancestor_matcher_class(
+            tree_sequence_builder_wrapper.tsb,
+            recombination=recombination,
+            mismatch=mismatch,
+            precision=precision,
+            extended_checks=extended_checks,
+        )
+        sys.stderr.write(
```
jeromekelleher, line 1922: This is a bit grotty - a comment explaining why we need to do this (rather than logging) would be helpful. Also, why not ``print(..., file=sys.stderr)``?

jeromekelleher (approved):
LGTM, a few trivial nitpicks. Merge away when you're happy.

tsinfer/inference.py
benjeffery: I had trouble getting dask to set the log level on the workers - just found https://github.com/dask/distributed/issues/6124 though which should fix that, so have changed the code to proper logging.

tsinfer/inference.py
benjeffery: Weird, I think it's safe to remove that!

benjeffery:
Fixed up hopefully!

jeromekelleher (approved):

PR #854, opened 2023-08-03
Update deps and fix evaluation.py for new pandas

benjeffery:


benjeffery:
No idea why we're getting a coverage drop here.

PR #856, opened 2023-08-29
Switch to sphinx-book-theme

benjeffery:


hyanwong:
Mostly works, thanks @benjeffery, and I like the light/dark/auto mode switch in the new toolbar at the top right. A few minor issues (observed in Safari/OS X):

Dark mode:

1. It would be nice to use the current very-dark-blue background, I think.
2. Output from tqdm's notebook mode is dark-on-dark (here from the tutorial.html page)
Screenshot 2023-08-29 at 13 43 25
3. Code syntax highlighting has possibly *too* intense a white for "normal" text 

Light mode:

White test for syntax highlighting is unreadable:
Screenshot 2023-08-29 at 13 46 25

jeromekelleher (approved):

benjeffery:
> White test for syntax highlighting is unreadable:

I think that might be fixed by not specifying the code theme - will test it out.

benjeffery:
Ok, we now have a different code theme for light mode:
Screenshot from 2023-09-04 11-25-19

benjeffery:
Having issues with filling up the disk on windows CI. According to the docs we should have 14GB (https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources) and the line of code here is only trying to use 1GB. Will re-run a few times to see if this is transitory.

PR #861, opened 2023-09-06
Make mapping additional sites optional

benjeffery:


jeromekelleher (approved):
LGTM. We might want a different name ("additional" isn't that helpful) but we're not documenting for now so let's not worry about it.

PR #872, opened 2023-11-10
Return sgkit sample_id in individual metadata

### LINKED ISSUE #871: Record sgkit sample_ids in individual metadata
As these are always present in sgkit, we should bring them through to the final TS. See https://github.com/benjeffery/tsinfer-snakemake/issues/17

benjeffery:
Fixes #871

### DIFF
--- a/tsinfer/formats.py
+++ b/tsinfer/formats.py
@@ -2512,10 +2512,21 @@ class SgkitSampleData(SampleData):
     @functools.cached_property
     def individuals_metadata(self):
         schema = tskit.MetadataSchema(self.populations_metadata_schema)
-        try:
-            return [schema.decode_row(r) for r in self.data["individuals_metadata"][:]]
-        except KeyError:
-            return [{} for _ in range(self.num_individuals)]
+        if "individuals_metadata" in self.data:
+            assert len(self.data["individuals_metadata"]) == self.num_individuals
+            assert self.num_individuals == len(self.data["sample_id"])
+            md_list = []
+            for sample_id, r in zip(
+                self.data["sample_id"], self.data["individuals_metadata"][:]
+            ):
+                md = schema.decode_row(r)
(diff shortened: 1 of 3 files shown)


benjeffery:
Looks like we have some OSX regressions to do with the unwriteable file tests. Will fix in another PR.

tsinfer/formats.py
```diff
@@ -2512,10 +2512,21 @@ def individuals_metadata_schema(self):
     @functools.cached_property
     def individuals_metadata(self):
         schema = tskit.MetadataSchema(self.populations_metadata_schema)
-        try:
-            return [schema.decode_row(r) for r in self.data["individuals_metadata"][:]]
-        except KeyError:
-            return [{} for _ in range(self.num_individuals)]
+        if "individuals_metadata" in self.data:
```
jeromekelleher, line 2515: Took me a few minutes to figure this out - maybe a comment explaining that we always include the sgkit_sample_id, whether or not the user specified other metadata? What happens if the user already has ``sgkit_sample_id`` defined in their metadata?

This needs a CHANGELOG update too.

jeromekelleher (commented):
LGTM, needs a bit of care with corner cases and docs though

tsinfer/formats.py
benjeffery: I think if the key already exists that we silently don't set it. I don't think it is worth a warning as sgkit_sample_id is so specific it is likely to be the same string.

Do we need a changelog? None of the sgkit stuff is released yet.

jeromekelleher: True enough - I can see us forgetting to document this addition to the metadata though, so worth starting the CHANGELOG for sgkit stuff now?

benjeffery:
Ok added a commit here, which prevents clobbering and tests that. Also a quick start at the sgkit changelog.

benjeffery:
@mergifyio rebase

Issue #870, opened 2023-11-09
No istes used for inference


bgyuris:
Dear tskit-dev team!

I always run into this issue when runnig tsinfer:

Number of polymorphic sites: 3906560
No sites used for inference
No sites used for inference
Inferred tree sequence 'sparrow_ts': 2 trees over 249.23985 Mb

I really do not get the problem. Maybe it is related to my vcf somehow?

Great tool anyway,
Thank you for the help

benjeffery:
Hi @bgyuris!
The requirements for a site to be used are listed at https://tskit.dev/tsinfer/docs/stable/inference.html#data-requirements
Note that the site must have a known ancestral allele and be biallelic. Are your sites valid?

hyanwong:
Thanks for reporting this. Can you provide us with a snippet of your VCF? Have you tried with increased verbosity, which should output logs? I'm sure we can help you fix it. 

I'll try to put an FAQ together too, and add this to it.

hyanwong:
When we say "No sites used for inference" it would be useful to report the numbers of non-inference sites of different sorts, I think (e.g. V with no defined ancestral allele, W presumed >bi -allelic, X singletons, Y monoallelic, U unphased, Z user-excluded). What do you think @benjeffery ?

benjeffery:
Yeah, would be good to have more detail - we'd have to log it while adding sites such that we don't have to scan when the condition is hit.

Issue #683, opened 2022-06-22
Raise warning when zero inference sites provided.


evolgenomics:
Hi team,

I was trying to run tsinfer on the 1000G GBR samples, focusing on a single 1Mbp region.

I imported using a VCF input copied off the tutorial. Essentially:

vcf = cyvcf2.VCF('GBR.variable.vcf.gz')
with tsinfer.SampleData(
        path="GBR.samples", sequence_length=chromosome_length(vcf)
) as samples:
        add_diploid_sites(vcf, samples)

# Do the inference
ts = tsinfer.infer(samples)

ts.dump("GBR.trees")

$ python tsinfer_convert.py

Sample file created for 182 samples (182 individuals) with 262165 variable sites.
Inferred tree sequence: 1 trees over 63.02552 Mb (182 edges)

$ tskit info GBR.trees
╔═════════════════════════╗
║TreeSequence             ║
╠═══════════════╤═════════╣
║Trees          │        1║
╟───────────────┼─────────╢
║Sequence Length│ 63025520║
╟───────────────┼─────────╢
║Time Units     │  unknown║
╟───────────────┼─────────╢
║Sample Nodes   │      182║
╟───────────────┼─────────╢
║Total Size     │237.6 MiB║
╚═══════════════╧═════════╝
╔═══════════╤═══════╤══════════╤════════════╗
║Table      │Rows   │Size      │Has Metadata║
╠═══════════╪═══════╪══════════╪════════════╣
║Edges      │    182│   5.7 KiB│          No║
╟───────────┼───────┼──────────┼────────────╢
║Individuals│    182│   5.4 KiB│         Yes║
╟───────────┼───────┼──────────┼────────────╢
║Migrations │      0│   8 Bytes│          No║
╟───────────┼───────┼──────────┼────────────╢
║Mutations  │6314841│ 222.8 MiB│          No║
╟───────────┼───────┼──────────┼────────────╢
║Nodes      │    183│   5.0 KiB│          No║
╟───────────┼───────┼──────────┼────────────╢
║Populations│      0│   8 Bytes│          No║
╟───────────┼───────┼──────────┼────────────╢
║Provenances│      2│1008 Bytes│          No║
╟───────────┼───────┼──────────┼────────────╢
║Sites      │ 262165│  14.8 MiB│         Yes║
╚═══════════╧═══════╧══════════╧════════════╝

I tried doing the whole chromosome, or a single 1Mbp, and got the same results.

[...]

PR #887, opened 2024-01-19
Better formatted error message for ancestral alleles

hyanwong:
This is pretty useful information, so quoting percentages is helpful to debug issues, I think.

### DIFF
--- a/tsinfer/formats.py
+++ b/tsinfer/formats.py
@@ -2405,11 +2405,20 @@ class SgkitSampleData(SampleData):
             except IndexError:
                 unknown_alleles[allele] += 1
             ret[i] = allele_index
-        if sum(unknown_alleles.values()) > 0:
+        tot = sum(unknown_alleles.values())
+        if tot > 0:
+            num_sites = len(string_allele)
+            frac_bad = tot / num_sites
+            frac_bad_per_type = [v / num_sites for v in unknown_alleles.values()]
+            summarise_unknown = [
+                f"'{k}': {v} ({frac * 100:.2f}% of sites)"  # Summarise per allele type
+                for (k, v), frac in zip(unknown_alleles.items(), frac_bad_per_type)
+            ]
             warnings.warn(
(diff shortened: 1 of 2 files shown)


tsinfer/formats.py
```diff
@@ -2405,11 +2405,16 @@ def sites_ancestral_allele(self):
...
                 unknown_alleles[allele] += 1
             ret[i] = allele_index
-        if sum(unknown_alleles.values()) > 0:
+        tot = sum(unknown_alleles.values())
+        if tot > 0:
             warnings.warn(
-                "The following alleles were not found in the variant_allele array "
-                "and will be treated as unknown:\n"
-                f"{unknown_alleles}"
+                "Ancestral alleles not found in the variant_allele "
```
jeromekelleher, line 2411: What'll happen here if there's thousands of them? The warning may take very long to generate, and may not be helpful. I think just keep it to the overall summary.

hyanwong: The summary is the same order of length, right? As in it was, `{'aaa': 3, 'bbb':4}`, and would now be `'aaa': 3 (x%)\n'bbb': 4 (y%)`.

I'm suggesting this because I would find it useful to know (a) the proportion of bad ancestral alleles and (b) if I accidentally wasn't e.g. converting them to uppercase etc in the right way. I think reporting this would help avoid lots of accidentally bad analysis, which is why we introduced it (see https://github.com/tskit-dev/tsinfer/issues/870 and https://github.com/tskit-dev/tsinfer/issues/683#issuecomment-1163465893. But happy to be guided by user experience.

I've been testing on the default 1000 genomes VCFs, and seeing what would be useful.

jeromekelleher: Ah, I didn't understand the code.

Can you do the computations outside of the string then? This is unnecessarily terse

hyanwong: Refactored to move the calculations above the string construction.

jeromekelleher (approved):

PR #889, opened 2024-01-20
Check duplicate positions

### LINKED ISSUE #888: SgkitSampleData allows duplicate positions
It's common to get duplicate positions in VCF files, and these can be reasonably read in to SGKit. I think we should fail with a sensible error if we detect duplicate positions on conversion to an SgkitSampleData file, though. 

Edit - should fail if neither duplicate site is masked out.

hyanwong:
Fixes #888

### DIFF
--- a/tsinfer/formats.py
+++ b/tsinfer/formats.py
@@ -2309,6 +2309,12 @@ class SgkitSampleData(SampleData):
                     " sgkit dataset, indicating that all the genotypes are"
                     " unphased"
                 )
+        if np.any(np.diff(self.sites_position) <= 0):
+            raise ValueError(
+                "Values taken from the variant_position array are not strictly "
+                "increasing (i.e. have duplicate or out-of-order values). "
+                "These must be masked out to run tsinfer."
+            )
 
     @functools.cached_property
     def format_name(self):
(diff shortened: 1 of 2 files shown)


hyanwong:
Should probably allow duplicate positions if they are masked

tsinfer/formats.py
```diff
@@ -2309,6 +2309,11 @@ def __init__(self, path):
                     " sgkit dataset, indicating that all the genotypes are"
                     " unphased"
                 )
+        if np.unique(self.sites_position).shape[0] != self.sites_position.shape[0]:
```
jeromekelleher, line 2312: There's no need for a unique here, that's much more expensive than necessary. Can just say the np.diff must be greater than 0

hyanwong: Oh, good catch. Yes, your suggested check should be fine. I have just updated the PR

hyanwong:
@mergifyio rebase

benjeffery:
@mergifyio rebase

benjeffery (approved):

benjeffery:
@mergifyio rebase

benjeffery:
@mergifyio rebase

PR #896, opened 2024-02-02
Add sample mask

benjeffery:
Needs a couple of extra tests for weird masks, but mostly there.

### DIFF
--- a/tsinfer/formats.py
+++ b/tsinfer/formats.py
@@ -2693,6 +2693,11 @@ class SgkitSampleData(SampleData):
 
     def _slice_haplotypes(self, samples_slice, sites=None, recode_ancestral=None):
         # Note that this function loads the entire slice into RAM.
+
+        # Error if samples_slice is not multiple of ploidy
+        if samples_slice[0] % self.ploidy != 0 or samples_slice[1] % self.ploidy != 0:
+            raise ValueError("Slice must be a multiple of ploidy")
+
         start, stop = samples_slice
         if recode_ancestral is None:
             recode_ancestral = False
@@ -2704,8 +2709,8 @@ class SgkitSampleData(SampleData):
         # Calculate individual's start and stop points based on ploidy
         indiv_start = start // self.ploidy
         indiv_stop = (stop + self.ploidy - 1) // self.ploidy  # round up the division
-
         individuals_index = np.where(self.individuals_mask)[0]
+
         unmasked_start = individuals_index[indiv_start]
         unmasked_stop = individuals_index[indiv_stop - 1] + 1
         gt_slice = gt[:, unmasked_start:unmasked_stop, :]
--- a/tsinfer/formats.py
+++ b/tsinfer/formats.py
@@ -2320,6 +2320,60 @@ class SgkitSampleData(SampleData):
                 "These must be masked out to run tsinfer."
             )
 
+        # Create zarr arrays for convenience when iterating over chunks
+        self.z_sites_mask = zarr.array(
+            self.sites_mask, chunks=self.data["call_genotype"].chunks[0], dtype=bool
+        )
+        self.z_individuals_mask = zarr.array(
+            self.individuals_mask,
+            chunks=self.data["call_genotype"].chunks[1],
+            dtype=bool,
+        )
+
+        # Find the first chunk from the left and right that contains an unmasked site
+        self.sites_first_chunk = None
+        self.sites_last_chunk = 0
+        for sites_chunk in range(self.z_sites_mask.cdata_shape[0]):
+            if np.sum(self.z_sites_mask.blocks[sites_chunk]) > 0:
+                if self.sites_first_chunk is None:
+                    self.sites_first_chunk = sites_chunk
+                self.sites_last_chunk = sites_chunk + 1
+        self.sites_first_chunk = (
+            0 if self.sites_first_chunk is None else self.sites_first_chunk
+        )
+
+        # Same for individuals
+        self.individuals_first_chunk = None
+        self.individuals_last_chunk = 0
+        for individuals_chunk in range(self.z_individuals_mask.cdata_shape[0]):
+            if np.sum(self.z_individuals_mask.blocks[individuals_chunk]) > 0:
+                if self.individuals_first_chunk is None:
+                    self.individuals_first_chunk = individuals_chunk
+                self.individuals_last_chunk = individuals_chunk + 1
+
+        logging.info(f"Number of sites after applying mask: {self.num_sites}")
+        logging.info(
+            f"Sites chunk range: {self.sites_first_chunk} - {self.sites_last_chunk}"
+            f"of {self.z_sites_mask.cdata_shape[0]}"
+        )
+        logging.info(
+            f"Number of individuals after applying mask: {self.num_individuals}"
+        )
+        logging.info(
+            f"Individuals chunk range: {self.individuals_first_chunk} - "
+            f"{self.individuals_last_chunk} of {self.z_individuals_mask.chunks[0]}"
+        )
+        haplo_mem_use = (
+            (self.sites_last_chunk - self.sites_first_chunk)
+            * self.z_sites_mask.chunks[0]
+            * self.z_individuals_mask.chunks[0]
+            * self.ploidy
+        )
+        logging.info(
+            f"Memory use for loading haplotypes (match_samples): "
+            f"{humanize.naturalsize(haplo_mem_use)}bytes"
+        )
+
     @functools.cached_property
     def format_name(self):
         return self.FORMAT_NAME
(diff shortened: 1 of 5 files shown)


jeromekelleher (commented):
Nice. I think the chunk_iterator could be a bit simpler and more generic by having a ``mask=[dim0mask, dim1mask]`` argument, and each dimension mask defaults to None (np.ones in that dimension), but we can log that as a follow up issue.

hyanwong:
Did we ever get to the root of the knotty question of whether to try to make the meaning of mask=1 and mask=0 the same between SGkit and `tsinfer`? I worry that they have opposite meanings at the moment. (SGkit sets `1` to mask *out* something, whereas our approach has been to use `1` to include a region. What convention do other bioinformatics pipelines use?

jeromekelleher:
Good point. We want to follow sgkit, no point in thinking any harder than that.

benjeffery:
> Did we ever get to the root of the knotty question of whether to try to make the meaning of mask=1 and mask=0 the same between SGkit and `tsinfer`? I worry that they have opposite meanings at the moment. (SGkit sets `1` to mask _out_ something, whereas our approach has been to use `1` to include a region. What convention do other bioinformatics pipelines use?

Yeah, I plan to flip the masks. Will file an issue for it.

benjeffery:
Would appreciate a quick review here. Seems to all be working at GeL and BMRC.

tsinfer/formats.py
```diff
@@ -2333,6 +2335,19 @@ def sequence_length(self):
...
+        try:
+            return self.data["samples_mask"][:].astype(bool)
+        except KeyError:
+            return np.full(self._num_unmasked_individuals, True, dtype=bool)
+
+    @functools.cached_property
+    def samples_mask(self):
+        # Samples in sgkit are individuals in tskit, so we need to expand
+        # the mask to cover all the samples for each individual.
+        return np.repeat(self.individuals_mask, self.ploidy)
```
hyanwong, line 2349: Instead of making a new cached array, can't you broadcast a view into the `individuals_mask` instead, so you don't need to make a copy?

hyanwong: Hmm, maybe not actually, since this is row-wise. Ignore me.

tsinfer/formats.py
```diff
@@ -2297,9 +2299,9 @@ def __init__(self, path):
         self.path = path
         self.data = zarr.open(path, mode="r")
         genotypes_arr = self.data["call_genotype"]
-        _, self._num_individuals, self.ploidy = genotypes_arr.shape
+        _, self._num_unmasked_individuals, self.ploidy = genotypes_arr.shape
```
hyanwong, line 2302: I interpreted "unmasked" as the number of individuals without the mask set, but instead it's the total number of individuals before masking. Would a better name be e.g. `_total_num_individuals` or `num_individuals_premask`, or something else maybe (also below for `_num_unmasked_samples`

tsinfer/formats.py
```diff
@@ -2445,9 +2460,9 @@ def provenances_record(self):
         except KeyError:
             return np.array([], dtype=object)
 
-    @property
+    @functools.cached_property
     def num_samples(self):
```
hyanwong, line 2464: Is it worth documenting this as the number of samples that have not been masked out (equivalent to the total number of samples in the dataset if there is no masking)?

tsinfer/formats.py
hyanwong: Although I can't actually see anywhere that `samples_mask` is used in the code (although it is in the tests)? m I missing something?

jeromekelleher:
Would it be simpler to assume that there is *always* an individual and site mask, and just try to preserve the old API with those imposed? From tsinfer's perspective, we don't care about stuff that has been masked out, and if you want to look at the raw data you go back to sgkit.

So, ``num_individuals`` and ``num_sites`` etc is what is left in the dataset after we've masked out stuff, and the actual masks are internal implementation details.

tsinfer/formats.py
benjeffery: Yes, this isn't used directly, but thought it would be useful thing to have.

jeromekelleher:
I think this is a nice pattern we could adopt here for the general iteration: https://github.com/pystatgen/vcf-zarr-publication/blob/a01de00e36d0918a7e47fb2f8c6b3a4fd810eb66/src/zarr_afdist.py#L110

So, for iterating over haplotypes we'd do:
```python
    # Use zarr arrays to get mask chunks aligned with the main data
    # for convenience.
    z_variant_mask = zarr.array(
        variant_mask, chunks=call_genotype.chunks[0], dtype=np.int8
    )
    for v_chunk in range(call_genotype.cdata_shape[0]):
        variant_mask_chunk = z_variant_mask.blocks[v_chunk]
        count = np.sum(variant_mask_chunk)
        if count > 0:
            v_chunk = call_genotype.blocks[v_chunk]
            for row in v_chunk:
                   yield row[sample_mask]
```

benjeffery:
This pretty much done - the test failure is odd, can't immediately recreate so will make the exact env that is failing.

tsinfer/formats.py
```diff
@@ -2333,6 +2356,26 @@ def sequence_length(self):
...
         return self._num_sites
 
+    @functools.cached_property
+    def individuals_mask(self):
+        if self._sgkit_samples_mask_name is None:
+            return np.full(self._num_unmasked_individuals, True, dtype=bool)
+        else:
+            try:
+                # We negate the mask as it is much easier in numpy to have True=keep
+                return ~(self.data[self._sgkit_samples_mask_name][:].astype(bool))
```
jeromekelleher, line 2366: It's pretty confusing here to have ``mask`` being used in both the positive and negative sense - so an input mask is False to keep, but the masks we use internally are True to keep. I think we need to use different words here.

So, a mask is a boolean array which is true if the corresponding item should be *excluded*.

How about a "select" array is boolean array which is true if the corresponding item should be kept. Then, ``x_select`` and ``x_mask`` arrays are logical inverses.

jeromekelleher (commented):
LGTM, but I think we need to fix the terminology here as it's horribly confusing. Let's change all the things that we currently have as ``x_mask`` to ``x_select``, and reserve the work "mask" to specifically mean "mask something out if true". Also change ``unmasked_x`` to ``selected_x``, I think would be a lot easier to follow.

tsinfer/formats.py
benjeffery: Yeah, good point.

hyanwong: Yeah, great call @jeromekelleher . I would find this much easier to follow too.

benjeffery:
Ok I've done the mask renaming.
Still unable to recreate the odd failure we're getting `AttributeError: 'dict' object has no attribute 'astype'` deep in some zarr code. Will try rolling back zarr.

tsinfer/formats.py
```diff
@@ -2348,46 +2390,52 @@ def sites_metadata(self):
...
-        try:
-            if (
-                self.data["variant_mask"].shape[0]
-                != self.data["variant_position"].shape[0]
-            ):
+    def sites_select(self):
+        if self._sites_mask_name is None:
+            return np.full(self.data["variant_position"].shape, True, dtype=bool)
+        else:
+            try:
```
jeromekelleher, line 2418: Basically all of the logic for this method could be moved to the ``__init__``, couldn't it? That would make it possible to catch these kinds of errors at init time rather than later on.

Why not also just store the ``sites_select`` array then rather than faffing with a cached_property? This is a read-only view, isn't it?

tsinfer/formats.py
```diff
@@ -2333,6 +2355,26 @@ def sequence_length(self):
     def num_sites(self):
         return self._num_sites
 
+    @functools.cached_property
+    def individuals_select(self):
+        if self._sgkit_samples_mask_name is None:
```
jeromekelleher, line 2360: Same comment as sites_select - can we just compute and store this at init time?

tsinfer/formats.py
```diff
@@ -305,7 +305,7 @@ def zarr_summary(array):
     return ret
 
 
-def chunk_iterator(array, indexes=None, mask=None, dimension=0):
+def chunk_iterator(array, indexes=None, mask=None, orthogonal_select=None, dimension=0):
```
jeromekelleher, line 308: mask should be ``select`` here, it's being used in the wrong sense currently.

jeromekelleher (commented):
LGTM, a couple of simplifications and minor comment. Good to merge then.

benjeffery:
@mergifyio rebase

tsinfer/formats.py
benjeffery: Fixed in https://github.com/tskit-dev/tsinfer/pull/896/commits/56df80090c8fcb36bc18ca192f97df063d4f51b2

tsinfer/formats.py
benjeffery: Fixed in https://github.com/tskit-dev/tsinfer/pull/896/commits/56df80090c8fcb36bc18ca192f97df063d4f51b2

tsinfer/formats.py
benjeffery: Fixed in https://github.com/tskit-dev/tsinfer/pull/896/commits/3fdf3b9b1df88a09a6506d015e34b7dce13ab49d

benjeffery:
Comments addressed.

jeromekelleher (approved):

PR #917, opened 2024-05-14
Batch ancestor matching

benjeffery:
Stacked on #896


tsinfer/inference.py
```diff
@@ -1434,17 +1431,30 @@ def __init__(
...
-                else 100 * 1024 * 1024 * 1024,
-            )
-        else:
-            self.resume_lmdb = None
+
+        self.match_data = collections.defaultdict(dict)
+        self.match_data_dir = match_data_dir
+        if match_data_dir is not None:
+            os.makedirs(match_data_dir, exist_ok=True)
+            for file in os.listdir(match_data_dir):
```
jeromekelleher, line 1439: Probably good to abstract this out into a function for testability. It takes a path and returns a dict of stuff.

tsinfer/inference.py
```diff
@@ -1434,17 +1431,30 @@ def __init__(
...
+                with open(os.path.join(match_data_dir, file), "rb") as f:
+                    (group, num_sites), batch_results = pickle.load(f)
```
jeromekelleher, line 1441: It's worthwhile making dataclass for this payload, we'll inevitably need to add more things to it.

jeromekelleher (commented):
LGTM. I'm in favour of dropping LMBD for simpler file-system based methods :+1:

hyanwong:
> I'm in favour of dropping LMBD for simpler file-system based methods 👍

If we do this, will it be possible to run a lightweight version of tsinfer on pyodide? Since pyodide already includes `zarr`, I'm thinking that maybe it's only LMBD that's the blocker on getting tsinfer to run in-browser for tutorials etc?

jeromekelleher:
Yes, seems likely. We'd have to drop the original SampleData format though.

benjeffery:
The latest commit implements part of the scheme suggested at #921. Remaining is the partition matching function.

tsinfer/inference.py
```diff
@@ -584,6 +549,228 @@ def match_ancestors(
...
+        current_partition = []
+        current_partition_work = 0
+        # TODO: Can do better here by packing ancestors
+        # into as equal sized partitions as possible
+        if group_index == 0:
+            partitions.append(group_ancestors)
+        else:
+            total_work = sum(ancestor_lengths[ancestor] for ancestor in group_ancestors)
+            min_work_per_job_group = min_work_per_job
+            if total_work / 1000 > min_work_per_job:
```
jeromekelleher, line 602: Probably good to cover this somehow, it's the sort of thing that would catch us out in real applications. Could make the ``1000`` a parameter, and then test with something small?

tsinfer/inference.py
```diff
@@ -584,6 +549,228 @@ def match_ancestors(
...
+                    partitions.append(current_partition)
+                    current_partition = [ancestor]
+                    current_partition_work = ancestor_lengths[ancestor]
+                else:
+                    current_partition.append(ancestor)
+                    current_partition_work += ancestor_lengths[ancestor]
+            partitions.append(current_partition)
+        # Make directories for the path data
+        if len(partitions) > 1:
+            os.mkdir(os.path.join(working_dir, f"group_{group_index}"))
```
jeromekelleher, line 618: I don't see the attraction of mixing os.path/os.mkdir etc with pathlib. I would make sure ``working_dir`` is a pathlib at the top of the function, and then just use pathlib methods on it:
```
     group_dir = working_dir / f"group_{group_index}
     group_dir.mkdir()
```

tsinfer/inference.py
```diff
@@ -584,6 +549,228 @@ def match_ancestors(
...
+        group = {
+            "ancestors": group_ancestors,
+            "partitions": partitions if len(partitions) > 1 else None,
+        }
+        ancestor_grouping.append(group)
+
```
jeromekelleher, line 624: We should implement this as a dataclass at some point, as this will be fragile across version changes etc. Dict is fine for now, but let's add a comment for future references.

jeromekelleher (commented):
Generally looks good. A couple of high-level things:

- The mix of pathlib and os.path/os functions is weird. I would suggest embracing pathlib by doing ``some_path = pathlib.Path(some_path)`` at the top of a function that takes a pathlike as an argument, and then exclusively using pathlib functions. It leads to much more readable code. (for new code, old code can stay as is)
- There's some stuff that's not tested, mostly error conditions. Good to cover these.

benjeffery:
Addressed the comments here - believe this is ready to merge.

benjeffery:
Comments addressed. I've added a bit more testing - some of the missing coverage in error cases is in code paths that will be removed by the sample matching batch refactor.

jeromekelleher:
Merge away

PR #930, opened 2024-06-06
Allow the same scheme name, if description matches

### LINKED ISSUE #929: Overwriting md schema
We've had problems in the past when taking tree seqs that are already inferred then trying to infer them again, because if the md schema name already exists, we bomb out, here:

https://github.com/tskit-dev/tsinfer/blob/1d45c0c8122d0680cee2fcfd4b23c7dbbcb7a497/tsinfer/inference.py#L118

I suggest that we add a description to the md schema field (e.g. for "sample_data_id", "sample_data_time" etc.) and if we detect that the description *also* matches in the existing metadata we then overwrite with a warning. The justification is that if the description matches (which presumably contains the word "tsinfer") then we are simply stomping on our own data.

hyanwong:
Fixes #929

### DIFF
--- a/tests/test_inference.py
+++ b/tests/test_inference.py
@@ -4533,9 +4533,37 @@ class TestAddToSchema:
 
     def test_name_collision(self):
         schema = MetadataSchema.permissive_json().schema
+        schema = tsinfer.add_to_schema(
+            schema, "name", definition={"type": "number", "description": "something"}
+        )
+        with pytest.raises(ValueError):
+            tsinfer.add_to_schema(
+                schema, "name", definition={"type": "number", "description": "alt"}
+            )
+
+    def test_name_collision_no_definition(self):
+        schema = MetadataSchema.permissive_json().schema
         schema = tsinfer.add_to_schema(schema, "name")
         with pytest.raises(ValueError):
             tsinfer.add_to_schema(schema, "name")
+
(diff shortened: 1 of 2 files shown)


hyanwong:
Happy to use permissive_json, but don't we still want to warn if we are stomping on existing data. Having a defined schema is a nice way to do that.

hyanwong:
Also, it's useful in tsdate because we are using short, non-intuitive names ("mn" and "vr") for the posterior means and variances, so documenting them is pretty useful IMO.

jeromekelleher:
Sure, whatever is the shortest path here. Note that the schema is only one thing - if you want to truly check that you're not clobbering then you'd have to check the actual metadata items on the way in, because the existing schema may not declare all the fields.

jeromekelleher (approved):

hyanwong:
Thanks. I think we should allow clobbering, but only if it's clear that the field was originally put in by tsinfer anyway. Checking the description is a reasonable way to do that (but actually, we should check the whole definition field, so I'll adjust this PR.). I think we warn when clobbering if the definition matches, but error out if the definition does not match.

benjeffery (approved):
I'm happy with throwing a warning in this case.

hyanwong:
> I'm happy with throwing a warning in this case.

Great. Just made a minor change. Ready to review and hopefully merge @benjeffery ?

PR #936, opened 2024-07-11
Upgrade actions, build with numpy2

benjeffery:


setup.py
```diff
@@ -34,24 +31,17 @@
...
     libraries=libraries,
-    # Enable asserts by default.
     undef_macros=["NDEBUG"],
     include_dirs=includes + [numpy.get_include()],
 )
 
 setup(
-    # The package name along with all the other metadata is specified in setup.cfg
-    # However, GitHub's dependency graph can't see the package unless we put this here.
-    name="tsinfer",
```
jeromekelleher, line 55: Has this issue been resolved on GitHub?

pyproject.toml
```diff
@@ -3,10 +3,79 @@ requires = [
...
+]
+dependencies = [
+    "numpy>=1.23.5",
+    "six",
+    "tqdm",
+    "humanize",
+    "daiquiri",
+    "tskit>=0.5.3",
+    "numcodecs>=0.6",
+    "zarr>=2.2,!=2.11.0,!=2.11.1,!=2.11.2",
```
jeromekelleher, line 58: Need to pin zarr < 3

setup.py
benjeffery: https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/about-the-dependency-graph#supported-package-ecosystems says it should be in pyproject.toml, so will remove.

benjeffery:
We need numpy<2 for tests here due to `numba` so I think the solution is to build with numpy2, then test with 1.x.

benjeffery:
Had to add another conda channel to get `cbgen` on 3.12. This doesn't have windows though, so skipping that test combo.

jeromekelleher:
> channel to get cbgen on 3.12.

Do we need cbgen? We should only need the minimal sgkit install, not any of the IO packages.

benjeffery:
> > channel to get cbgen on 3.12.
> 
> Do we need cbgen? We should only need the minimal sgkit install, not any of the IO packages.

If we skip sgkit in the conda install and do it by pip that might be possible. I don't see how to skip the IO packages with conda.

.circleci/config.yml
```diff
@@ -115,6 +115,7 @@ jobs:
           command: |
             python -m venv venv
             source venv/bin/activate
+            pip install numpy>2
```
jeromekelleher, line 118: >=

PR #943, opened 2024-07-23
Numpy2 changes

hyanwong:
With this, almost all tests pass under numpy2, except for one of the 8-bit conversion routines. Do you know what's going on here @benjeffery ?

```python
    def get_site_genotypes_subset(self, site_id, samples):
        start = site_id * self.encoded_genotypes_size
        g = np.zeros(len(samples), dtype=np.int8)
        if self.genotype_encoding == constants.GenotypeEncoding.ONE_BIT:
            for j, u in enumerate(samples):
                byte_index = u // 8
                bit_index = u % 8
                byte = self.genotype_store[start + byte_index]
                mask = 1 << bit_index
                g[j] = int((byte & mask) != 0)
        else:
            for j, u in enumerate(samples):
>               g[j] = self.genotype_store[start + u]
>> E               OverflowError: Python integer 255 out of bounds for int8

```

### DIFF
--- a/tsinfer/algorithm.py
+++ b/tsinfer/algorithm.py
@@ -111,7 +111,8 @@ class AncestorBuilder:
                 g[j] = int((byte & mask) != 0)
         else:
             for j, u in enumerate(samples):
-                g[j] = self.genotype_store[start + u]
+                # NB missing data (-1) is stored as 255 in the genotype_store
+                g[j] = self.genotype_store[start + u].astype(np.int8)
         gp = self.get_site_genotypes(site_id)
         np.testing.assert_array_equal(gp[samples], g)
         return g
@@ -129,6 +130,8 @@ class AncestorBuilder:
         if self.genotype_encoding == constants.GenotypeEncoding.ONE_BIT:
             assert np.all(genotypes >= 0) and np.all(genotypes <= 1)
             genotypes = np.packbits(genotypes, bitorder="little")
+        else:
+            assert np.all(genotypes <= 127)
         start = site_id * self.encoded_genotypes_size
         stop = start + self.encoded_genotypes_size
         self.genotype_store[start:stop] = genotypes
--- a/tsinfer/constants.py
+++ b/tsinfer/constants.py
@@ -21,6 +21,8 @@ Collection of constants used in tsinfer. We also make use of constants defined i
 """
 import enum
 
+import numpy as np
+
 C_ENGINE = "C"
 PY_ENGINE = "P"
 
@@ -28,16 +30,16 @@ PY_ENGINE = "P"
 # TODO Change these to use the enum.IntFlag class
 
 # Bit 16 is set in node flags when they have been created by path compression.
-NODE_IS_PC_ANCESTOR = 1 << 16
+NODE_IS_PC_ANCESTOR = np.uint32(1 << 16)
 # Bit 17 is set in node flags when they have been created by shared recombination
 # breakpoint
-NODE_IS_SRB_ANCESTOR = 1 << 17
+NODE_IS_SRB_ANCESTOR = np.uint32(1 << 17)
 # Bit 18 is set in node flags when they are samples inserted to augment existing
 # ancestors.
-NODE_IS_SAMPLE_ANCESTOR = 1 << 18
+NODE_IS_SAMPLE_ANCESTOR = np.uint32(1 << 18)
 # Bit 20 is set in node flags when they are samples not at time zero in the sampledata
 # file
-NODE_IS_HISTORICAL_SAMPLE = 1 << 20
+NODE_IS_HISTORICAL_SAMPLE = np.uint32(1 << 20)
 
 # What type of inference have we done at a site?
 INFERENCE_NONE = "none"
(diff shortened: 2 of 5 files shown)


hyanwong:
P.s. The np.asarray changes are as a result of this:

```
>       self.data["sites/time"][:] = np.array(value, dtype=np.float64, copy=False)
E       ValueError: Unable to avoid copy while creating an array as requested.
E       If using `np.array(obj, copy=False)` replace it with `np.asarray(obj)` to allow a copy when needed (no behavior change in NumPy 1.x).
E       For more details, see https://numpy.org/devdocs/numpy_2_0_migration_guide.html#adapting-to-changes-in-the-copy-keyword.
```

hyanwong:
Also there is a comment in constants.py "# TODO Change these to use the enum.IntFlag class", but instead I have forced them into the np.uint32 class, so we can compare with the flags array. I'm not sure there's a way to use enum.IntFlag such that the constant values are treated as np.uint32? I think msprime uses enum.IntFlag for some bitflags now, so has it been figured out there, or is this uint+numpy2 thing likely to bite there too?

jeromekelleher:
I think the uint32 thing is more important than making them nice Enums. The msprime enumerations will have the same problems as using standard Python integers.

jeromekelleher:
Looks like we might be assuming uint8 for genotypes currently somewhere, not int8?

hyanwong:
> Looks like we might be assuming uint8 for genotypes currently somewhere, not int8?

Good call. That looks like here: https://github.com/tskit-dev/tsinfer/blob/9d8f9349c953da269d69b1b6a6352f10c7c5abb9/tsinfer/algorithm.py#L90

We can't use np.int8 for the one-bit encoding, because `np.unpackbits` requires `uint`, so I think we'll have to define the `genotype_store` as np.uint8 for the one-bit encoding or np.int8 for the 8-bit one. I'll add that to this PR.

hyanwong:
>  I'll add that to this PR.

Do we need to change anything in the C interface to account for uint vs int genotypes when using 1-bit vs 8-bit packing? I don't think so: it appears as if that just uses a raw list of bytes (see https://github.com/tskit-dev/tsinfer/blob/6b2116de58be2b77a69615cf852b5223f7e4aeea/lib/ancestor_builder.c#L270)

benjeffery:
I'm confused - I thought we needed numpy<2 for numba? Or are you saying that installing >=2 works?

hyanwong:
The most recent numba version is compatible with numpy-2: https://numba.readthedocs.io/en/stable/release/0.60.0-notes.html, but:

> "users should expect some type and numerical deviations with respect to normal Python behavior while using Numba with NumPy 2.0."

It seems to pass the test suite however. I guess we might as well merge this PR in anticipation? Assuming what I've done is sensible.

tsinfer/algorithm.py
```diff
@@ -86,9 +86,11 @@ def __init__(
         self.encoded_genotypes_size = num_samples
         if genotype_encoding == constants.GenotypeEncoding.ONE_BIT:
             self.encoded_genotypes_size = num_samples // 8 + int((num_samples % 8) != 0)
-        self.genotype_store = np.zeros(
-            max_sites * self.encoded_genotypes_size, dtype=np.uint8
-        )
+            self.genotype_store = np.zeros(
+                max_sites * self.encoded_genotypes_size, dtype=np.uint8
+            )
+        else:
```
jeromekelleher, line 92: I don't think it's a good idea to switch between uint8 and int8 - we should use one or the the other consistently.

I think the simplest thing to do is to stick with uint8 internally for the genotype store, but validate on the way in that values are less than <127.

benjeffery (approved):

jeromekelleher (commented):
We should just check that values are less then 127, there's no point in switching dtypes here.

jeromekelleher:
I assume we're silently masking an error here at the moment by wrapping 255 to -127. Much better to update the test involved to not use 255.

hyanwong:
> I assume we're silently masking an error here at the moment by wrapping 255 to -127. Much better to update the test involved to not use 255.

Aren't we using -1 to mean "missing data" (which we don't allow in the 1-bit encoding)? The failing test is `tests/test_inference.py::TestMissingDataRoundTrip::test_simple_example`

tsinfer/algorithm.py
hyanwong: If we stick with uint8, we could preserve `-1` values on output by using `.astype(np.int8)`? E.g.

```python
def get_site_genotypes_subset(self, site_id, samples):
        ...
        else:
            for j, u in enumerate(samples):
                # NB missing data (-1) is stored as 255 in the genotype_store
                g[j] = self.genotype_store[start + u].astype(np.int8)

hyanwong: Yeah, we do check that already. The problem is when we are doing the standard 8-bit encoding, using np.uint8 for the store, sticking a "-1" in there (which weirdly we seem to be able to do just fine, and it gets stored as 255), then pulling it out again.

Is the `astype` workaround acceptable?

hyanwong: (in the latest commit I also added an explicit assert for <=127 for the 8-bit encoding)

jeromekelleher: Ah, we're already casting the output to a int8. Yeah, what you have is good then.

benjeffery:
@mergifyio rebase

Issue #927, opened 2024-06-06
Check all alleles at a site are unique before inference


hyanwong:
I'm finding that tsinfer isn't correctly inferring the genotypes for Sgkit files in the trivial instance below. This seems pretty worrying. What do I have wrong?

```python
import sgkit as sg
import tsinfer

ds = sg.simulate_genotype_call_dataset(n_variant=10, n_sample=4, missing_pct=0, phased=True, seed=1)
ds['variant_allele'] = ds['variant_allele'].astype(str) # hack around https://github.com/tskit-dev/tsinfer/issues/810
ds.update({'variant_ancestral_allele': ds['variant_allele'][:,0]})
ds.to_zarr('/tmp/ds.zarr', mode='w')
sd = tsinfer.SgkitSampleData('/tmp/ds.zarr')
ts = tsinfer.infer(sd)
print("Inferred_genos    Zarr_genotypes    SampleData_genos  Ts alleles")
for v, sgv, sdv in zip(ts.variants(), ds.call_genotype, sd.sites_genotypes):
    bad = "<<<< BAD!!!" if not (v.genotypes == sdv).all() else ""
    print(v.genotypes, sgv.values.flatten(), sdv, v.alleles, bad)
```

Shows that the 7th and 8th site are incorrect:

```
Inferred_genos    Zarr_genotypes    SampleData_genos  Ts alleles
[1 0 1 0 1 0 1 1] [1 0 1 0 1 0 1 1] [1 0 1 0 1 0 1 1] ('G', 'C') 
[0 1 1 0 0 0 0 0] [0 1 1 0 0 0 0 0] [0 1 1 0 0 0 0 0] ('G', 'A') 
[1 0 1 0 1 1 0 0] [1 0 1 0 1 1 0 0] [1 0 1 0 1 1 0 0] ('T', 'A') 
[1 0 1 1 1 0 1 1] [1 0 1 1 1 0 1 1] [1 0 1 1 1 0 1 1] ('G', 'A') 
[1 1 1 1 0 1 0 0] [1 1 1 1 0 1 0 0] [1 1 1 1 0 1 0 0] ('C', 'G') 
[0 0 1 1 1 0 0 1] [0 0 1 1 1 0 0 1] [0 0 1 1 1 0 0 1] ('G', 'A') 
[0 0 0 0 0 0 0 0] [0 1 0 1 1 0 0 1] [0 1 0 1 1 0 0 1] ('T',) <<<< BAD!!!
[...]

PR #944, opened 2024-07-24
VariantData

benjeffery:


benjeffery:
@hyanwong The API is currently:
```python
VariantData(
        path,
        ancestral_allele_name_or_array="variant_ancestral_allele",
        sample_mask_name_or_array=None,
        site_mask_name_or_array=None,
        sites_time_name_or_array=None,
    )
```

I just have to add a few more tests, but does this look good to you?

hyanwong:
We might want to swap e.g. new array masks or flip some ancestral states. Would we jut create a new VariantData object in that case, or would there be the ability to edit the one we have already, e.g. via

```
vd = VariantData(...)
vd.set_ancestral_allele(my_numpy_array)
```

Incidentally, although VariantData is a reasonable name, I'm not sure about using "vd" for the shorthand object 😬

hyanwong:
I'm just looking at the [tutorial page](https://tskit.dev/tsinfer/docs/latest/tutorial.html#toy-example). In this, we invent a dataset, and use the `SampleData.add_site()` function to add data that can be read by tsinfer. Is there a way to do this using the new input format (i.e. to create a VCF Zarr file from scratch, without starting with a VCF)? Or do we keep the old interface around for teaching purposes (I would prefer not to, I think).

benjeffery:
> I'm just looking at the [tutorial page](https://tskit.dev/tsinfer/docs/latest/tutorial.html#toy-example). In this, we invent a dataset, and use the `SampleData.add_site()` function to add data that can be read by tsinfer. Is there a way to do this using the new input format (i.e. to create a VCF Zarr file from scratch, without starting with a VCF)? Or do we keep the old interface around for teaching purposes (I would prefer not to, I think).

The dataset is just a set of arrays, so you can create one from scratch or use `sgkit.simulate_genotype_call_dataset` for a simple example dataset.

hyanwong:
Since you are writing tests, do you have time to roll in a check for duplicate alleles (https://github.com/tskit-dev/tsinfer/issues/927) @benjeffery ? That would make using `simulate_genotype_call_dataset` less prone to giving weird results in the tutorial.

hyanwong:
How do we specify an unknown ancestral allele? I'm guessing we provide any non-allele character, e.g. "." in the input array?

hyanwong:
Thanks. Incidentally, I wonder if we should transparently convert an array of numpy byte strings as ancestral alleles into "proper" strings, by calling `.astype(str)` on the AA array? That saves some hassle with `sg.simulate_genotype_call_dataset` which makes byte strings

hyanwong:
How do we specify the AA on the command-line, by the way?

benjeffery:
> I wonder if we should transparently convert an array of numpy byte strings as ancestral alleles into "proper" strings, by calling `.astype(str)` on the AA array?

That's done already.



> How do we specify the AA on the command-line, by the way?

There is no command line support for VariantData yet.

hyanwong:
The name "ancestral_allele_name_or_array" is quite long. That's OK, but I wonder if we just get people to use positional arguments, or if we want to enforce using the actual parameter name. The following looks OK to me:

```python
vdata = tsinfer.VariantData("data.vcz", ancestral_alleles)
```

benjeffery:
> The name "ancestral_allele_name_or_array" is quite long.

The current code supports positional for the AA.

hyanwong:
> > The name "ancestral_allele_name_or_array" is quite long.
> 
> The current code supports positional for the AA.

Great, as long as we are happy recommending that, all's good. Thanks

tsinfer/formats.py
```diff
@@ -2292,56 +2292,67 @@ def populations(self):
...
+class VariantData(SampleData):
 
     FORMAT_NAME = "tsinfer-sgkit-sample-data"
     FORMAT_VERSION = (0, 1)
 
-    def __init__(self, path, sgkit_samples_mask_name=None, sites_mask_name=None):
+    def __init__(
+        self,
+        path,
+        ancestral_allele_name_or_array,
```
jeromekelleher, line 2303: The ``_name_or_array`` suffix is redundant for all these isn't it? We don't put type information into most things, and since these all share the same suffix I don't see what problem it solves.

jeromekelleher (commented):
Looks good as the top-level API, but I think we can shorten the names without loss of information. Should be simple global search and replace job.

tsinfer/formats.py
benjeffery: Good point - fixed

PR #945, opened 2024-07-26
Update tutorial for new VariantData format

hyanwong:
First pass at changing the tutorial to match the new VariantData format. Based off #944 

~I haven't sorted out the population specification in the sparrows data example yet~


docs/tutorial.md
```diff
@@ -23,72 +23,86 @@ kernelspec:
...
+[VCF Zarr](https://github.com/sgkit-dev/vcf-zarr-spec/) (.vcz) format. The standard
+route to create such a file is by conversion from a VCF file, e.g. using
+[vcf2zarr](https://sgkit-dev.github.io/bio2zarr/vcf2zarr/overview.html) as described later in this
+document. For a quick introduction, however, we will instead create an example file using
+[sgkit](https://sgkit-dev.github.io/sgkit/latest/).
+
 
 ```{code-cell} ipython3
+import sgkit
+ds = sgkit.simulate_genotype_call_dataset(n_variant=8, n_sample=3, missing_pct=0, phased=True, seed=79)
```
jeromekelleher, line 36: I don't think we want to use sgkit apis, and certainly not simulate_genotype_call_dataset as it's fragile and not particularly useful. It would be better to have an example VCF and import it using vcf2zarr.

jeromekelleher (commented):
I haven't gone through the details, but I think starting from a simple VCF example is much more useful that the original alignment. Everyone knows what VCF looks like now.

docs/tutorial.md
hyanwong: OK, but we do that below. This bit of the tutorial was intended to introduce people to the tsinfer concept without having to deal with the ickiness of VCFs, I thought?

IMO it's better not to talk about specific formats like VCF, Plink, BGEN etc at the start, but keep it general? We could make our own zarr arrays, I suppose, but that's getting bogged down too quickly into detail.

hyanwong: Also FWIW we currently round-trip the msprime data via VCF, but I'm hoping we can bin that once tskit2zarr is released, as it brings all sorts of extra dependencies (bgzip writer, tabix utility, etc).

hyanwong: OK, In https://github.com/tskit-dev/tsinfer/pull/945/commits/319619b0e72ce5eac81f50d330f666fbb8221f23 I removed sgkit and simply load up an existing .vcz file at the start

jeromekelleher:
Using existing data set is a good call. The tooling isn't there yet, I don't think there's much point in trying to force it through for this PR.

hyanwong:
> I haven't gone through the details, but I think starting from a simple VCF example is much more useful that the original alignment. Everyone knows what VCF looks like now.

Hmm, is that true? I don't really know what a VCF looks like (and I don't really want to learn about it either). At least for teaching, it's clearer to me to start with a set of alignments (i.e. haplotype data), not site-by-site data, which as a biologist, I always find slightly more confusing to deal with. For example, when storing viral genomes, I think of the sequence of the virus, not the variation at each site. Surely that's more of a natural thing to think about, biologically?

Anyway, the SGkit stuff is removed now, and I think it all still reads fine. Maybe we can merge (once it's building properly) and iterate from there? It would be useful for teaching to have examples which work using the new input format.

hyanwong:
*Finally* got this to build: that was a bit of a marathon. I think we can merge this, and also close https://github.com/tskit-dev/tsinfer/pull/865 which I had to roll into this PR to get it to build properly.

hyanwong:
P.s. happy to squash all these commits together, if that helps.

benjeffery (approved):

PR #956, opened 2024-08-28
Variant data

hyanwong:
Add example of excluding chromosomes using site masks, and change `zarr.load()` to the more recommended `zarr.open()`.

We have had some users try to infer with VCFs containing multiple chromosomes, so the examples here are useful.


jeromekelleher (approved):
LGTM

PR #959, opened 2024-09-04
Experimental hmm

jeromekelleher:
Just opening for visibility and discussion here, not for merging. I've based the changes on 0.3.3 for simplicity here as sc2ts isn't compatible with some of the internal changes in tsinfer (yet). I'll cherry pick these changes onto another PR if we decide to implement this for real (probably adding an option to rescale or not by n)

A quick implementation of the changes discussed in https://github.com/jeromekelleher/sc2ts/issues/242 where we want to simplify the HMM and make it easier to reason about which Viterbi paths we care about.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ CFLAGS=-std=c99 -g -O3 -march=native -funroll-loops -ffast-math \
 all: _tsinfer.cpython-34m.so 
 
 _tsinfer.cpython-34m.so: _tsinfermodule.c 
-	CC="${CC}" CFLAGS="${CFLAGS}" python setup.py build_ext --inplace
+	CC="${CC}" CFLAGS="${CFLAGS}" python3 setup.py build_ext --inplace
 
 ctags:
 	ctags lib/*.c lib/*.h tsinfer/*.py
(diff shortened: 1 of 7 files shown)


jeromekelleher:
Note: the only test that's failing is one that's poking into the details of parameter values. Everything else is passing fine.

jeromekelleher:
Some quick notes here on using c11 atomics - getting things to compile on Unix platforms is easy, but Windows is problematic. It seems that MSVC doesn't support stdatomic.h, so I've hacked around it by just not making these variable in question an atomic. In the short term this is fine as we're only using this in sc2ts and this doesn't support Windows. It would raise some questions about whether putting this into tskit is worth the hassle, though.

hyanwong:
I don't know anything about atomics, but https://devblogs.microsoft.com/cppblog/c11-atomics-in-visual-studio-2022-version-17-5-preview-2/ suggests that `stdatomic.h` is coming to windows, and can be enabled in the current MSVC using the `/experimental:c11atomics` switch?

jeromekelleher:
Good to know. I guess that'll filter down to Python builds sometime around 3.15 or so (Python builds against specific MSVC, 14.x here for what we're working with).

benjeffery:
Very nice! I assume you mean just bringing across only the code that removes the requirement?

jeromekelleher:
I think this is something we need to merge in and release as part of tsinfer soon. I think we can expose the core functionality (do you divide by n in the HMM) as a "private" API that's used by sc2ts, and keep the standard tsinfer behaviour as it is by default. There's probably some exploration to be done anyway on what effect this really has on the HMM that @duncanMR will be looking in a few months.

benjeffery:
I believe all this is now merged in other PRs.

Issue #893, opened 2024-01-23
Corner case: cannot encode a SNP next to an indel in standard VCF encoding


hyanwong:
In the VCF format, indels are encoded by using adding the letter *before* the variant to the allele list, and reporting the position of the variant as the indel position minus one (see 5.1.1 in http://samtools.github.io/hts-specs/VCFv4.4.pdf). E.g.:

```

ATGC---CC
ATGCGGGCC
123456789 (position)
```

Would be reported as a variant at position 4, with alleles `"C"` and `"CGGG"`, rather than a variant at position 5 with alleles `""` and `""`. The "C" is taken from the reference genome sequence. However, if there is also a SNP at position 4, we would register two sites, both at position 4. Tskit does not allow sites with duplicate positions, so this situation can't be easily encoded in a tree sequence, even though it's a valid representation of genetic variation.

Other than mask such sites out (if there are any), we could re-encode such sites as at position 5 with one blank allele (`""`). However, this could cause problems for interpreting the SGkit format, as we currently calculate allele numbers etc on the assumption that trailing `""` values in the alleles list can be removed.

This is probably not much of an issue for human data. But in some other datasets, such as _Anopheles_, every other base is a SNP, so I imagine this situation occurs quite often.

hyanwong:
On reflection, I think that the SGkit format should use the VCF style (i.e. use the position one base pair *before* the actual indel, e.g. pos=9, REF=`"ATCT"`, ALT=`"A"`), whereas `tskit` should translate that to a position *at* the indel site (pos=10, REF=`"TCT"`, ALT=`""`). That way we can reasonably encode indels with an adjacent SNP.

PR #960, opened 2024-09-04
Check for duplicate alleles

### LINKED ISSUE #927: Check all alleles at a site are unique before inference

hyanwong:
Fixes #927.

I assume we are happy to take the hit of looping through all the sites and doing a `set()`. An alternative would be to do an `np.unique` by row, but then we would need to remove the duplicate `""` entries.

### DIFF
--- a/tests/test_variantdata.py
+++ b/tests/test_variantdata.py
@@ -675,9 +675,28 @@ def test_sgkit_ancestor(small_sd_fixture, tmp_path):
 
 
 class TestVariantDataErrors:
+    def simulate_genotype_call_dataset(self, *args, **kwargs):
+        # hack around bug in sgkit where duplicate alleles are created
+        # since this is just for testing, it doens't need to be efficient
+        # Force a seed:
+        if "seed" not in kwargs:
+            kwargs["seed"] = 123
+        ds = sgkit.simulate_genotype_call_dataset(*args, **kwargs)
+        variant_alleles = ds["variant_allele"].values
+        allowed_alleles = np.array(
+            ["A", "T", "C", "G", "N"], dtype=variant_alleles.dtype
+        )
+        for row in range(len(variant_alleles)):
+            alleles = variant_alleles[row]
+            if len(set(alleles)) != len(alleles):
+                # Just use a set that we know is unique
+                variant_alleles[row] = allowed_alleles[0 : len(alleles)]
+        ds["variant_allele"] = ds["variant_allele"].dims, variant_alleles
+        return ds
+
     def test_missing_phase(self, tmp_path):
         path = tmp_path / "data.zarr"
-        ds = sgkit.simulate_genotype_call_dataset(n_variant=3, n_sample=3)
+        ds = self.simulate_genotype_call_dataset(n_variant=3, n_sample=3)
         sgkit.save_dataset(ds, path)
         with pytest.raises(
             ValueError, match="The call_genotype_phased array is missing"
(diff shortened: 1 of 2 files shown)


hyanwong:
I'll test. Meanwhile, I messed up the GH commits with another PR, so will fix and force push

benjeffery:
LGTM - I don't think the loop through the alleles will be that expensive for most cases - unless lots of sites have hundreds of alleles.

This has made me realise the vcfzarr spec should probably be more explicit that zero-length alleles are not allowed. It is currently there as implicit as it states `""` is a fill value.

hyanwong:
> This has made me realise the vcfzarr spec should probably be more explicit that zero-length alleles are not allowed. It is currently there as implicit as it states "" is a fill value.

Zero-length alleles are not allowed in VCF files, so I guess that's implicit too. However, we might want to allow representation of indels internally like that, otherwise there are forms of genetic variation that can be represented in a VCF that can't be represented in the tsinfer input (see https://github.com/tskit-dev/tsinfer/issues/893#issuecomment-1923359429)

hyanwong:
Rolled into #963 as part of review, and tested with a large file.

PR #963, opened 2024-09-04
Properly treat blank ancestral alleles

### LINKED ISSUE #962: Do not warn about unknown states if ancestral allele is "N"
In #960 @benjeffery said:

> This has made me realise the vcfzarr spec should probably be more explicit that zero-length alleles are not allowed. It is currently there as implicit as it states "" is a fill value.

In that case, I presume that we can always assume that an ancestral allele of `""` means "intentionally marked as unknown", so that we can recommend that people use that to mark sites of unknown ancestral allele. Furthermore, I suspect that means we can exclude such alleles from the "warning" output that normally says:

```
An ancestral allele was not found in the variant_allele array for the {tot} sites ({frac_bad * 100 :.2f}%) listed below.
```

?

hyanwong:
There was a bug in the previous code: since "" is the fill-value for alleles, if an ancestral allele was "", it wasn't treated as missing.

This PR fixes that bug, but also fixes #962: the string ~""~ "N" is treated as meaning "deliberately set the ancestral allele as unknown". ~This is OK because "" is not a valid VCZ allele string. Therefore if the only ancestral alleles that don't match are ""~, then it's confusing to issue a warning, because setting unknown alleles is a deliberate user choice.

### DIFF
--- a/tsinfer/formats.py
+++ b/tsinfer/formats.py
@@ -31,6 +31,7 @@ import queue
 import sys
 import threading
 import warnings
+from typing import Union  # noqa: F401
 
 import attr
 import humanize
@@ -2091,12 +2092,11 @@ class SampleData(DataContainer):
             ids = np.arange(0, self.num_sites, dtype=int)
         for j in ids:
             anc_idx = ancestral_allele_array[j]
-            alleles = tuple(alleles_array[j])
             site = Site(
                 id=j,
                 position=position_array[j],
                 ancestral_allele=anc_idx,
-                alleles=alleles,
+                alleles=tuple(alleles_array[j]),
                 metadata=metadata_array[j],
                 time=time_array[j],
             )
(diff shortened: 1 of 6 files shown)


benjeffery (approved):
~LGTM, thanks @hyanwong~

I've thought about it more and think that unknown should be explicitly `N` or another character as that is usually the case in FASTA etc. `""` would probably only get into the ancestral allele as a mistake right?

hyanwong:
The problem with "N" is surely that it could match "N" in the actual VCF? It would require extra hardcoded logic *not* to match "N" against existing alleles. It seemed a bit neater to me to pick something that explicitly isn't allowed as an allele in a VCF file.

I guess it's fine hardcoding a specific string as long as it's part of the VCF spec and unlikely to change? I'm wondering why `N` rather than e.g. `.` or `n` or `-`, or whatever all the other conventions are? I'm no VCF expert, so happy to defer to people who are.

Note that inference will find a different ancestral allele from any what we set here.

hyanwong:
Just looking at how to do this. What alleles strings do we want to treat as "unknown ancestral state"? I assume `"N"`, `"n"` and `""`. The VCF spec says nothing about `"."` in the REF and ALT fields (it's only in the GT field).

If we want "`N`" and "`n`" to be the "deliberately unknown" state, we would skip the warning if these were the only non-matching alleles.

hyanwong:
I changed the behaviour to using (and documenting) `N` as the standard unknown ancestral allele. I also wrote the docstring from VariantData, so that this behaviour is described. It would be good to merge this, so that I can start adding documentation for the other changes such as `contig_id` that I've been making.

tsinfer/formats.py
```diff
@@ -2412,16 +2455,18 @@ def __init__(
...
-        converted = np.zeros(self.num_sites, dtype=np.int8)
+        converted = np.full(self.num_sites, -1, dtype=np.int8)
         for i, allele in enumerate(self._sites_ancestral_allele):
-            allele_index = -1
-            try:
-                allele_index = np.where(allele == self.sites_alleles[i])[0][0]
-            except IndexError:
-                unknown_alleles[allele] += 1
-            converted[i] = allele_index
+            if not (allele in {"", "N", "n"}):  # All these must represent unknown
```
jeromekelleher, line 2460: We shouldn't treat empty string as legal input here as it has a defined meaning in VCZ

jeromekelleher (commented):
LGTM, just needs a bit of simplification

hyanwong:
Note that after discussion with @benjeffery, this also changes the parameter name to "ancestral_state" (to match tskit, and so that it is clearer that it differs from the index (which is known as the "ancestral allele"). It also fixes #927. I have tested it on a 200Mb 1000 genomes VCF, and the check does not noticeably increase the construction time of the VariantData object (which in this case was ~0.7 secs)

tsinfer/formats.py
hyanwong: Now done. See comment below about changing the name to "ancestral_state", to match what we used previously.

hyanwong:
Also note that I duplicated the `sites()` constructor so that in the VCF Zarr case, it trims the allele list so that it doesn't include the padding values. This can then be used when checking the ancestral allele states (and is fast)

Issue #763, opened 2022-11-03
sgkit: Use sequence length from dataset


benjeffery:
See also https://github.com/pystatgen/sgkit/issues/464

jeromekelleher:
Added 0.4.0 milestone, but no particular rush with this I think if the feature hasn't been released upstream.

benjeffery:
Fixed by #964

Issue #765, opened 2022-11-03
sgkit: Error out on multiple contigs


benjeffery:
Error should have code snippet for filtering to a single contig

benjeffery:
Fixed by #964

PR #964, opened 2024-09-04
Deal with multiple contigs and sequence lengths

### LINKED ISSUE #949: VariantData sequence_length not being taken from zarr `contig_length` field
I'm having problems getting `VariantData.sequence_length` to match the value in the contig_length field of the zarr file. It wasn't clear until reading the code that I needed to set `z.attrs["sequence_length"]`. If there isn't a `z.attrs["sequence_length"]` defined, should we take the value from `contig_length` instead?

I was having the problem when trying to round-trip data from a tree sequence via a vcf file into another inferred tree sequence, and the sequence lengths didn't match up.

hyanwong:
Introduces a `contig_id` parameter to variant_data, which adds non matching contigs to the site mask, as described in #949. Fixes #949

### DIFF
--- a/tsinfer/formats.py
+++ b/tsinfer/formats.py
@@ -2338,6 +2338,11 @@ class VariantData(SampleData):
         reasonable approximation to the relative order of ancestors used for inference.
         Time values are ignored for sites not used in inference, such as singletons,
         sites with more than two alleles, or sites with an unknown ancestral state.
+    :param int sequence_length: An integer specifying the resulting `sequence_length`
+        attribute of the output tree sequence. If not specified the `contig_length`
+        attribute from the undelying zarr store for the contig of the selected variants.
+        is used. If that is not present then the maximum position plus one of the used
+        variants is used.
     """
 
     FORMAT_NAME = "tsinfer-variant-data"
@@ -2351,7 +2356,11 @@ class VariantData(SampleData):
         sample_mask=None,
         site_mask=None,
         sites_time=None,
+        sequence_length=None,
     ):
+        self._sequence_length = sequence_length
+        self._contig_index = None
+        self._contig_id = None
         try:
             if len(path_or_zarr.call_genotype.shape) == 3:
                 # Assumed to be a VCF Zarr hierarchy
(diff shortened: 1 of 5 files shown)


hyanwong:
Also fixes #763 I think

benjeffery:
Thanks @hyanwong - I've been some thinking on API design here and am thinking a simpler solution is to error out if there is more than one contig in the used variants. 
As for `contig_length`, I don't think this is the same conceptually as tskit `sequence_length`? As in tskit positions have to be less than sequence length, whereas there is no such requirement in vcf zarr, so we risk causing odd (to the user) errors downstream if we use `contig_length`.

hyanwong:
I think checking contig length if present and erroring out early if it is problematic seems fine?

benjeffery:
> people want to be able to take a vcf with multiple chromosomes in it (which is the common input format), select a chromosome, and run inference without too much extra thought

I think this scheme enables that - you `bio2zarr` then write a contig-specific mask, which is a one-liner we can put in the tutorial. I don't want to add things like `contig_id` to the API when they overlap and interact with existing functionality (masking).

andrewkern:
> Thanks @hyanwong - I've been some thinking on API design here and am thinking a simpler solution is to error out if there is more than one contig in the used variants. As for `contig_length`, I don't think this is the same conceptually as tskit `sequence_length`? As in tskit positions have to be less than sequence length, whereas there is no such requirement in vcf zarr, so we risk causing odd (to the user) errors downstream if we use `contig_length`. So how about we check if `contig_length` is greater than the positions and error early if it isn't, and allow the user to override with a `sequence_length` argument. If both are missing use the `largest position+1`.

Having the option to override the `sequence_length` is a feature that I'm looking for in a project that i'm doing right now. I've been using tsinfer to create tree sequences from windowed chunks of a VCF file that I will analyze downstream in a ML pipeline.

hyanwong:
LGTM. Fine to merge, I reckon.

hyanwong:
(p.s. #765 suggests that the error message which currently reads

> "Sites belong to multiple contigs ({contig_names}). "
> "Please restrict sites to one contig using the sites_mask argument."

Should have a code snippet to show users how to do this?

benjeffery:
> Should have a code snippet to show users how to do this?

Good idea, added

benjeffery:
@jeromekelleher This is now ready for your review.

tsinfer/formats.py
```diff
@@ -2351,7 +2356,10 @@ def __init__(
         sample_mask=None,
         site_mask=None,
         sites_time=None,
+        sequence_length=None,
     ):
+        self._sequence_length = sequence_length
+        self._contig_index = None
```
jeromekelleher, line 2362: Need to define ``self._contig_id`` here also, or will get attribute error when not defined.

jeromekelleher (approved):
LGTM, spotted one minor issue

tsinfer/formats.py
benjeffery: Fixed, and I've added checking this to the tests.

PR #967, opened 2024-09-05
Remove mention of sgkit

hyanwong:
Note that this changes the FORMAT_NAME, so is no longer backward-compatible with previous tsinfer data files. But I think that's fine, because we only save the .vcz file normally anyway.

There are still references to sgkit in the test suite, but I think that's fine too.


jeromekelleher (approved):
LGTM

benjeffery (approved):

PR #981, opened 2024-12-05
Do not use mismatch in ancestor matching when using `infer()`

### LINKED ISSUE #980: Do not allow mismatch in ancestor matching when inferring in a a single step
We think using mismatch in ancestor matching is problematic. We should probably not do this by default, therefore, when doing `tsinfer.infer(vdata, recombination_rate=xxx)`. People who really want to do this can use `match_ancestors()` separately.

hyanwong:
Fixes #980

### DIFF
--- a/tests/test_inference.py
+++ b/tests/test_inference.py
@@ -4309,6 +4309,17 @@ class TestMismatchAndRecombination:
                     extended_checks=True,
                 )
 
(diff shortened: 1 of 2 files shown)


jeromekelleher (approved):
LGTM. We should update the changelog to note this change in behaviour (technically breaking change)

PR #990, opened 2025-01-14
Record more tsinfer parameters in provenance

hyanwong:
These are useful when coming to inspect how the tree sequence was inferred (Savita would have found it useful recently).

### DIFF
--- a/tsinfer/inference.py
+++ b/tsinfer/inference.py
@@ -380,6 +380,11 @@ def infer(
         record = provenance.get_provenance_dict(
             command="infer",
             mismatch_ratio=mismatch_ratio,
+            path_compression=path_compression,
+            precision=precision,
+            simplify=simplify,
+            post_process=post_process,
+            # TODO: maybe record recombination rate (which could be a RateMap)
         )
         tables.provenances.add_row(record=json.dumps(record))
         inferred_ts = tables.tree_sequence()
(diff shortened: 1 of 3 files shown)


tsinfer/inference.py
```diff
@@ -377,10 +377,16 @@ def infer(
...
+        params = {  # TODO: maybe record recombination rate (which could be a RateMap)
+            "mismatch_ratio": mismatch_ratio,
+            "path_compression": path_compression,
+            "precision": precision,
+        }
+        if simplify is None:
+            params["post_process"] = post_process
+        else:
+            params["simplify"] = simplify
+        record = provenance.get_provenance_dict(command="infer", **params)
```
jeromekelleher, line 389: This seems like a backward step duplicating the logic for dealing with simplify.

Would it be better to pass everything to ``get_provenance_dict`` and sort out any messiness that's needed in there, once?

hyanwong: Yeah, it is a bit messy, I thought it might be overkill to keep the backwards compatibility, but then realised that we would potentially be reporting *wrong* parameters, as `simplify` now sets the post_process variable to True.

I've now done what you suggested and stuck the logic in get_provenance_dict, thanks for the suggestion.

jeromekelleher (approved):

benjeffery (approved):

PR #1000, opened 2025-02-25
Rename map_additional_sites

### LINKED ISSUE #862: Document `map_additional_sites`` argument
Added in #861 

Should possibly think of a different name ("map_non_inference_sites"?)

This is really part of the "post processing" so we should consider how that fits in.

benjeffery:
Fixes #862 

I've gone with `overlay_non_inference_sites` as I think it gets across what is happening better than `map`, but happy to change.
I considered moving the functionality to `post_process` but as it needs the sample data and post process doesn't, I've left it where it is for now.

### DIFF
--- a/tests/test_inference.py
+++ b/tests/test_inference.py
@@ -3422,22 +3422,22 @@ def get_default_inference_sites(sample_data):
 
 
 class TestMapAdditionalSitesToggle:
-    def test_map_additional_samples_toggle(self):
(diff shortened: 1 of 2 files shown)


jeromekelleher (approved):
SGTM

@hyanwong, thoughts on potential breakage?

jeromekelleher:
Let's do it so

PR #1001, opened 2025-02-27
Add ancestral state conversion method

benjeffery:
Adds a method to prepare the ancestral state array from a string.


jeromekelleher:
Can you give an example of how this would be used in a typical (i.e. top level documentation) example?

benjeffery:
> Can you give an example of how this would be used in a typical (i.e. top level documentation) example?

Just finishing that up on a separate PR.

benjeffery:
@jeromekelleher I've added the docs change here

docs/usage.md
```diff
@@ -68,17 +68,21 @@ on this issue to provide some recommendations.
...
 
 ```{code-cell}
 import tsinfer
 
-# For this example take the REF allele (index 0) as ancestral
-ancestral_state = ds['variant_allele'][:,0].astype(str)
-# This is just a numpy array, set the last site to an unknown value, for demo purposes
-ancestral_state[-1] = "."
+ancestral_string = "GGCTCAG."
+ancestral_state, _ = tsinfer.VariantData.allelic_states_from_string(
```
jeromekelleher, line 82: This seems like a footgun - why do we make ``allelic_states_from_string`` a class method and ask the user to pass in the variant_position, when they'll be making a VariantData from that dataset? Why are we passing back something that's not used?

Is there some way we could pass the ancestral state to the constructor itself?

jeromekelleher (commented):
API smells a bit off to me. Is there some other way we can organise this?

docs/usage.md
jeromekelleher: That's what I was wondering, yeah

benjeffery: I've tacked on a commit with one way to do it by allowing the user to pass a numpy array, zarr key or FASTA string to `__init__`. I'm not sure I like this though and I think a separate argument might be better.

jeromekelleher: Not seeing the commit, did you push?

benjeffery:
@jeromekelleher method now saves to VCF Zarr as discussed.

jeromekelleher (commented):
I think we need to be crystal clear about about zero vs one-based indexes here, and update the documentation to clarify at a minimum

benjeffery:
Ok, I've updated everything to be 1-based and to make that clear in the docs.

@hyanwong I had to `+1` the positions in the docs example - but that seems to have broken it with a "missing data is unsupported" error. I must be missing something as I don't get how changing the positions can make it break?

docs/usage.md
```diff
@@ -68,22 +68,27 @@ on this issue to provide some recommendations.
...
+import zarr
 
-# For this example take the REF allele (index 0) as ancestral
-ancestral_state = ds['variant_allele'][:,0].astype(str)
-# This is just a numpy array, set the last site to an unknown value, for demo purposes
-ancestral_state[-1] = "."
-
-vdata = tsinfer.VariantData("_static/example_data.vcz", ancestral_state)
+vcf_zarr = zarr.open("_static/example_data.vcz")
+ancestral_string = "GGCTCAG."
```
jeromekelleher, line 83: What I find simplest is to prepend (say) "X" to the FASTA string so that it's effectively in one-based indexes. I think we need to help people out a bit more here as they'll definitely get it wrong.

jeromekelleher (commented):
Happy to merge this as is, but we need a better user story for how to deal with importing data from fasta.

What the user needs to see is something like
```python
reader = pyfaidx.Reader("ancestral_state.fa")
# Convert to 1-based coordinates by prepending arbitrary character
chr1_as = "X" + reader["chr1"] 
tsinfer.add_ancestral_state_array(zarr_path, chr1_as)
```

benjeffery:
@hyanwong don't worry about the docs error, have a fix

benjeffery:
Ok, hopefully this makes sense - I've updated the docs to show loading in a FASTA from the example directory and what one should do if the positions were one-based, both in the code and text.

jeromekelleher (approved):
LGTM - let's ship this and refine later as we see how users get on

PR #1002, opened 2025-03-05
Deprecate SampleData

benjeffery:


jeromekelleher (approved):

benjeffery:
@jeromekelleher I've added quite a bit here since you reviewed, renaming function arguments where it made sense to. If you think this is too radical let me know.

pyproject.toml
```diff
@@ -80,3 +80,6 @@ include-package-data = true
 
 [tool.pytest.ini_options]
 testpaths = ["tests"]
+filterwarnings = [
+    'ignore:SampleData'
```
jeromekelleher, line 84: nice :+1:

jeromekelleher (approved):
LGTM. We're not going to actually break anyone's code with this, right?

benjeffery:
> LGTM. We're not going to actually break anyone's code with this, right?

The only case where we would is if someone had specified a keyword arg for `sample_data`, which as it is the first argument in almost all cases seems unlikely.

Issue #911, opened 2024-03-21
Improve ancestor building by using more information


hyanwong:
It looks like #903 is being caused by bad choices of ancestral vs derived states in the ancestor builder. We do have more information for building, e.g. we know after the first inference (or if we use the PBWT) whether we think that many of the haplotypes contributing to the consensus form a cluster of recent close relatives.

We should think about how to use this information to improve ancestor reconstruction. @petrelharp suggested building a "decision tree" rightwards and leftwards from each site, branching whenever we make a choice of 0 or 1 for an adjacent site, and picking e.g. the longest path. He pointed out that this can be (mostly) kept between ancestors, so should be efficient.

Issue #903, opened 2024-02-13
Investigate why sample nodes attach to deep roots / ultimate ancestor


hyanwong:
In trees from real data, we often find sample nodes attached via long branches to deep in the trees, often to the root. We should investigate why this is happening.

I wondered at first if these deep-rooted samples would be partially removed if we get the sites in the correct order, which [seems to produce much better dating results](https://github.com/tskit-dev/tsinfer/issues/877#issuecomment-1830420672). But some experimentation seems to suggest this doesn't help much (and unsurprisingly, neither does a round of dating + reinference):

<img width="494" alt="Screenshot 2024-02-13 at 20 27 37" src="https://github.com/tskit-dev/tsinfer/assets/4699014/9b19f029-5aa7-4581-897a-3b99a88cbd88">

```python
import tsinfer
import tsdate
import numpy as np
import stdpopsim
species = stdpopsim.get_species("HomSap")
model = species.get_demographic_model("AmericanAdmixture_4B11")
contig = species.get_contig("chr22", mutation_rate=model.mutation_rate, length_multiplier=0.05)
engine = stdpopsim.get_engine("msprime")
true_data = {}
inferred_data = {}
truedate_data = {}
infdated_data = {}

n_reps = 5
for s in (1, 10, 100):
    samples = {"AFR": s, "EUR": s, "ASIA": s, "ADMIX": s}
    num_samples = sum(samples.values()) * 2
    true_data[num_samples] = np.zeros(num_samples)
    inferred_data[num_samples] = np.zeros(num_samples)
    truedate_data[num_samples] = np.zeros(num_samples)
    infdated_data[num_samples] = np.zeros(num_samples)
    for rep in range(n_reps):
        sim_ts = engine.simulate(model, contig, samples, seed=rep+123)
        print(sim_ts.num_samples, "samples,", sim_ts.num_trees, "trees,", sim_ts.num_sites, "sites")
        tTS = tsinfer.infer(tsinfer.SampleData.from_tree_sequence(sim_ts, use_sites_time=True))
        iTS = tsinfer.infer(tsinfer.SampleData.from_tree_sequence(sim_ts, use_sites_time=False))
        dTS = tsdate.variational_gamma(
            tsdate.util.split_disjoint_nodes(iTS.simplify()),
            mutation_rate=model.mutation_rate,
            population_size=sim_ts.diversity() / (4 * model.mutation_rate),
        )
[...]

PR #1012, opened 2025-03-14
Revise ancestor generation algorithm to improve performance

duncanMR:
A major limitation of tsinfer 0.4 has been that ancestors with high frequency focal sites are excessively long. This reduces parallelism and wastes computation time in ancestor matching. We've explored a number of solutions (e.g. #911), but we've found a simple one that seems to work really well.

It's easiest to explain with an example:
image
Constructing an ancestor with focal site 5 and moving to the left, we start with a sample set of C-F. Since the focal AC is 4, we can only use sites with an AC of 5 or higher as inference sites. In the current implementation, we would ignore sites 1-4 entirely, but we are then missing an informative signal. Since B also carries the derived allele at site 3, and is the only other non-carrier at site 2, it seems that C has recombined into a clade with B and should be excluded from the sample set.

In the new approach, we still only insert a mutation into the ancestral haplotype if it is older than the focal site, but we use sites of all frequencies for determining when to exclude samples from the sample set. However, we only count conflicts at sites where there are carriers outside of the sample set (i.e. the derived AC in the full population > AC in the sample set).

Initial validation of the new approach looks promising. I ran a small out-of-Africa sim of 1mbp with 200 samples. Using the old validation code, I can match the inferred ancestors to simulated ones and compute `max(true_left - inferred_left, inferred_right - true_right)`, which I call the max overshoot. While the old algorithm increasingly overestimates the length of the ancestors as the focal frequency increases, the new algorithm does not. We don't seem to make many ancestors shorter than their true length either.

image

I've done some testing on larger simulations with similar results, but there is still a lot more validation to do, e.g. with mispolarised alleles and sequencing errors. I did run @hyanwong's code for analysing sample-to-root edges from #903, and in simulations we don't seem to see any change. However, there are a lot more sample-to-root edges in real data with the new method, which is fixed by using a small mismatch ratio for sample matching.

The new approach does require more work from the CPU for ancestor generation, but it seems negligible in comparison to the gains in parallelism and speed of ancestor matching. Comparing the methods on chr20q of the 1000 Genomes Project, we see that fewer ancestor groups are needed and there is a 5X decrease in CPU time needed for matching. The decrease in wall time is more modest: it seems that we aren't using the 126 threads of this EPYC system as effectively in the new approach.

image

I've implemented the new algorithm in C and Python: the biggest change required is that the derived allele count of a site needs to be stored in the ancestor builder.

### DIFF
--- a/lib/ancestor_builder.c
+++ b/lib/ancestor_builder.c
@@ -419,8 +419,9 @@ ancestor_builder_compute_ancestral_states(const ancestor_builder_t *self, int di
     tsk_id_t last_site = focal_site;
     int64_t l;
     tsk_id_t u;
-    size_t j, ones, zeros, tmp_size, sample_set_size, min_sample_set_size;
+    size_t j, ones, zeros, tmp_size, sample_set_size, min_sample_set_size, derived_count;
     double focal_site_time = self->sites[focal_site].time;
+    double site_time;
     const site_t *restrict sites = self->sites;
     const size_t num_sites = self->num_sites;
     allele_t consensus;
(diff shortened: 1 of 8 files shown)


jeromekelleher:
Not sure why you're touching _tsinfermodule.c here @duncanMR, but that needs to be dropped from the diff. Your editor probably got overzealous there.

lib/ancestor_builder.c
```diff
@@ -676,6 +683,7 @@ ancestor_builder_add_site(ancestor_builder_t *self, double time, allele_t *genot
     pattern_map = &time_map->pattern_map;
     site = &self->sites[site_id];
     site->time = time;
+    site->derived_count = derived_count;
```
jeromekelleher, line 686: Why are we passing in the derived count? Is this the same as the number of entries in the genotypes > 0? 

If so, we should just compute it there, and leave the signatures alone. Lets not worry about one extra iteration over the genotypes at  site.

jeromekelleher (commented):
Generally looks great, but not obvious that we need the API changes to pass through derived count (definitely a premature optimisation if it can be computed easily).

lib/ancestor_builder.c
duncanMR: I did it that way because `genotypes` only contains the samples in the initial sample set once the loop starts, but for this algorithm we need the derived count across all samples for all sites. So, to calculate the derived count we would need to call `ancestor_builder_get_site_genotypes` for every site in the loop instead of `ancestor_builder_get_site_genotypes_subset`, which I expect would have a significant performance impact.

duncanMR:
@benjeffery I've simplified the changes based on Jerome's suggestions. I think the plan is to merge this in after 0.4.1 has been released, if that's okay with you?

benjeffery:
Yep, makes sense to me! 0.4.1 should be tomorrow.

jeromekelleher (approved):
LGTM!

hyanwong:
A trivial point, but perhaps you want to change the plot in the [large_scale.md](https://github.com/tskit-dev/tsinfer/blob/main/docs/large_scale.md) doc file, to show the new matching profile, if it is much different. Here's what we currently have (I guess @benjeffery knows how it was generated):

Screenshot 2025-03-19 at 14 45 04

benjeffery:
That plot was from a GeL chromosome, but any similar will do that illustrates the point.

jeromekelleher:
Good to merge @benjeffery?

benjeffery:
Merging! Thanks @duncanMR this is awesome!

benjeffery:
@mergifyio rebase

PR #1013, opened 2025-03-17
Doc tidying

### LINKED ISSUE #1007: Add an indel to the "toy" example
It would be good to change one of the sites in the "toy example" to contain a (short) indel, to show that we can use these perfectly well in tsinfer. Would this mess with the idea of a FASTA ancestral state though (I think it should be OK though)?

I have had question posed about using indels, so this would be good to demonstrate.

### LINKED ISSUE #1008: Use non-adjacent position in the toy example
I think it's a bit confusing that the toy example in the docs uses positions 0, 1, 2, 3, 4, 5. You therefore can't tell if the "Site:" label is referring to the site id or the site position.

If the site position, then values of 0 might seem very weird to bioinformatics people, who are used to counting positions from 1. Perhaps we can get around that by starting e.g. at position (say) 4?

hyanwong:
Fixes #1008. Fixes #1007. Tidies up some of the viz labels.


hyanwong:
I'm wondering if the table should read "TS node" rather than "TS sample". After all, these are node IDs not (really) sample IDs.

```
TS node  	0	1	2	3	4	5
TS individual	0	0	1	1	2	2
------------------------------------------------------------
Pos (anc state)
16   (G)	T   	T   	T   	G   	T   	G   
44   (G)	G   	G   	G   	G   	T   	T   
50   (C)	G   	G   	G   	C   	C   	C   
55   (T)	A   	T   	T   	T   	A   	T   
71   (C)	C   	C   	C   	C   	C   	C   
75   (A)	T   	T   	T   	A   	T   	T   
85   (G)	C   	G   	G   	C   	C   	C   
95   (AGGT)	AGGT	AGGT	A   	AGGT	AGGT	AGGT
```

docs/usage.md
```diff
@@ -32,34 +32,39 @@ document. However, for the moment we'll just use a pre-generated dataset:
...
-sites = np.arange(ds['call_genotype'].shape[0])
-print(" " * 22, "Site:", " ".join(str(x) for x in range(8)), "\n")
-for sample in range(ds['call_genotype'].shape[1]):
-    for genome in range(ds['call_genotype'].shape[2]):
-        genotypes = ds['call_genotype'][:,sample, genome]
-        print(
-            f"Diploid sample {sample} (genome {genome}):",
-            " ".join(alleles[sites, genotypes])
-        )
+assert all(len(np.unique(a)) == len(a) for a in vcf_zarr['variant_allele'])
```
jeromekelleher, line 43: We need to be careful here not to give people patterns that will have poor performance on larger datasets. Elementwise iteration over zarr can have very poor performance, so you're nearly always better off getting the numpy array, like`` vcf_zarr['variant_allele'][:]``. I know it's weird, but it's better than having stuff running really slowly.

docs/usage.md
```diff
@@ -32,34 +32,39 @@ document. However, for the moment we'll just use a pre-generated dataset:
...
-print(" " * 22, "Site:", " ".join(str(x) for x in range(8)), "\n")
-for sample in range(ds['call_genotype'].shape[1]):
-    for genome in range(ds['call_genotype'].shape[2]):
-        genotypes = ds['call_genotype'][:,sample, genome]
-        print(
-            f"Diploid sample {sample} (genome {genome}):",
-            " ".join(alleles[sites, genotypes])
-        )
+assert all(len(np.unique(a)) == len(a) for a in vcf_zarr['variant_allele']) 
+assert any([np.sum(g) == 1 for g in vcf_zarr['call_genotype']]) # at least one singleton
```
jeromekelleher, line 44: I think it would be better here to explicitly load the genotypes into a numpy array, so people can see this is a toy example. Like,
```
G = vcf_zarr['call_genotype']][:] # read full genotype matrix into memory
# do stuff with G
```

jeromekelleher (commented):
Looks good overall, but the current documentation example would have horrible performance if copied and pasted and applied to larger dataset (especially with Zarr v3). Need to be much more explicit and careful about loading arrays into memory, and help the reader understand what's going on.

jeromekelleher:
Basically, always get your data into a numpy array and interact with the Zarr store as little as possible.

docs/usage.md
hyanwong: Very happy to change this to the most efficient style. Maybe worth adding a comment about this in the code?

docs/usage.md
hyanwong: This isn't actually shown, BTW. The code is hidden

jeromekelleher: Best change it now anyway, as the code could actually be slow enough to break the build with Zarr v3 (which we'll support at some point).

So, load any Zarr array you're accessing into a numpy array at the start, and then interact with that array.

benjeffery:
Thanks @hyanwong! I agree that not element wise operations shouldn't be promoted, but otherwise looks great.

hyanwong:
Thanks both. Changes made. I assume that `vcf_zarr['variant_position'][-1]` to get the last position is still OK, though? As is something like `alleles = vcf_zarr['variant_allele'][site_id].astype(str)` to get the allele vectors one-by one?

hyanwong:
> As is something like alleles = vcf_zarr['variant_allele'][site_id].astype(str) to get the allele vectors one-by one?

Scratch that, I just pull the whole alleles matrix over now, so as to avoid repeatedly accessing the zarr store

jeromekelleher (approved):

benjeffery:
@mergifyio requeue

PR #1025, opened 2025-03-29
Tidy the usage docs, including ancestral state specs

### LINKED ISSUE #1024: Should the documented example use 1-based .vcz positions
At the moment, the first "Toy Example" reads from a .vcz file but then when setting ancestral states assumes that these variants have 0-based positions. However, I thought that VCF files were basically not allowed to be 0-based, so this is might be confusing to most users, who will pretty much always be using 1-based indexing, I imagine.

Should we simply assume that example_data.vcz is 1-based, and hence append a padding character to the FASTA string? The example code is there, just commented out, so it's a trivial change.

I'm tidying this bit of the manual, so can implement this if that seems like a good idea.

hyanwong:
Also fixes #1024


jeromekelleher (approved):
LGTM

docs/usage.md
```diff
@@ -63,31 +63,64 @@ for site_id, pos in enumerate(positions):
...
+Ancestral states can be specified in several ways:
+
+* Using an existing array in the .vcz file. Some VCF files already contain an ancestral
+  state, e.g. in the `AA` ("ancestral allele") `info` field. In this case the `variant_AA`
+  field of the .vcz file can be specified, as follows.
+  ```{code}
+  vdata = tsinfer.VariantData("file.vcz", ancestral_state="variant_AA")
+  ```
+  A worked example is shown in the {ref}`real data example<sec_usage_read_vcf_inference>`
+  later on this page
```
benjeffery, line 102:
```suggestion
  later on this page.
```

benjeffery (approved):
Very nice, thanks! I found one teeny tiny nit.

docs/usage.md
hyanwong: Nice spot. And it does indeed look like a nit egg.

Added in latest commit

PR #1043, opened 2025-09-09
Error on zarr v3

### LINKED ISSUE #1041: Add check for zarr v<3?
Some people have started hitting problems because they are installing tsinfer with zarr v3. I think it would be useful to emit a warning that this won't work, and recommend downgrading to < v3. I don't know if this can be done by tweaking the installation config, or whether we should just drop in an explicit check somewhere?

benjeffery:
Fixes #1041


hyanwong:
LGTM

benjeffery:
FYI I tried for a bit to write a compatibility layer - but it gets really hairy around object encodings and you have to roll your own LMDBStore.

PR #1051, opened 2025-09-29
Close various files as we're running out of space on Windows CI

benjeffery:


benjeffery:
Windows still tormenting me here - refactoring the CLI tests to remove problematic temp file handling, so did the full pytest conversion and removal of Python 2 stuff.

benjeffery:
Linux can cleanup a file left open, windows can't so I need to make sure everything is `closed` properly.

jeromekelleher:
Would an alternative be to have a destructor handler on the ancestor data class which closes the file when the object is destroyed?

benjeffery:
> a destructor handler on the ancestor data class which closes the file when the object is destroyed?

_Maybe_ we'd be reliant on GC running before we ran out of disk - which it probably would.
One issue would be depending on the destruction order the bits that `close` needs might not be there. I think the right approach would be a [finaliser](https://docs.python.org/3/library/weakref.html#weakref.finalize)  on the class. I'll try that  out in a different PR.

PR #1052, opened 2025-10-02
Add finaliser to close open stores

benjeffery:


jeromekelleher (approved):
I don't really understand this, but if it fixes up the Windows woes then +1

Would be good to document why we're bothering with this -- basically it's our sloppy test suite and Windows?

benjeffery:
> I don't really understand this

We hold a `weakref` to the store. When that is the only remaining reference, the finaliser is called and then the object is garbage collected.

PR #1054, opened 2025-10-02
Add low-level disable-weight-by-n flag

### LINKED ISSUE #1047: Add low-level flag to HMM to turn on/off weighting by n
Like this commit: https://github.com/tskit-dev/tsinfer/pull/959/commits/bd0deedb9a0efd42dc83cd01770fd867c9460c1d

We want to divide by n by default, and allow it to be turned off in an undocumented API

benjeffery:
Fixes #1047

Did we need to bring this up to the high-level? I wasn't clear on that.


lib/ancestor_matcher.c
```diff
@@ -905,21 +909,21 @@ ancestor_matcher_run_forwards_match(
                        && likely(L_cache[u] == CACHE_UNSET)) {
                     u = parent[u];
                 }
-                L_child = L_cache[u];
-                if (unlikely(L_child == CACHE_UNSET)) {
-                    L_child = L[u];
-                }
+                    L_child = L_cache[u];
```
jeromekelleher, line 912: What's happened here? Looks like clang-format has run amok or something.

_tsinfermodule.c
```diff
@@ -1274,8 +1274,9 @@ AncestorMatcher_init(AncestorMatcher *self, PyObject *args, PyObject *kwds)
     int ret = -1;
     int err;
     int extended_checks = 0;
+    int disable_weight_by_n = 0;
```
jeromekelleher, line 1277: Let's call it ``weight_by_n``, default=1 for this inferface. Agree with the way the C flags are done.

jeromekelleher (commented):
Looks good, let's just keep it at the low-level for now. A few minor things need sorting.

PR #1057, opened 2025-10-06
Add atomic blkalloc

### LINKED ISSUE #1050: Add copy of tsk_blkalloc using Atomics
Following https://github.com/tskit-dev/tsinfer/pull/959/commits/57f9209dcca48fbe70d397dba67511f4def3ae49  #959

benjeffery:
Fixes #1050

### DIFF
--- a/_tsinfermodule.c
+++ b/_tsinfermodule.c
@@ -14,6 +14,7 @@
 #define IS_PY3K
 #endif
 
+
 #define MODULE_DOC \
 "Low-level tsinfer interface."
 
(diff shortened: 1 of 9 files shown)


jeromekelleher:
Looks good. I'd just return SSIZE_MAX or something on Windows, it'll be obviously wrong but not break code calling it.

_tsinfermodule.c
```diff
@@ -1535,6 +1536,12 @@ AncestorMatcher_get_total_memory(AncestorMatcher *self, void *closure)
 {
     PyObject *ret = NULL;
```
jeromekelleher, line 1538: This will potentially break some tests. Better to do something like this:
```c
unsigned long val;

...

#if defined(TSI_NO_ATOMICS)
     val = (unsigned long) PY_SSIZE_T_MAX);
#else    
     val = ancestor_matcher_get_total_memory(self->ancestor_matcher));
endif;
ret = Py_BuildValue("k", val);

```

Always keep control flow as simple as possible in C.

PR #1058, opened 2025-10-14
Add a dummy site at the end of the sequence

Duncan-JR:
Since we use half-open intervals in tskit, and tsinfer works with discrete sites, ancestors that include the rightmost inference site (let's call its index `r`) have to be defined as ending at site index `r+1`. Since this extra site has no position or haplotypes, we have to handle it as a special case at times, hard-coding it to have a position of `ts.sequence_length`. 

This work-in-progress PR seeks to explicitly encode this rightmost extra site, which I've tentatively called a breakpoint, in the ancestor data. This is the first step toward adding support for multiple user-provided breakpoints; for example, the user could specify the start of the centromere as a breakpoint, such that tsinfer would run on each chromosome arm separately. 

So far, I've implemented the extra site in the Python engine up until ancestor generation. Next step is to get ancestor and sample matching to work. Here is a small simulated example to show how the change affects ancestral haplotypes:

image

### DIFF
--- a/_tsinfermodule.c
+++ b/_tsinfermodule.c
@@ -185,9 +185,7 @@ AncestorBuilder_add_terminal_site(AncestorBuilder *self)
     if (AncestorBuilder_check_state(self) != 0) {
         goto out;
     }
-    Py_BEGIN_ALLOW_THREADS
     err = ancestor_builder_add_terminal_site(self->builder);
-    Py_END_ALLOW_THREADS
     if (err != 0) {
         handle_library_error(err);
         goto out;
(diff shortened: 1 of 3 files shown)


lib/tests/tests.c
```diff
@@ -532,6 +532,68 @@ test_ancestor_builder_one_site(void)
...
+    allele_t genotypes_derived[4] = { 1, 0, 1, 0 };
+    allele_t genotypes_terminal[4] = { 0, 0, 0, 0 };
+    allele_t ancestor[2];
+    tsk_id_t start, end, focal_sites[1];
+
+    ret = ancestor_builder_alloc(&ancestor_builder, 4, 2, -1, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    ret = ancestor_builder_add_site_default(&ancestor_builder, 4, genotypes_derived);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    ret = ancestor_builder_add_site(&ancestor_builder, 5, genotypes_terminal, true);
```
jeromekelleher:
```
ret = ancestor_builder_add_terminal_site(&ancestor_builder);
```

PR #1117, opened 2026-03-09
Tsinfer 1 design draft

jeromekelleher:
Here's a first pass at a design document for tsinfer 1.0. It should be a reasonable sketch of what I'm hoping to get to.

It would be good to get your input here @benjeffery @hyanwong @Duncan-JR

### DIFF
--- a/design.md
+++ b/design.md
@@ -152,12 +152,20 @@ sample_time = 1          # scalar: all samples in this source at time=1
 
 ### Configuration file
 
-A single TOML file describes the entire pipeline. The standard case — a single
-cohort, with ancestor VCZ written to disk, then samples matched against it:
+A single TOML file describes the entire pipeline. There is one `[match]`
+section that lists all sources — the ancestor VCZ and all sample sources —
+together. The grouping algorithm reads `sample_time` from each source and
+assembles all haplotypes into an ordered sequence of groups automatically.
+
+The ancestor VCZ is auto-identified as the source whose path matches
+`[ancestors].path`. Haplotypes from it do not produce individuals; haplotypes
+from all other sources do. No explicit `create_individuals` flag is needed.
+
+Standard case — a single cohort with modern samples only:
 
 ```toml
 [ancestral_state]
-path  = "annotations.vcz"        # VCZ containing the ancestral allele array
+path  = "annotations.vcz"
 field = "variant_ancestral_allele"
 
 [[source]]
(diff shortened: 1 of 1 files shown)


design.md
```diff
@@ -0,0 +1,640 @@
...
+2. A linesweep over genomic positions identifies all pairs of ancestors where
+   one is older and their spans overlap, adding a directed dependency edge.
+3. A topological sort of the dependency graph assigns each ancestor a group
+   index. Ancestors with no inter-group dependencies may end up in the same
+   group even if they come from different time epochs, maximising the
+   parallel work available per group.
+4. Same-time ancestors are un-merged, restoring the original ancestor
+   identities within each group.
+
+**Epoch regime (later ancestors).** For large epochs the linesweep dependency
```
hyanwong: We might need to consider what happens when we use site times given by an earlier round of inference (e.g. using tsdate in an iterative process)

hyanwong (commented):
This all looks very clean to me. The only things that are not covered that I can think of are:

1. historical samples (probably not part of the MVP, but we should make sure that the possibility remains open, as this is becoming more and more requested). Similarly we might eventually want the ability to add batches of samples iteratively (e.g. successive generations within a pedigree), with the possibility of samples copying from other samples.
2. check it works with the iterative pathway, where we pass in site times from tsdate. We know that this can improve inference if we have the time to do it.
3. Is there a public API for matching a single new sample (from an alternative VCF) against an existing tree sequence? This is a frequently requested facility.
4. I'm not sure how we override e.g. ancestral states, if they aren't present in the original .vcz samples file? The ability to do this was useful: removing the VariantData wrapper means we presumably need another way, e.g. by editing the vcz file directly? What about defining populations, metadata, etc: do we just hardcode where those are stored in the vcz?

Duncan-JR:
This looks great to me @jeromekelleher! I just wanted to clarify a few things:

1. Do you want us to include the distributed Snakemake pipeline in tsinfer 1.0, or keep it as a separate repo? While some bits of @benjeffery's pipeline will no longer be needed (like site density filtering), most of the early QC filters will still be useful, although not strictly necessary to run tsinfer in general.
2. While not in the MVP, are you planning for us to include the automatic genome chunking functionality in the configuration stage? I'm thinking that the parameters (such as the maximum allowed gap between sites) would be additional fields in the `toml` configuration, with some helper functions to help determine their optimal values?

PR #1178, opened 2026-03-31
Path compression 

jeromekelleher:


jeromekelleher:
This implementation is naive and slow, so not ready for merging yet.

hyanwong:
Just FYI, we had always (I think) intended to try to use more shared recombinations where possible. In particular, where an identical path switch is present on one side of a pair of edges, but not the other. However, this requires a bit more thought than the current path compression, so I assume it's nothing to do with the MVP, and we are leaving it until later.

For my future self, here's the diagram from the paper:

Screenshot 2026-03-31 at 07 39 58

jeromekelleher:
I'm going to merge this for now to avoid conflicts, but disable path compression by default for now and log an issue to follow up with a more performant implementation (ideally after testing this version out on some real data to verify it does the same thing as the 0.x version).

PR #1181, opened 2026-04-01
Add recomb mismatch

### LINKED ISSUE #1180: Add basic support for recombination and mismatch parameters
To implement changes to how the Li and Stephens HMM operates in tsinfer 1.0, we need to be able to specify in the config for each source:

1. Recombination map files that will be loaded in and converted to recombination probabilities as before. For now, we can reuse the functions in tsinfer 0.5 `recombination_rate_to_dist` and `recombination_dist_to_prob` to convert the map into probabilities for the HMM: I may alter these later if necessary.
2. A single recombination rate that is internally converted to an array of length `num_sites-1` to feed into the HMM. No need to support an array of arbitrary distinct recombination rate values as in tsinfer 0.5. Mutually exclusive with specifying the recombination map, of course. 
3. A single mismatch probability that is also converted into a uniform array of length `num_sites` for the HMM. 

At a later stage we may want to vary mismatch probabilities by site, but it isn't worth implementing that yet until we know exactly how it's going to work.

jeromekelleher:
Closes #1180 

This adds ``recombination`` and ``mismatch`` to the MatchSourceConfig @Duncan-JR - this should be enough for you to work with while we figure out the actual parameterisation that we want.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -98,6 +98,9 @@ Sample VCZ → `infer_ancestors` → Ancestor VCZ → `match` → raw `tskit.Tre
 
 - Do not be overly defensive - defend only against circumstances that can
   occur within the current codebase.
+- Do not make production code more complex for the sake of minimising 
+  changes to the test suite. Simplicity and clarity of the production code 
+  is imperitive.
 - Prefer dataclasses over tuples when returning multiple values.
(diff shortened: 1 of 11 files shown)


Duncan-JR:
Thanks Jerome, that's perfect!

